PR #5, opened 2017-04-20
Adding Metric Presets

mrpatpat:
This pull request sets RLOC, MCC, MCC as default metrics. If these metrics do not exist in the loaded map then the first three found metrics are used. 
The sample map was adjusted in order to make use of this change.


mrpatpat:
Will be implemented with scenarios

PR #23, opened 2017-06-08
Link to SonarQube Code URL in sonarimporter

ukinimod:
With this pull request it is possible to fill the link field in the cc.json when importing metrics from SonarQube. Later they can be displayed in the visualization.


FernandoES:
Reviewed, everything ok

PR #33, opened 2017-07-04
Maps with less than three metrics

### LINKED ISSUE #17: Simple cc.json does not display anything
[test.txt](https://github.com/MaibornWolff/codecharta/files/1052519/test.txt)
[test2.txt](https://github.com/MaibornWolff/codecharta/files/1052518/test2.txt)

I generated a very simple json file consisting of only one node and two metrics (blabla and blubblub). When I open this file in the online demo at https://maibornwolff.github.io/codecharta/visualization/app/ it has no selected metrices and it shows two nodes "root" and "Parent Leaf". If I choose one metric it shows nothing.

I also tried a different file with the same result.

mrpatpat:
Allows to load maps with less than three metrics. Fixes #17

### DIFF
--- a/visualization/app/codeCharta/core/settings/settingsService.js
+++ b/visualization/app/codeCharta/core/settings/settingsService.js
@@ -67,8 +67,6 @@ class SettingsService {
 
         this.settings.map = data.currentmap;
 
-        console.log(data.metrics, this.getMetricByIdOrLast(1, data.metrics));
-
         if(data.metrics.indexOf(this.settings.areaMetric) === -1){
             //area metric is not set or not in the new metrics and needs to be chosen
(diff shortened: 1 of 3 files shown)


visualization/app/codeCharta/core/settings/settingsService.js
```diff
@@ -151,7 +151,7 @@ class SettingsService {
      * @param {Settings} settings
      */
     correctSettings(settings){
-        var result = settings;
+        let result = settings;
```
thorsten: "const" should be better than "let" as the value won't change anymore.

ukinimod:
Thx. Fixes #17. Code looks good. 👍

Issue #18, opened 2017-06-07
changing display or color settings resets zoom


silasg:
when I change (at least) invert colors or toggle delta cubes, (at least) the z axis zoom is reset. however, the UI control for this zoom is not updated.

mrpatpat:
When I invert the colors or toggle the delta cubes after i changed the zoom level, the z-axis zoom is not reset. Can you specify the problem and add your browser version ?

silasg:
Sorry I meant y axis scaling, not z axis zoom ;) using chrome 58 on mac os, but I think I’ve also ween this using the windows electron container.





__________________________________________________

Blockchain: Euphorie oder Schockstarre – mehr in den
Praxis-Tipps https://www.maibornwolff.de/newsletter/1-2017
__________________________________________________

Silas Graffy
Senior IT Architect


Mobil +49 151 544 22 047
silas.graffy@maibornwolff.de
Baseler Straße 10, 60329 Frankfurt am Main

MaibornWolff GmbH, Theresienhöhe 13, D-80339 München
http://www.maibornwolff.de, Telefon +49 89 544 253 000
USt-ID DE 129 299 525, Amtsgericht München HRB 98058, GF:
Volker Maiborn, Holger Wolff, Jens Rieger, Alexander Hofmann

__________________________________________________

On 9. Jun 2017, at 12:03, mrpatpat <notifications@github.com<mailto:notifications@github.com>> wrote:


When I invert the colors or toggle the delta cubes after i changed the zoom level, the z-axis zoom is not reset. Can you specify the problem and add your browser version ?

mrpatpat:
Thank you for the contribution. I was able to reproduce this issue. I will fix it as soon as possible.

mrpatpat:
fix is merged. It will be part of the next release.

PR #34, opened 2017-07-04
fixing #18 and correcting variable usage (const, let, var)

mrpatpat:
This suggested fix for #18 adds the scaling to the settings object instead of directly calling the scale method from the controller. 
The settings object survives changes like metric switches. This also means that scaling is a part of a Scenario (but can easily be removed from it by modifying scenarioService#applyScenario or the scenario class itself).


ukinimod:
Tested fix for #18. Worked correctly on my machine (win7, chrome).

I also like the change to const.

Issue #20, opened 2017-06-07
make it easier to find out to which package/folder a class belongs


silasg:
two ideas for this goal (both could be implemented, it's not an xor)

1. add a slider or similar to change the margin/padding between buildings. then it would be easier to hover the ground of a block
2. add a tooltip for the name of the selected block showing the complete path and not only the leaf element.

ukinimod:
Thank you Silas for your suggestions. We are considering another option that adds an additional view displaying the hierarchical structure. I also think we should improve the visability of the packages.

ukinimod:
There was another request from a user for a feature like that.

mrpatpat:
In #37 we implemented a margin slider. In general, it works. Problems appear as soon as you turn the margin value higher than the half of the shorter side of a building. This causes the building to disappear since the margin overlaps the building size. The margin actually is more like CSS padding in this case. 
Before releasing this feature we need a way to handle this problem. 

We could (in no particular order): 
* Just accept it 
* Limit the margin slider according to the shortest side of a building
* Limit the slider to "small" values and accept that some "small" buildings disappear
* Increase district/folder area while increasing building margin and therefore acting like CSS margin
* Something else ?

silasg:
Don`t know the implications to implementation and effort but why not leaving the building sizes as they are and just *adding* more space around the elements?

mrpatpat:
Full paths in details are now implemented. This will be part of the next release.

PR #37, opened 2017-07-21
Margin slider

mrpatpat:
adding a slider to set the margin between buildings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [Unreleased]
 ### Added
-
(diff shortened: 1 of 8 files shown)


ukinimod:
Don't know what this function should do exactly. It does strange things in some examples, deleting buildings etc.

mrpatpat:
You are absolutely right. The idea came from #20 and should allow to adjust the margin around buildings. I do not know yet how to solve the problem of disappearing buildings. A discussion in #20 should help.

mrpatpat:
Merged changes into branch [dataServiceRef](https://github.com/MaibornWolff/codecharta/tree/dataServiceRef)

PR #38, opened 2017-07-21
Colorslider

mrpatpat:
adding a color slider for easy color range selection


ukinimod:
I like the functionality. The slider is a bit slow on big projects, but I think that is ok. Exact values are a bit difficult to adjust. Maybe it would be possible to also adjust it with certain values?

PR #55, opened 2017-12-13
Introducing CodeChurn as additional metric in SCMLogParser

ukinimod:
Hi.

I refactored the SCMLogParser in order to introduce new SCM metrics,
e.g. number of lines changed. This is also an preparation for the ability to follow renames and deletes.

Maybe using jdk9 could be a problem with travis ci. This I could not check.

Could one of you please review the changes and make propositions.


ukinimod:
As I had some spare time I actually added CodeChurn as an additional metric in the SCMLogParser. It's only functional for git and logs created via "git log --numstat", i.e. it is necessary to specify "--input-format GIT_LOG_NUMSTAT" as parameter calling scmlogparser.

CodeChurn is defined in this case as sum of additions and deletions in commits concerning this file.

I'd highly appreciate your feedback. Before merging CHANGELOG and README should be changed accordingly.

analysis/import/SCMLogParser/src/main/java/de/maibornwolff/codecharta/importer/scmlogparser/SCMLogParserParameter.java
```diff
@@ -53,15 +52,37 @@ public boolean isAddAuthor() {
...
         }
+        switch (inputFormatNames) {
+            case GIT_LOG:
+                return new GitLogParserStrategy();
+            case GIT_LOG_NUMSTAT:
+                return new GitLogNumstatParserStrategy();
+            case SVN_LOG:
+                return new SVNLogParserStrategy();
+            default:
+                throw new IllegalArgumentException("--git or --svn or --input-foramt must specified");
```
mrpatpat: typo in format

analysis/import/SCMLogParser/src/main/java/de/maibornwolff/codecharta/importer/scmlogparser/parser/git/GitLogNumstatParserStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+        return commitLines.stream()
+                .filter(commitLine -> commitLine.startsWith(AUTHOR_ROW_INDICATOR))
+                .map(this::parseAuthor)
+                .findFirst();
+
+    }
+
+    String parseAuthor(String authorLine) {
+        String authorWithEmail = authorLine.substring(AUTHOR_ROW_INDICATOR.length());
+        int beginOfEmail = authorWithEmail.indexOf('<');
```
mrpatpat: Personally, I would define "<" as a static final constant, since there are already a few of them. This would be consquent and would remove a "magic value" by giving it some context. As far as I know the minimal author line syntax for git commits is indeed "name <email>" therefore this solution should be ok for this use case.

mrpatpat:
I think there are no problems with upgrading to jdk9. According to [this travis build log](https://travis-ci.org/MaibornWolff/codecharta/builds/317735839) there are no breaking errors and everything works just fine.

README.md
```diff
@@ -9,7 +9,7 @@ CodeCharta is a beautiful tool for visualizing code as cities (3D tree maps).
```
mrpatpat: The analysis readme file mentions java 8 not 9

analysis/import/SCMLogParser/README.md
```diff
@@ -7,20 +7,29 @@ Generates visualisation data from repository (Git or SVN) logs. It supports the
...
 ### Executing the SCMLogParser
 
-> `ccsh scmlogparser <log file> [--git|--svn] [--a] [<output file>]`
+See `ccsh -h` for help.
+
+> `ccsh scmlogparser <log file> [--git|--svn] [--add-author] [-o <output file>]`
+
+or
+
+> `ccsh scmlogparser <log file> --input-format [GIT_LOG|GIT_LOG_NUMSTAT|TSVN_LOG] [--a] [-o <output file>]`
```
mrpatpat: --a caused an error and is not defined in --h

analysis/import/SCMLogParser/src/main/java/de/maibornwolff/codecharta/model/input/metrics/CodeChurn.java
```diff
@@ -0,0 +1,26 @@
+package de.maibornwolff.codecharta.model.input.metrics;
+
+import de.maibornwolff.codecharta.model.input.Modification;
+
+public final class CodeChurn implements ModificationMetric<Integer> {
+    public static final String NUMBER_OF_COMMITS = "code_churn";
```
mrpatpat: NUMBER_OF_COMMITS should be named CODE_CHURN, or since it is just a string for the metric name, METRIC_NAME. It also is not used anywhere else and can therefore be private, as far as i can see.
Since metricName() is only used to get a specific value of the instance, it is OK in my opinion to return the string instead of the static final constant. The meaningless string value gets its context and meaning from the method name metricName() and a named constant is not needed in my opinion.
There are some other metric classes which have the same naming/visibility issues. 
Apart from these minor things, I like the concept behind the model implementation. Easy to understand and use.

analysis/import/SCMLogParser/README.md
mrpatpat: I cant find it in in the Parameter class either. Maybe it is some old parameter ?

analysis/import/SCMLogParser/src/main/java/de/maibornwolff/codecharta/importer/scmlogparser/parser/git/GitLogNumstatParserStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final Predicate<String> GIT_COMMIT_SEPARATOR_TEST = logLine -> logLine.startsWith("commit");
+    private static final String AUTHOR_ROW_INDICATOR = "Author: ";
+    private static final String DATE_ROW_INDICATOR = "Date: ";
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("EEE MMM d HH:mm:ss yyyy ZZZ", Locale.US);
+    private static final String FILE_LINE_REGEX = "\\d+\\s+\\d+\\s+\\S+";
+
+    @Override
+    public List<String> listSupportedMetrics() {
+        return Arrays.asList(
+                "code_churn",
```
mrpatpat: These names are used in other classes, too. Maybe it is better to extract them into a common enumeration. May also be overengineered in this use case but could prevent errors from typo's in the future

PR #58, opened 2018-01-12
Merging map nodes in delta maps

mrpatpat:
The delta view between maps now correctly displays nodes as "added" (fully green, delta = height) and "removed" (fully red, delta = height). Map 1 gets all missing nodes from Map 2 and vice versa. The newly added nodes have metrics of 0. Only the unary metric stays 1, therefore it is necessary to choose it as the area metric to see them. 
Also added origin name in node details.

### DIFF
--- a/visualization/app/codeCharta/codeMap/rendering/geometryGenerator.ts
+++ b/visualization/app/codeCharta/codeMap/rendering/geometryGenerator.ts
@@ -58,9 +58,15 @@ export class geometryGenerator {
             y : n.z0,
             z : n.y0,
             width : n.width,
-            height : Math.max(n.height, 5.0),
+            height : this.ensureMinHeightIfDeltaNotNegative(n.height, n.heightDelta),
             depth : n.length
         };
+    }
+
+    private ensureMinHeightIfDeltaNotNegative(x: number, d: number): number {
+        if(d >= 0) {
+            return Math.max(x, 5.0);
+        } else return x;
     }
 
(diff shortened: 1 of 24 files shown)


visualization/app/codeCharta/codeMap/rendering/geometryGenerator.ts
```diff
@@ -58,11 +58,17 @@ export class geometryGenerator {
...
             width : n.width,
-            height : Math.max(n.height, 5.0),
+            height : this.ensureMinHeightIfDeltaNotNegative(n.height, n.heightDelta),
             depth : n.length
         };
     }
 
+    private ensureMinHeightIfDeltaNotNegative(x: number, d: number): number {
+        if(d >= 0) {
+            return Math.max(x, 5.0);
```
ukinimod: Where comes this magic value 5? Maybe some kind of configuration of CC? Is there a coupling to another magic value?

visualization/app/codeCharta/core/data/data.deltaCalculator.service.ts
```diff
@@ -48,10 +49,120 @@ export class DeltaCalculatorService {
...
+            });
+
+            return mapCopy;
+
+        } else {
+            return map;
+        }
+    }
+
+    private insertLeftIntoRightWithZeroMetrics(left: HierarchyNode<CodeMapNode>, right: HierarchyNode<CodeMapNode>) {
```
ukinimod: may factor in function in smaller subfunctions, corresponding to the code comments?

PR #72, opened 2018-01-30
Adding scenarios

ukinimod:
- Defined further scenarios via json file
- Added description for metrics and scenarios

Please also functional review. In my opinion the spacing in the scenario tab is not right. But I don't know how to fix that.


visualization/app/codeCharta/core/scenario/scenarios.json
```diff
@@ -0,0 +1,83 @@
...
+        "x": 0,
+        "y": 300,
+        "z": 1000
+      },
+      "margin": 1,
+      "deltaColorFlipped": false
+    }
+  },
+  {
+    "name": "Coverage",
```
mrpatpat: We should have a distinctive name between the two coverage scenarios

ukinimod: Oh, good point! I'll change that.

PR #73, opened 2018-02-01
Ccsh rework

ukinimod:
Created a different version of ccsh as the old one was not easy to maintain and also had several other drawbacks.
There is one unified library for all command line interfaces.
The new ccsh is located in the bin/ccsh directory. There is no need to call ccsh install any longer.
There is room for further improvement, for example redesigning the call of subcommands. We use 'ccsh import scmlog' instead of 'ccsh scmlogparser'.

Only issue:
- the help of scmlogparser no longer prints the info about the metrics and commands for the git and svn client.


mrpatpat:
Reminder for merging: We need to change some shell scripts which use the cli e.g. gh-pages script and maybe some deployment configurations since the path/usage changed. The node wrapper may need some rework, too.

Issue #64, opened 2018-01-25
slider input


mrpatpat:
sliders should offer an additional textbox to allow precise value input

mrpatpat:
implemented in #78. This will be released as soon as #78 is merged and released

PR #78, opened 2018-02-13
Dependencies

mrpatpat:
This PR adds an experimental dependency visualization. It uses arrows which point from the selected building to its dependencies. Dependency visualization can be toggled. 
bildschirmfoto 2018-02-13 um 15 58 07
This feature requires changes in the accepted json schema. It will accept an optional array of dependencies and is therefore not a breaking change.
This PR also includes some fixes and enhancements:
- canvas mouse event listener are now limited to the canvas dom element. UI events will not trigger the canvas listeners anymore -> this allows changing parameters without the arrows disappearing
- canvas mouse events distinguish now between click and drag. Dragging does not reset selection anymore -> this allows dragging the scene without the arrows disappearing
- slider input #64 -> enhancement (propably should have been a unique PR...)
- rz slider initialization bug (color slider did not initialize handles correctly)-> bug fix
- increasing test coverage -> enhancement (propably should have been a unique PR...)


ukinimod:
Hi @mrpatpat. 

I tried your change, but I had no examples with dependencies. Are there any test data in the repo?

I compared performance of the app of this version with deactivated dependencies with the previous version and there seemed to be essentially no impact in performance. So I'd consider merging after code review.

mrpatpat:
Hi @ukinimod,
thank you for your feedback. There should be dependency data in sample1.json (loaded by default). Two of five nodes should have arrows.

PR #85, opened 2018-02-23
adding all metrics to detail view

mrpatpat:


ukinimod:
I like this. I just wonder if it is better to switch to non-detailed view when hovering changes or not. Anyway, I'd appreciate merging.

Issue #96, opened 2018-03-02
Keeping Settings when loading new files


ukinimod:
At the moment the url is the only possibility to save visualization settings, e.g. margin, perspective, scale etc. I think it would be great if loading a new file would not change the pre-selected settings.

Also, it would be great if it were possible to open delta maps via url, if e.g. automatically comparing maps between two commits.

mrpatpat:
This is actually a very important feature. Sadly we cannot persist all settings, some settings are map related e.g. selected Metrics. Do you think we should start thinking about the usage of local storage ? This would allow us to save data between sessions.

ukinimod:
Ok, I mixed two issues, again.

1. Foremost it would be great if loading new files via open would preserve the actual settings. I think it is not necessary to reset the settings to default when loading a new file. 
2. @mrpatpat I think, we have to consider saving the settings on local storage. Otherwise it is not possible to have the same view on a source between two sessions. I think we could also consider storing settings directly in the json file with the visualization data - at least as a first step. 
3. Some url parameter don't make sense as is. E.g. delta as url parameter is not needed when only one file can be given via url an no comparison or reference map can be chosen. I would love to use the delta view via url, but if it is not possible we should drop delta as url parameter.

mrpatpat:
1. I started implementing this feature. This was once a requirement but it is not necessary anymore.
2. I think there are pro's and con's for saving individual settings data in a cc.json file. I always thought of this file as a raw data container. I think I would prefer a local storage solution which remembers the settings per file/project name. This keeps it out of the file and makes it more personal for the user. We need to remember that local storage can be cleared and is tied to the given browser/computer combination.
3. Absolutely. I will implement this.

PR #99, opened 2018-03-08
Keep settings when loading a new map

mrpatpat:
Implementing part 1 of #96


ukinimod:
This fix didn't keep all settings, e.g. metrics was changed when reloading.

PR #105, opened 2018-03-26
Settings panel labels

alschmut:


mrpatpat:
Is this the PR related to #105 ? If yes, we could close it for the same reason #105 was closed. We basically fixed this bug on another branch by removing an old ui framework.

alschmut:
Yes, this is the complementary PR.

PR #106, opened 2018-03-27
Frontend refactoring

mrpatpat:
# Description

This PR mainly replaces materialize-css with the angularJS material library. 
It suits the needs of our current setup better and has a better documentation. It also integrates better into angularJS. 
This PR also refactors the ui package in order to achieve a better modularization and more clean code. 
Implements a regex filter in the treeview tab. This allows simple package/class filtering via regular expressions. 
Implements a dialog service which allows dialogs (error, prompt, alert, etc. ) to be shown. 

# Full changelog

### Added
- SASS support
- simple regex filter
- Reset Button
- Dialog Service replaces console log calls and window.alert calls
- linking tree view and map hover
- auto fit scene button
- angularJS material

### Removed
- materialize-css

### Fixed
- less flickering and artifacts


alschmut:
@mrpatpat would you like to resolve the conflicts, as I am not 100% sure, if I had missed anything important?

Issue #121, opened 2018-04-05
Input-field in colourSettings doesn't work as expected


alschmut:
# Bug

## Expected Behavior
When I change the numbers inside the input-fields of the colorSettingsPanel:
- the rangeSlider adapts to that value
- the map is supposed to adapt to the new colors as well

When lowering the "To"-number below the "From"-number:
- the "From"-number should lower as well (just like the slider does).

When raising the "From"-number above the "To"-number:
- the "To"-number should raise as well (just like the slider does).

## Actual Behavior
When I change the numbers inside the input-fields of the colorSettingsPanel:
- the rangeSlider adapts to that value
- the map DOES NOT adapt to the new colors

When lowering the "To"-number below the "From"-number:
- the "From"-number DOES NOT lower as well and just keeps as it is

When raising the "From"-number above the "To"-number:
- the "To"-number DOES NOT raise as well and just keeps as it is

### Screenshots

<img width="393" alt="bildschirmfoto 2018-04-05 um 16 16 12" src="https://user-images.githubusercontent.com/23529226/38371604-2aa5fe76-38ed-11e8-81a5-df806e7ad703.png">

## Steps to Reproduce the Problem

  1. checkout master or refactoring_frontend branch
  2. open the settingsPanel
  3. open the colorSettingsPanel
  4. raise or lower the input-fields

## Specifications

  - Version: 1.10.0
  - Branch: master and frontend_refactoring
  - Platform: ?
  - Subsystem: ?
  - Browser: Chrome and Safari

FernandoES:
Solution developed in branch colorSettingCorrecture

FernandoES:
Solution in Pull Request #132, please have a look and approve/reject modify it

PR #132, opened 2018-04-23
Avoid "From" from being upper than "To" -1 in Color Range

FernandoES:
# Solution for Issue #121


alschmut:
The slider works almost as expected. It just "crashes" with negative numbers. But the map doesn't adapt the new color range. I thought its because you didn't apply the settings as before `this.settingsService.applySettings(this.settingsService.settings);`, but it didn't work either when calling it. Did you experience the same issue?

FernandoES:
It neither adapt the new color for me, both with the numbers and with the sliders. Previously it didn't adapt it with the colors but it did with the slider. I'm looking for a solution now

FernandoES:
When it comes to negative numbers I check that if I set "to" to 0, "from" is set to -1, in the update the lower value for to will be 1, like it is in the slider. It is not possible for me to input negative numbers, neither typing them or clicking in the reduce button. Does it make sense for some metric to have negative values?

FernandoES:
The new push should solve all the above mentioned problems, by:
 - Set upper and lower limits for "from" and "to" values.
 - added option "onChange" to rzslider as an equivalent for ng-change
- the new methods now call to onSliderChange, previously they called to onToSliderChange by a typo

Issue #150, opened 2018-05-29
Hide empty middle packages


mrpatpat:
Original feature request was #119.

## Description

In order to minimize nesting levels in the tree view and map we need to merge empty middle packages into one package. This can be implemented in the data loading and transformation step.

e.g. 
"src"/"main"/"java" (3 levels) becomes "src.main.java" (one level) as long as src, main and java do not have siblings.

mrpatpat:
![screen](https://user-images.githubusercontent.com/2622069/40681020-3ddf45c2-6388-11e8-9a6d-07b1bff6520e.PNG)

Functionality implemented. Only tests missing now

PR #157, opened 2018-05-29
Feature/compact middle package

mrpatpat:
# compact middle package

## Description

This PR implements #150. It compacts emoty middle packages to a single node.

screen

Attributes/Deltas are taken from the last package in the compacted path. Since they all are empty middle packages, they do not have any differences in their values (property of treemaps).

Link is taken from the last package since it fits the resulting path the best. We should show all links in the future but this should not be part of this PR.

## Definition of Done

* [x] **All tests pass**    


FernandoES:
Everything seems good to me, but a consequence that I find funny. I consider it not a failure, but something that we should take if into account.I'll describe the procedure to get it:

I compare using deltas two files that contain the same leaves, but one of them contains one empty middle folder (called middle Parent) and the other doesn't. This triggers a green delta in the leaves of that folders both previously and with this update.
The difference comes when, with active deltas, I go to TreeView and open root and the delta parent of the other middle parent in one case and of the leaves in the other.There I get the following:
Previously:
bildschirmfoto 2018-05-30 um 17 36 25
with this update
bildschirmfoto 2018-05-30 um 17 34 29

mrpatpat:
@FernandoES this is actually a case I did not think about. I think I can fix it by changing the processing order. Delta's should be applied before compacting.

FernandoES:
Nice, I'll merge it

Issue #164, opened 2018-06-01
Automatic margin to better visualize package structure.


Richargh:
## Description

Margin probably needs to be automatic. 

If the area metric is RLoC and the complete tree map has 1.000.000 RLoC the margin needs to be bigger than if the complete tree map has 1.000 RLoC.

For an area of 100 RLoC the margin 20 seems to be good. For an area of 1.000 RLoC the margin of 15 seems to be good. For an area of 1.000.000 RLoC the margin of 10 might be good. This needs to be evaluated while working on this issue. 

## Definition of Done

A task/pull request will not be considered to be complete until all these items can be checked off.

* [ ] **All** requirements mentioned in the issue are implemented
* [ ] Does match the Code of Conduct and the Contribution file 
* [ ] Task has its own **GitHub issue** (something it is solving)
  * [ ] Issue number is **included in the commit messages** for traceability
* [ ] **Update the README.md** with any changes/additions made
* [ ] **Update the CHANGELOG.md** with any changes/additions made
* [ ] **Enough test coverage to ensure that uncovered changes do not break functionality**
* [ ] **All tests pass**    
* [ ] **Descriptive pull request text**, answering:
  + What problem/issue are you fixing?
  + What does this PR implement and how? 
* [ ] **Assign your PR to someone** for a code review
  + This person _will be contacted **first**_ if a bug is introduced into `master`
* [ ] **Manual testing** did not fail

FernandoES:
In order to generalize I suggest (and implemented) to make the margin depending on the areaMetric.

The criteria used is margin = log2(areaMetric), but this is still opened to suggestions and debate.

When the metric used for area is changed the margin is also changed.

There is a failure since the size to map fitting is done before the ultimate margin calculation, so the map is wrong fitted in screen

PR #188, opened 2018-06-30
Automatic margin

FernandoES:
Automatic Margin

Original issue #164 

## Description

Added a feature that computes automatically the margin of a map depending on the mean area of the buildings. 

This feature can be deactivated by deactivating the tick in the display settings panel or by moving the margin slider. The slider deactivates the automatic/dynamic margin computing because otherwise this would reset the margin to the computed one, undoing the value selection of the user.

When activated the dynamic margin 

The margin is four times the square root of the mean area of the buildings of map. 
Why this value/equation? 
The mean area is used because it determines the part of the area required by buildings (almost, how deep are class in terms of how many packets in packets in packets... have also a low influence). 
The square is used in order to allow nice margin values in a long set of area values (it could easily be 1-10 but also 100 000 to 1 000 000)
The four was included by practice by considering it the constant that allows a nicer margin. It is an arbitrary value, but deleting it produces such small margin

## Definition of Done


visualization/app/codeCharta/core/settings/settings.service.ts
```diff
@@ -236,6 +237,63 @@ export class SettingsService implements DataServiceSubscriber, CameraChangeSubsc
...
+        if(this.dataService.data.renderMap != null && this.settings.dynamicMargin){
+            let  root: CodeMapNode = this.dataService.data.renderMap.root;
+
+            // Previous parameters are wrapped to introduce them into goThroughMap() and make value changes
+            // remain, with a pointer-like behaviour
+            let numberOfBuildingsWrapped = {numberOfBuildings: 0};
+            let areaParametersWrapped = {areaMetricName: this.settings.areaMetric, totalArea: 0};
+
+            this.dataService.goThroughMap( root, this.countBuildings, numberOfBuildingsWrapped);
+            this.dataService.goThroughMap( root, this.calculateTotalArea, areaParametersWrapped);
```
alschmut: I am not sure why, but for some projects the properties `numberOfBuildingsWrapped` and `areaParametersWrapped` have the value 0 and the `margin` just uses the default number of 15. For example I tested it with my own `codecharta.json` and `git.json` files, which were imported from the SCMLogParser.

visualization/app/codeCharta/core/data/data.service.ts
```diff
@@ -43,6 +43,28 @@ export class DataService {
...
+     * @param {any[]} parameters
+     *
+     * Function that goes through every node in a map and applies there a function
+     * whose parameters are hold in parameters array.
+     *
+     * The function input in fn is required to get their arguments into a single array, which allows to input
+     * as many variables as desired and also to remain the computed value in all calls to fn that recursive
+     * procedure requires.
+     */
+    public goThroughMap(map: CodeMapNode, fn: Function, parameters: any[]|{}){
```
mrpatpat: You can import "hierarchy" from d3 and use it to achieve the exact same functionality in a more readable way. It also makes this method unnecessary. Take a look at this [line](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/core/data/data.decorator.service.ts#L72)

PR #192, opened 2018-07-16
Feature/oop parser

Richargh:
## Description

This issue solves the detour to always have to set up and configure a SonarQube project, just to import some wanted metrics.

Create a simple selfwritten parser (starting with the rLoC metric, other metrics like mcc will follow in later issues) for an individual java file. 

(!) Internally each file should be translated into a language-independent format. This technical detail is important, because we want to be able to add parsing functionality for future languages like JavaScript, C#, Kotlin with minimal effort. Other languages which might require a different language-independent format are SQL languages and functional languages such as Clojure.

Given the following interface:

`ccsh sourcecodeparser RealLinesWithEmptyLines.java`

then the output should create a readable tabular output in the console like this:

Line | RLoC | Code
---   |---      |---
1      |           | /\*
2      |           | \* Lots of legal jargon
3      |           | \*/
4      |  1       | package none;
5      |           | 
6      |  2       | import foo;
7      |           | 
8      |  3       | @Annotation
9      |  4       | public class Foo {
10     |           | 
11     |           |     /\* This method does nothing\*/
12     |  5       |     public void noop(){
13     |           |           // I need another comment here
14     |           |      }
15     |           |   
16     |           |  }

## Definition of Done

* [x] **All tests pass**    


analysis/import/SourceCodeParser/script/sample-output.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+echo Can only be run after jar is created by "gradle build"
+echo ------------------------------------------------
+echo    FolderOutput
+echo ------------------------------------------------
+java -jar build/libs/codecharta-sourcecodeparser-1.13.0.jar src/test/resources/ -out=table
```
mrpatpat: The version string should change accordingly to the gradle.properties file which is the single point of truth for analysis versioning. Our release script updates only the gradle.properties version number.

analysis/import/SourceCodeParser/src/main/kotlin/de/maibornwolff/codecharta/importer/sourcecodeparser/oop/domain/metrics/OopMetricCalculationStrategy.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class OopMetricCalculationStrategy: MetricCalculationStrategy {
+
+    override fun calculateMetrics(line: Line, previousMetrics: MetricMap): MetricMap {
+        return MetricMap(MetricType.values().map { it to singleMetric(it, line, previousMetrics) }.toMap())
+    }
+
+    private fun singleMetric(metricType: MetricType, line: Line, previousMetrics: MetricMap): Int{
+        return when(metricType){
+            MetricType.LoC -> line.lineNumber
```
mrpatpat: The calculation strategies for metrics (e.g loc or rloc) could be extracted into classes or strategies. I think your abstraction of source code allows reusing these strategies for other programming paradigmas than OOP. More complex metrics would make this when operation unreadable. This is just a note for the future. In my opinion this solution is good enough for this PR.

analysis/import/SourceCodeParser/src/test/java/de/maibornwolff/codecharta/importer/sourcecodeparser/oop/infrastructure/antlr/java/AntlrAbstractSyntaxTreeTest.java
```diff
@@ -0,0 +1,115 @@
+package de.maibornwolff.codecharta.importer.sourcecodeparser.oop.infrastructure.antlr.java;
```
mrpatpat: Is this test really necessary ? As far as I can see it checks Antlr itself and should be out of our testing scope. Is there a reason why this test is needed ?

Issue #172, opened 2018-06-20
Add temporal coupling to ccsh and visualization


Richargh:
parent of #210, parent of #211, parent of #212, parent of #213, parent of #214, parent of #215, parent of #216, parent of #217, parent of #218, parent of #219

Use [feature/172/visualizing-edges](https://github.com/MaibornWolff/codecharta/tree/feature/172/visualizing-edges) to checkout a branch.

## Description
1. Build an Importer for the SCM metric tool CodeMaat https://github.com/adamtornhill/code-maat. Check out http://www.adamtornhill.com/code/crimescenetools.htm to download an executable codemaat.jar file.

    ![codemaatimport-modell](https://user-images.githubusercontent.com/23529226/45157514-4ba62780-b1d9-11e8-801f-fd79cf40d072.jpg)

    The command takes a CSV table with the coupling data provided by CodeMaat and returns a cc.json schema conform JSON file. The `coupling.cc.json`-file will not have any Treemap-Metric data and therefor needs to get merged with another file, which contains usual metric data for every file.

    `ccsh codemaatimport coupling.csv -o coupling.cc.json`

2. Visualize the temporal coupling data inside the CodeMap. Other Issues how to implement in concrete terms follow.

## Issues still to create:
- by @ukinimod: Personally I would map the pairing rate to the colour. I find this more intuitive than the height, although the height could be better to keep the overview. Nevertheless I don`t figure displaying hierarchy, coupling and three metrics at the same time is sensible in any way (yes, I'm always for a minimal feature set), therefore this feature could imho be deleted.
- by @richargh: It might also be interesting to see coupling between large packages. No clue how to visualize that however.
- by @richargh: I wonder if it makes sense to color the buildings that are paired together in the same color. In addition to the arrows. Or maybe give them "hats" that are the same color.
- by @ukinimod: It's probably only vaguely related to this pr, but I guess this context is good enough to answer an offline question of @Richargh.

    what I and some collegues, including @silasg, would find very interesting (mostly in the analysis use case) to be able to display the (especially temporal) coupling for certain hotspots. I'm not quite able to put it in good words, but I would describe it as the degree the "foulness" of this class / file is spread across the project.

    One possible visualization that we discussed for this, is a way to select the file, that is considered as hotspot. Then something like the pairing rate for all (temporal) coupled (with a possibly configurable limit) are coloured by their degree (e.g. pairing rate) of dependency.

alschmut:
[...]

Issue #182, opened 2018-06-25
Place to collect additions to the cc.json


Richargh:
Please add data that you consider adding to the followng bullet point list:

* an id for each "File" so can realibly compare files that have been moved (data might come from version control system)
* user settings (camera angle, position, selected metrics)
* metric description (long text that describes what each metric means, will be created by the corresponding importer)
* markingColor for highlighted packages, marked packages, pinned buildings
* exclude tag, which can treat the node, at least by default, as non-existant #181 
* different types of dependencies/edges

## Current cc.json top-structure
```json
{
    "projectName": "codecharta",
    "nodes": [],
    "apiVersion": "1.0",
    "edges": []
}
```

Richargh:
@alschmut can you explain your critique in a different way? I don't understand it.

Richargh:
New cc.json idea. Look out for all the comments with `// NEW`

```json
{
    "projectName": "rojectName",
    "apiVersion": "1.0",
    "createdOn": "9.8.2018", // NEW some form of timestamp with date and time
    "nodes": [{
        "name": "root",
        "type": "Folder",
        "attributes": {},
        "link": "",
        "children": [{
            "name": "src",
            "type": "Folder",
            "attributes": {},
            "link": "",
            "children": [{
                "name": "de",
                "type": "Folder",
                "attributes": {},
                "link": "",
                "color": 1, // NEW, can be a specific id, that maps to our default color
                // OR a string with a hex code "color": "#123"
                "children": [{
                    "name": "Foo.java",
                    "type": "File",
                    "attributes": {
                        "mcc": 39,
                        "lines_of_code": 212,
                        "rloc": 182
                    },
                    "fractionAttributes": {
                        "coverage": 100 // NEW
                    },
                    "link": "",
                    "children": [{
                        "name": "maibornwolff",
                        "type": "Folder",
                        "attributes": {},
                        "link": "",
                        "children": [{
                            "name": "codecharta",
                            "type": "Folder",
                            "attributes": {},
                            "link": "",
                            "children": [{
                                "name": "Bar.java",
                                "type": "File",
                                "attributes": {
                                    "mcc": 39,
                                    "lines_of_code": 212,
                                    "rloc": 182
                                },
                                "fractionAttributes": {
                                    "coverage": 50 // NEW
                                },
                                "link": "",
                                "children": []
                            }]
                        },
                        {
                            "name": "Blub.java",
                            "type": "File",
                            "attributes": {
[...]

PR #194, opened 2018-07-24
Feature/code maat temporal coupling import

alschmut:
# Feature/code maat temporal coupling import

## Description

Implementing #172 and updating the [cc.json](https://github.com/MaibornWolff/codecharta/blob/feature/code-maat-tc-importer/analysis/tools/ValidationTool/src/main/resources/cc.json) schema.

The new CodeMaatCouplingImport Module imports CSV temporal coupling data from CodeMaat and returns a valid cc.json file with only dependencies (no treemap nodes).

Check out the [README.md](https://github.com/MaibornWolff/codecharta/blob/feature/code-maat-tc-importer/analysis/import/CodeMaatCouplingImporter/README.md) for more detailed information about the usage.

To merge the coupling data with the normal metrics cc.json files the MergeFilter was updated as well and will merge the dependency-lists when using the `--recursive` strategy.

### Questions:
- Would you like to change the sub-command-name `ccsh codemaat-couplingimport ...`?

### Further changes:
- The mutable objects inside the `Project.kt` interface and `ProjectBuilder.kt` still needs to become immutable objects. I would do this in a future request.
- The Pull Request for visualizing the coupling data will follow

## Definition of Done

* [x] **All tests pass**    


alschmut:
Thanks for your detailed review @Richargh! 
1. calling it `ccsh codemaatimport` with the option `--import coupling` seems fine
2. The error occurs, because the old dependency object was an array and now is an object of possibly multiple dependency types. If we always only add and never change the existing schema it will be backwards compatible. So you want to add an `edges` object and keep the old `dependency` object?
3. Isn't your edges-list too generic? What if we decide to add different dependency types later? How would you like to add the ingoing and outgoing dependencies? Your suggestion only contains one list of all "edge-objects".
4. But I quite like the idea to use attributes as well here. The `fractionAttributes` is fine for me as well. We could also do that inside the node objects.
```
{
    "fromNodeName": "/root/codecharta/Foo.java",
    "toNodeName": "/root/codecharta/Bar.java",
    "attributes": {
        "averageRevs": 9
    },
    "fractionAttributes": {
        "pairingRate": 100
    }
}
```
5. Yes, I will remove the (dependant)nodeFilename.
6. For the general cc.json schema we continue working on #182 

So for now, as soon as we decide for a final schema, I will work on the cc.json and codemaatimport before merging, correct?

Richargh:
@alschmut 
1. I don't think we need the `import` flag right now. We might add it at a later point and make --import coupling the default then.
2. Is there any way to deprecate stuff in the cc.json schema? I prefer to deprecate dependencies and add edges, yes. 
3. edges might be too generic. That's why I'd like to discuss the change with as many people as possible :) Regarding your comments: Can you think of dependency types that you cannot describe using attributes and fractionAttributes? I think that is flexible enough to describe the *weight* of an edge but maybe it's not? You can describe ingoing and outgoing dependencies by creating two edges. One `{"fromNodeName":"Foo.java", "toNodeName": "Bar.java", "attributes":"..."}` and reverse `{"fromNodeName":"Bar.java", "toNodeName": "Java.java", "attributes":"..."}`. I just checked the DOT language and how they describe edges. Basically each edge gets an direction, a style, a color, a label. And that's it. 
6. I'll update #182 with a new suggestion.

Can you try out the edges format in development? Perhaps even add attributes that you do not currently need. In what other way can a node depend on another node besides temporal? Structure101 connect classes using edges. An edge can mean that a class subclasses,implements another class/interface. Or that a class calls a method on another class. Or that a method of one class returns a different class. Or that a class imports a namespace. Just expand the cc.json with some fake data and see what happens :) **Maybe you'll find that during development that the edges suggestion is not good enough.** That would also be great feedback. One thing I wonder about is if it makes sense to group certain attributes together. Right now `"averageRevs": 9` and `"pairingRate": 100` have no semantic relationship and I wonder if that is an issue.

alschmut:
I have opened new issues from all your comments to work on those before merging the new branch [feature/172/visualizing-edges](https://github.com/MaibornWolff/codecharta/tree/feature/172/visualizing-edges) into master

Issue #181, opened 2018-06-25
A filter to remove unneeded parts from the cc.json


Richargh:
## Context

Source code sometimes contains vendor code (angular, hibernate, etc.) which shows up in the cc.json and then in the visualization. We do not need that code, because it is beyond our scope and frankly out of our control. 

## Description

We want to permanently remove vendor code from our cc.json, even if it is placed on some form of sublayer. 

The interface should be the one suggested by @ukinimod below:

`ccsh filter -exclude "src/main/de/**/*Test.java" test.cc.json > filtered_test.cc.json`

ukinimod:
What would be practical for me is an integration in ccsh via regex (e.g. ant-style regex), e.g. via
ccsh filter -exclude "src/main/de/**/*Test.java" test.cc.json > filtered_test.cc.json
or 
ccsh filter -f filter.txt test.cc.json > filtered_test.cc.json

Consequently, these filtered nodes should be treated, at least by default, as non-existant in the visualization.

Usually I try to already filter out these files with the analysis tools, e.g. understand, by setting the filters there, but in case we no longer have the sources it is not possible to redo the analysis with the correct filters. Therefor it's probably a non-standard use-case for me.

mrpatpat:
I like this feature. What do you think about a flag for something like a soft and hard filter? The hard filter removes the nodes completely from the file and the soft filter just presets the visibility.

alschmut:
@mrpatpat I like the idea of a flag!

ukinimod:
@mrpatpat I personally have no need for a soft filter in my analysis. If needed one could use the filter in the visualization. If someone needs the soft fiter it could maybe implemented later?

Richargh:
Reopenend because there are many cases where items should not be in the cc.json at all. node_modules, vendor librararies or similar significantly slow down the `ccsh mergefilter`. To combat this issue we want to permantenly remove them from the cc.json.

PR #201, opened 2018-08-06
Feature/add temporal coupling in visualization

alschmut:
# Feature/add temporal coupling in visualization

## Description

This PR adds the option to visualize the temporal coupling data, which can be imported with the new CodeMaat-CouplingImport.

1. There is a new Temporal Coupling Panel with a list of the dependency couples in the settings sidebar
2. Every Couple has a number of shared commits and a relative number "Pairingrate", which gives you information about the percentage of coupling.
    - When hovering the list item a tooltip/title shows expanded information and the full filepath
3. To adjust the list with Couples there are different settings:
    - Slider: set minimum number of shared commits
    - Checkbox: Use Pairingrate Height (or use metric from Metric Chooser)
    - Checkbox: "Intelligent" Filter to remove couples, like `*.log`, `build.gradle` or `CHANGELOG*` etc. (This Filter has a lot of potential for future improvements)
    - Checkbox: Show Connections turns the connecting arrows on and off
4. The TC list always syncs with the CodeMap and only shows the couples, when both nodes are visible inside the CodeMap
5. To try out the visualization without having to import the couples before, two test files ([codecharta.json](https://github.com/MaibornWolff/codecharta/blob/feature/add-temporal-coupling-visualization/analysis/test/data/codecharta/codecharta.json) and [git.json](https://github.com/MaibornWolff/codecharta/blob/feature/add-temporal-coupling-visualization/analysis/test/data/codecharta/git.json)) are added inside the directory `/analysis/test/data/codecharta/...`

- The [cc.json](https://github.com/MaibornWolff/codecharta/blob/feature/add-temporal-coupling-visualization/visualization/app/codeCharta/core/data/schema.json) schema of course needed to change. It is now the same as the one inside the Validation Tool! The `/codecharta/cc.json` file has been removed, because of duplicated file content.
- I updated a lot of tests with its test json data and usually only added `type` and `dependencies`

bildschirmfoto 2018-08-06 um 13 19 02

bildschirmfoto 2018-08-06 um 13 02 28

## Definition of Done

* [ ] **All** requirements mentioned in the issue are implemented
* [ ] Task has its own **GitHub issue** (something it is solving)
* [ ] **All tests pass**    


ukinimod:
I will do a more systematic code review later, first some comments:
- I like the ability to be able to visualize temporal coupling (or coupling in general) in codecharta.
- I like the temporal coupling tab in the sense that it lists the hotspot dependencies. I can figure using this in audits to find strange dependencies and highlight them in a presentation.
- I guess this feature is not compatible with aggregation and delta.
- Filtering unnecessary files could be done in a different way, e.g. filtering when generating the log, filtering in the csv, filtering the uninteresting nodes and only show dependencies for visible nodes. I think adding the "intelligent filter" with hardcoded values could be removed before merging.
- Personally I would map the pairing rate to the colour. I find this more intuitive than the height. Furthermore I don`t figure displaying hierarchy, coupling and three metrics at the same time is really necessary (yes, I'm always for a minimal necessary feature set), therefore this feature could imho be deleted.
- average-revs sind NICHT die gemeinsamen Commits sondern der Durchschnitt der Commits.

Detailed review later...

alschmut:
Thanks as well for your feedback!

3. At the moment it is not compatible with aggregation and delta. Aggregation can be implemented as a simple aggregation of both dependency lists, but to get dependencies with nodes from different original projects, we need to try to combine the log files and analyze those. Although files from different projects will (I think so) never be in the same commit. But CodeMaat also allows the connection of files by the day (when the commit was made) instead of the commit itself.
4. Okey I agree with the filter. This will then be implemented with #181.
5. The problem with using the normal chosen metric is, that some buildings might be very small and the User can't see them on first sight. But to mention that: I also thought about the idea to color all other buildings in a light gray. Then the colored coupled buildings get highlighted again even more. @ukinimod sorry, I couldn't follow which feature you wanted to be deleted.
6. Oh, so i misunderstood that, thanks.

alschmut:
7. Yes it filters the list by AverageRevs.
8. Okay, it actually were nice to put the new `edges.attributes[]` inside the existing metric-chooser. Maybe at the top or bottom of the select-list and marked as edge-metric.
9. I have added the "intelligent filter", because sometimes many log and gradle files occurred in the top 10 couples, but as @ukinimod said, we should better exclude files inside the analysis.

10/11/12/13. Yes, we can implement something like this to improve usability afterwards :)

ukinimod:
It's probably only vaguely related to this pr, but I guess this context is good enough.

what I, and some collegues talked about, would find very interesting (mostly in the analysis use case) to be able to display the (especially temporal) coupling for certain hotspots. I'm not quite able to put it in good words, but I would describe it as the degree the "foulness" of this class / file is spread across the project. 

One possible visualization that we discussed for this, is a way to select the file, that is considered as hotspot. Then something like the pairing rate for all (temporal) coupled (with a possibly configurable limit) are coloured by their degree (e.g. pairing rate) of dependency.

alschmut:
I have opened new issues from all your comments to work on those before merging into master

PR #260, opened 2018-10-31
PR for "Buildings in the delta view are not colored correctly #253"

mrpatpat:
## Description

This PR triggers the bug by testing for negative values and fixes it. There was an unnecessary Math.abs() call which made the value always positive. When deltas are inverted it used red for positive.

## Definition of Done

A task/pull request will not be considered to be complete until all these items can be checked off.

* [x] **All** requirements mentioned in the issue are implemented
* [x] Does match the Code of Conduct and the Contribution file 
* [x] Task has its own **GitHub issue** (something it is solving)
  * [x] Issue number is **included in the commit messages** for traceability
* [ ] **Update the README.md** with any changes/additions made
* [x] **Update the CHANGELOG.md** with any changes/additions made
* [x] **Enough test coverage to ensure that uncovered changes do not break functionality**
* [x] **All tests pass**    
* [x] **Descriptive pull request text**, answering:
  + What problem/issue are you fixing?
  + What does this PR implement and how? 
* [x] **Assign your PR to someone** for a code review
  + This person _will be contacted **first**_ if a bug is introduced into `master`
* [x] **Manual testing** did not fail

Issue #254, opened 2018-10-30
Show deleted buildings in delta view


Richargh:
In the delta view two cc.json (`file` and `compare to`) can be compared. If a building existed in `compare to` but no longer in `file` it is not visible in the visualization, if I select a metric like *rloc*. This needs to change.

**Given** a building exists in  `compare to ` but no longer in `file`
**and** I open the delta view
**When** I select the metric *rloc* for size and *mcc* for height
**Then** I see a red building (or green if colors are inverted) and use the metrics from `compare to`.

alschmut:
So with other words: For deleted files in delta view, use the metrics from the deleted file and color it by default red

Richargh:
Yes. Also, that is a great and uncomplicated summary :)

PR #261, opened 2018-10-31
Feature/254/make deleted buildings visible

mrpatpat:
## Description

Solves #254. Connects #254

Checks if a building in delta mode existed in a previous revision and was deleted in the currently viewed. If yes it creates an artificial area value for this node corresponding to its previous area value (equal to current absolute area delta). 
Please note that this area value is only artificial and only visible in the 3D view. This means that the detail view and other metric displays still show an area of 0 with a negative delta which is the truth for the currently rendered map (e.g. the building has 0 rloc as the area metric in the current file but the visualized area is the area of its previous revision).

We can only assume file deletions therefore a deleted node is defined as a node which has:

- another origin than the current map
- a current area of 0
- a current area delta smaller than 0

The building is red by default since it has a negative delta value. There is a bug which made it green, so I already merged the fix #260 into this branch.

Needs #260

## Definition of Done

* [ ] **Update the README.md** with any changes/additions made
* [x] **All tests pass**    


FernandoES:
I found a little problem. Do the following to reproduce it:
- load the website and change no option
- select sample2.json
- You get this:
bildschirmfoto 2018-11-01 um 10 29 53
- Then select Delta view and in the option compare to, select sample1. Then go back to Single view, you should get the same as in the previous picture, since you are viewing the same, but you get this:
bildschirmfoto 2018-11-01 um 10 33 52
It is not only ordered other way, which should not be a problem (imo) but you also have a building more, the one selected "sample1 only leaf", which only exist in sample1.

This building should remain in the Delta view, that is the point of this issue, but not in the Single view. 

Note: If it should also remain in Single view, I missmatched the issue and then is no problem :)

PR #270, opened 2018-11-06
Feature/268/add isolate type to blacklist

### LINKED ISSUE #268: Store Isolated Node in Blacklist
# Feature Request

The reason to go for this feature was actually the bug of hiding/showing nodes within the TreeView and  the blacklist. Instead of doing a workaround for the bug I decided to create this feature request.

### Description

When isolating a node we had a conflict between:
- changing the current visibility state of all nodes and 
- setting up a completely new state (like we do since using the blacklist)

So to be able to also isolate a node, we could store that node inside the blacklist as well. It is not possible to isolate more than one node (as that doesn't make sense). When rendering the new map, it now can read from the blacklist if and which node is supposed to be isolated.

Isolating a node does following:
1. keep the excluded nodes excluded
2. keep the hidden nodes hidden
3. only show the isolated nodes with its visible children

Good side-effect:
When the user wants to download that map for a later point, it also saves the isolated node and shows it just as the user left it.

### Question
Do you think **blacklist** is still the correct word and if not do you have a better idea?

alschmut:
# Feature/268/add isolate type to blacklist

Connects #268 Closes #268

### Added
- Isolated node shows up in the blacklist
- NodeContextMenu: Option to Unisolate node in TreeView
- NodeContextMenu: Option to only hide dependent edges

 ### Changed
- Isolating a node does not remove the blacklist items of type Hide

### Question
- Do you think **blacklist** is still the correct word and if not do you have a better idea?

## Definition of Done

* [x] **All tests pass**    

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased]
 ### Added
+- Isolate node shows up in the blacklist
+- Option to Unisolate node from NodeContextMenu in TreeView
+- Option to only hide dependent edges
 
 ### Changed
+- Isolating a node does not remove the blacklist items of type Hide
 
 ### Removed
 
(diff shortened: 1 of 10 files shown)


Richargh:
I am confused by this feature. Is it supposed to be in "Review" or is it a remnant of #268 that should be closed as well?

PR #335, opened 2019-01-28
Tech/224/scenario as update object

### LINKED ISSUE #224: Scenarios should only be about metrics
# Feature request
As an auditor, I want scenarios that help me select important metric combinations in one click, so I can quickly switch between different views.

Currently the scenarios reset everything, including weird things such as margin. That is confusing and does not reflect what I want. 

To be usable the scenarios should only be about metrics. The coverage scenario, for example, should only set the three metrics and the color range. A building with >= 80% coverage should be green. A building with >= 20% coverage should be yellow. All other buildings red.

alschmut:
# Tech/224/scenario as update object

closes #224 
Merge permission: @mrpatpat, @alschmut 

## Description

- Scenarios will only have to hold necessary settings, as it will no longer update all settings, but just the settings, which are mentioned inside the scenario
- Tech: That also means, when adding an attribute to `settingsService.settings`, we no longer have to update the (Default)-Scenario object (which is loaded initially) to not loose the value of the new settings-attribute.
- Different from the issue #224: Scenarios are still able to override other settings attributes and not *only metrics*. That makes sense for example for `neutralColorRange`, when the metric `coverage` is used and the `Invert Colors` actually makes sense. Or using predefined range-boundaries for the percentage of coverage.

## Definition of Done

* [x] **All tests pass**    


visualization/app/codeCharta/core/scenario/scenario.service.ts
```diff
@@ -52,10 +49,23 @@ export class ScenarioService {
         }
     }
 
-    /**
-     * Returns an array of all scenarios.
-     * @returns {Scenario[]} all scenarios
-     */
+    private updateSettingsUsingScenario(settings, scenarioSettings): any {
```
mrpatpat: try using spread operator

visualization/app/codeCharta/core/scenario/scenario.data.ts
```diff
@@ -3,41 +3,21 @@ import {CodeMap} from "../data/model/CodeMap";
...
-            invertHeight: false,
-            dynamicMargin: true,
-            isWhiteBackground: false,
-            blacklist:[],
-            searchPattern: ""
+            colorMetric: "mcc"
         },
         autoFitCamera: true
-    };
+    } as Scenario;
```
mrpatpat: use Partial type for settings

visualization/app/codeCharta/core/scenario/scenario.service.ts
alschmut: Works perfectly. It even works in nested levels 👍

visualization/app/codeCharta/core/scenario/scenario.data.ts
alschmut: Implemented: see further commits.

PR #332, opened 2019-01-27
Bug/330/dynamic-color-range

alschmut:
# Bug/330/dynamic-color-range

closes #330 
depends on #338 
Merge permission: CC-Member

## Description

- This PR includes the merged branch [bug/reset-button-used-default-scenario](
https://github.com/MaibornWolff/codecharta/tree/bug/reset-button-used-default-scenario) as the `SettingsService.getDefaultSettings()` was needed
- adapt colorRange when changing colorMetric or onload
- the range is cut in thirds
- without having to change the range manually, it adapts automatically
- Deleted unused displaySettings component
- Use only `settingsService.applySettings()` instead of `settingsService.onSettingsChanged()`
- Make `settingsService.onSettingsChanged()` private

## Definition of Done

A task/pull request will not be considered to be complete until all these items can be checked off.

* [ ] **All** requirements mentioned in the issue are implemented
* [ ] Does match the Code of Conduct and the Contribution file 
* [ ] Task has its own **GitHub issue** (something it is solving)
  * [ ] Issue number is **included in the commit messages** for traceability
* [ ] **Update the README.md** with any changes/additions made
* [ ] **Update the CHANGELOG.md** with any changes/additions made
* [ ] **Enough test coverage to ensure that uncovered changes do not break functionality**
* [ ] **All tests pass**    
* [ ] **Descriptive pull request text**, answering:
  + What problem/issue are you fixing?
  + What does this PR implement and how? 
* [ ] **Assign your PR to someone** for a code review
  + This person _will be contacted **first**_ if a bug is introduced into `master`
* [ ] **Manual testing** did not fail

PR #338, opened 2019-02-02
Bug/reset-button-used-default-scenario

alschmut:
# Bug/reset-button-used-default-scenario

Merge permission: CC-Member

## Description

While refactoring the Scenario-Service, I didn't see, that the resetButtons use the defaultScenario, which doesn't hold all the settings anymore. So for some settings it return `null` and applied that to the settings.

- This PR renames `settingService.getInitialSettings(...)` into `getDefaultSettings(...)`
- ResetButton component now uses this `getDefaultSettings(...)` to overwrite the settings
- resetSettingsButton.component triggers `onSettingsChanged()` instead of `applySettings()`
- Refactor `SettingsService.computeMargin()`
- Update and refactor `settings.spec.ts`

## Definition of Done

* [ ] **All** requirements mentioned in the issue are implemented
* [ ] Does match the Code of Conduct and the Contribution file 
* [ ] Task has its own **GitHub issue** (something it is solving)
* [ ] **Update the README.md** with any changes/additions made
* [ ] **Update the CHANGELOG.md** with any changes/additions made
* [ ] **All tests pass**    
* [ ] **Descriptive pull request text**, answering:
* [ ] **Manual testing** did not fail


visualization/app/codeCharta/core/settings/settings.service.ts
```diff
@@ -85,19 +85,19 @@ export class SettingsService implements DataServiceSubscriber, CameraChangeSubsc
         });
     }
 
-    private getInitialSettings(renderMap: any, metrics: string[]): Settings {
+    public getDefaultSettings(renderMap: any, metrics: string[]): Settings {
```
mrpatpat: it propably was like this before but is it possible to type the map correctly ?

alschmut:
> When I click the size reset button something weird happens. The slider moves to 15 and then back to the correct value. This seems to be a visual bug only. Otherwise this feature looks good.

This regards to the dynamic margin slider, doesn't it? Yea I've seen it, but actually thought it was like this before. Just checked the released version and it doesn't do it there. I gonna have another look, but probably at the end of the week.

visualization/app/codeCharta/core/settings/settings.service.ts
alschmut: Good point, I'll check it.

alschmut:
This PR was merged into #332

Issue #178, opened 2018-06-22
Tech: Centralize all user settings in single service that also handles cascading.


Richargh:
# Feature Request

## Description

This is a tech story with no value for the user, but still important for future features. 

The single service centralizes all settings a user can set. Modules are notified of changes using something like rx.js. 

Cascading means that settings can overwrite each other if they are from different sources. F.ex. the default settings can be overwritten by url parameters.

## TODO
* [x] map is immutable, take map out of settings, markingPackage in settings instead in map
* [x] Simplify settings to only handle getting and updating settings from partial settings objects
* [x] Updated `settings` object, which uses three sub-objects. See interface [codeCharta.model.ts#L32](https://github.com/MaibornWolff/codecharta/blob/feature/178/centralize-settings/visualization/app/codeCharta/codeCharta.model.ts#L32)
* [x] Create `codeCharta.model.ts` file which holds all globally needed interfaces
* [x] Create `codeCharta.service.ts` file which combines simplified functionality from former `codeCharta.component`, `data.service`, `data.validator.service`, `data.loading.service`. The `codeCharta.service.ts` get called from `codeCharta.component` and is supposed to be the global connection to the core functionality like onLoadApplication
* [x] Deleted `data.service`, `data.validator.service`, `data.loading.service`
* [x] Implement ScenarioService behaviour onLoad and onFileImport(?)
* [x] Merge fileSettings from visible `CCFiles` into `settingsService.settings.fileSettings`
* [x] Check if #352 is fixed
* [x] Implement UrlUtils which updates the settings. Don't forget about the `mode=Delta`
* [x] Update content of the dialog when clicking `URL Parameters` button inside sidebar. Only show file and mode keys
* [x] Remove `URL Parameters` functionality from `UrlLUtils` and `SettingsService`, but keep option to import files and set a `mode` (renderState) from within the url
* [x] Adapt the new settings/fileState/metric behaviour step by step to the UI components
* [x] Adapt all tests and try to follow the scheme for setting all objects like [nodeContextMenu.component.spec.ts](https://github.com/MaibornWolff/codecharta/blob/feature/178/centralize-settings/visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.spec.ts)
* [x] Everywhere -> Remove unused injections
* [x] scenarioHelper -> Refactor getScenarioByName array[string]
* [x] Tests -> update skipped Tests 
* [x] settingsService -> onCameraChanged
[...]

PR #347, opened 2019-02-13
Feature/337/markingcolor in settings

### LINKED ISSUE #337: Extract markingPackage from CodeMapNode and save in new settingsObject
# Feature request

relies on #178 

## Description

We want the `CodeMapNode` itself to be **immutable**, to have only one Object to update (which is the settingsObject). So we have to do:
- Save the markedPackages in the settings
- DecorateNodes while generating the renderMap with `node.markingColor`

An example of a `markingColor` object (which is the current name for the attribute on the `CodeMapNode`) could be:

Please comment for suggested changes! 
- Can somebody think of a better word than `name`?
- The optional `name ` could be used for replacing the complex-path inside the legendPanel. Of course when selecting a packageColor, the user needs to set that value manually.
- We could also use references instead of the hex color.

```
"coloredPackage": [
  {
    "path": "/root/dir/test/*",
    "color": "0xFF1D8E",
    "attributes": {
      "name": "Test"
  }
]
```

Should the `color` be inside or outside the `attributes` object?

alschmut:
# Feature/337/markingcolor in settings

closes #337 
Merge permission: CC-Member

## Description

- Add colorService to handle color convertions and colorPixelGeneration
- Add object `settings.markedPackages`
- Instead of reading and writing the markingColor into the map, we now have a separate settings-object which only contains a list of unique top-level markedPackages
- When rendering the codeMap it's reading from those `settings.markedPackages` and colors the packages
- The legendPanel is reading from the `settings.markedPackages` and combines items with the same color
- When removing a markingColor form a package
    - it would first remove the color from the currently focused building, if it's the top-level markedPackage
    - if the currently focused building just adopts the color from a parent, it removes the upper parents markingColor
- The `settings.markedPackage` is not yet exported when downloading the currently used map. I think we will figure that out with #178 
- We store the hexColor `#FFFFFF` inside the settings. We might change those into variables in the future or already with #178 to be able to using color-sets

## Definition of Done

* [ ] **Update the CHANGELOG.md** with any changes/additions made
* [ ] **All tests pass**    


visualization/app/codeCharta/ui/codeMap/rendering/geometryGenerator.ts
```diff
@@ -176,11 +188,13 @@ export class GeometryGenerator {
             uvs[i * uvDimension + 0] = data.uvs[i].x;
             uvs[i * uvDimension + 1] = data.uvs[i].y;
 
-            let color: THREE.Vector3 = RenderingUtil.colorToVec3(data.colors[i]);
+            if (data.colors[i]) {
```
alschmut: @mrpatpat Do you know if the `if` statement could cause any problems? Just in the tests, there was given a `colors`-array with 12 hex color-values and 12 undefined/null values... Could you please check that as well? Just remove the `if`-statement and you'll see which test fails..

alschmut: Solved

PR #399, opened 2019-02-26
Bug/398/api Version Missing in Aggregated Map

### LINKED ISSUE #398: Downloaded aggregated cc.json can't be loaded again while importing
# Bug
Downloaded aggregated cc.json-files can't be loaded again when trying to import

## Expected Behavior
Aggregated codemap should be loaded without an error.

## Actual Behavior
Error message. File can't be loaded at all.

  1. Open multiple files
  2. Download the aggregated version
  3. Try to open the aggregated cc.json again

alexhunziker:
# Added Missing apiVersion in Aggregated Map

closes #398 
Merge permission: {CC-Member | @member}

## Description

- Added apiVersion to aggregatedMap (the generated cc.json was invalid without this and could therefore not be re-imported once downloaded)
- Added apiVersion to package.json as a custom field
- Updated snapshots

## Definition of Done

* [x] **All tests pass**    


visualization/app/codeCharta/core/multipleFile/multipleFile.service.ts
```diff
@@ -91,6 +91,7 @@ export class MultipleFileService {
 		let outputMap: CodeMap = {
 			projectName: "Aggregation of following projects: " + this.projectNameArray.join(", "),
 			fileName: "Aggregation of following files: " + this.fileNameArray.join(", "),
+			apiVersion: "1.0",
```
alschmut: I think this is a bit more complex, as we should definitely check the versions of both aggregated maps. What should happen if they have different api versions?

alexhunziker: That is a valid question. If we allow to import maps with different apiVersions at the same time, what is the expected apiVersion of the aggregate, if downloaded? Are we planning to update the maps to the newer version and will we even allow to download an aggregate in this case?

Since these questions might not be trivial to answer, I just implemented this - admitedly simplistic - fix to make sure the download is a valid cc.json file, that can be imported to codecharta again.
But if we have answers to the above questions I agree implementing a target solution is better than this fix.

alschmut: if we can aggregate the maps without errors, we should allow the option to download it. I think this means: 
- Selected maps might have different apiVersions (we might want to handle that in the future), 
- but (I guess) we always create a new valid `codeMap`/cc.json, which is using the latest apiVersion. 
- The question is just: where to get the latest apiVersion from.. [Here](https://github.com/MaibornWolff/codecharta/commit/c274ce1e3f4103ac94874fed58bbd26ecac0f52a) you see, where the apiVersion is persisted.

alexhunziker: yes i agree, maps that can be aggregated should be downloadable.

so, if i understand correctly, the target solution is to take the current codeCharta version number as the apiVersion (regardless of the apiVersion of the aggregated maps), since this will always lead to a valid .cc.json. That sounds reasonable, i will implement it like this.

alschmut: yes, that's what I thought, because we will create a new map (if possible) nevertheless. And if thats possible, it gets the current apiVersion

visualization/package.json
```diff
@@ -29,6 +29,9 @@
     "tsconfig.webpack.json",
     "cli.js"
   ],
+	"de_maibornwolff": {
```
alschmut: Why is the object called `de_maibornwolff`...? Isn't it rather something like `cc_json_file`, although I'm not super happy with that as well. What do you think?

alexhunziker: I put de_maibornwolff there since it was suggested here https://stackoverflow.com/questions/10065564/add-custom-metadata-or-config-to-package-json-is-it-valid that only one top-level key should be added, in which the custom keys are nested. However i have no strong opinion regarding this; we can also call it differently.

alexhunziker: do you think cc_json_file is better?

alschmut: I actually don't know..

alschmut: We decided to use the name `"codecharta" : { "apiVersion": "..." }`

Issue #365, opened 2019-02-18
Arithmetic Median for relative values


Richargh:
# Feature Request
If a metric is relative the arithmetic median for the numbers should be shown when hovering over folders instead of summing them. Also show an arithmetic median symbol.

**When** I hover over a single building
**Then** the metric wihout any symbol is shown

**Given** I have a cc.json that contains relative attribute types
**When** I hover over a package
**and** the metric selected is a relative one
**Then** the airthmetic median for that metric with the correct symbol should be shown.

## Calculating the Median

If you have three numbers 2,3,2.000 the median is 3. If you have an event amount of numbers it's apparanetly the sum of the two middle numbers, divided by two. At least that is my understanding, please confirm from a different source as well.

This [list of statistical symbols](https://www.rapidtables.com/math/symbols/Statistical_Symbols.html) suggests that the median symbol looks like a wave above the number.

NearW:
We don't create attributeTypes in the cc.json right now, which should give us information about whether a metric value should be interpreted as a relative or absolute one. This needs to be implemented before or while working on this issue

NearW:
We will still have problems for metrics that depend on rloc.
For example: I have a folder with 2 files. fileA has 1.000.000 lines of code and fileB has 20. fileA has a test-coverage of 100% and fileB has 0%. We'd display 50% even though this value is not representive for that folder.

Richargh:
> We'd display 50% even though this value is not representive for that folder.
Well it's correct fromt the point of view of the mean. But, yeah it's not representative. Do you have a different average idea? The one I know is the median which is: 

> generally used for skewed distributions
> A median can be computed by listing all numbers in ascending order and then locating the number in the centre of that distribution.

If you have three numbers 2,3,2.000 the median is 3. If you have an event amount of numbers it's apparanetly the sum of the two middle numbers, divided by two. For your example it's still 50% coverage. I see no way around that besides also displaying a deviation. Do you have another idea?

alschmut:
We could use this awesome lighting `fa-icon` to point out that the value might not be representative (when the deviation is too high. Well when is it too high..?). And on hover we can show the deviation value. Okey, I admit its more like a joke, but an idea 😄 
<img width="40%" src="https://user-images.githubusercontent.com/23529226/59531874-1274b380-8edf-11e9-8d4c-63c64cffcdb3.jpeg">

Richargh:
[...]

PR #549, opened 2019-05-28
display sigma letter next to metric values

### LINKED ISSUE #364: Show the sum symbol next to the metric sum
Currently CC sums the metrics for the buildings, when hovering over folders. It would be nice to also see a sum symbol to know what it does.

NearW:
# sigma letter next to metric values

closes #364

## Description
- letter sigma displayed left of the metric value (not a fontawesome-icon)
- Fixed wrong interface of AttributeTypes
- AttributeTypes will no be set to `{nodes: [], edges: []}` instead of `{}`
- AttributeType are no longer 0 and 1. The enum references strings "absolute" and "relative" now

Boyscout:
- Fixed bug and added e2e-test, that ensures metric sum is shown for folders

## Definition of Done

- [x] **All tests pass**


alschmut:
@Richargh were you thinking about two different symbol?. One for an absolute sum and one for relative percentage values (e.x. metric: test-coverage). This data could potentially be stored in `fileSettings.attributeTypes` (although only the codemaatimporter provides that data)

alexhunziker:
grafik

I prefer a smaller gap between the sum symbol and the metric, i think that way it becomes clearer to what the sum symbol is refering. What do you think?

NearW:
@alschmut https://github.com/MaibornWolff/codecharta/issues/365
Relative metrics is another issue

Richargh:
I have a question: is the sum symbol displayed when I hover over a single building? In the screenshot it looks like it is and I feel like that's wrong. I want the sum symbol to only be displayed when I acutally sum values. Because at first it's not there and suddenly it pops up when I hover over a directory. That should get the user thinking.

Also we can then replace the sum/sigma symbol with the average or mean symbol (if there is one) for attributes that are relative according to the attributetypes.

alschmut:
@Richargh the current problem is, that we do not have the data from the importers to actually differnatiate between the absolute and the relative metrics. So we should (as @NearW mentioned) implement the relative-symbol together with #365

alschmut:
@Richargh Okey I think I got it: So that means we do something like
```
if (attribiteTypes.includes(metric) && attribiteTypes[metric].isRelative)
  // show relative symbol
else 
  // show sum/sigma symbol
```
but still we would have to calculate cumulative relative values for the folders. Because we always sum it up at the moment. In my opinion the whole topic about *relative metrics* is still a separate issue 😄

Richargh:
Relative metrics and calculating relative metrics is definately the separate issue #365. Your example-code is also part of that separate issue.

For the future: In another separate story we also need to describe all the attribute types in all importers as well or add a new `ccsh attributeFilter` that describes all attributes. 

But back to my original question: can we please only show the sum symbol, when we actually sum something? 

**When** I hover over a single building
**Then** the metric wihout a sum symbol is shown

**When** I hover over a package
**Then** the metric sum with a sum symbol is shown

Richargh:
What's the status here?

visualization/app/codeCharta/codeCharta.model.ts
```diff
@@ -113,17 +113,17 @@ export interface ColorRange {
 }
 
 export interface AttributeTypes {
-	nodes?: {
+	nodes?: Array<{
```
alschmut: Remove the question mark?

visualization/app/codeCharta/state/metric.service.ts
```diff
@@ -64,6 +74,20 @@ export class MetricService implements FileStateServiceSubscriber, SettingsServic
 		return metric ? metric.maxValue : undefined
 	}
 
+	public getAttributeTypeByMetric(metricName: string, settings: Settings): AttributeType {
+		let attributeType: AttributeType = null
```
alschmut: Can we refactor these loops to something like the following?
```
return settings.fileSettings.attributeTypes.nodes
    .find(x => Object.keys(x)[0] === metricName)
```

visualization/app/codeCharta/state/settings.service.ts
```diff
@@ -13,41 +13,39 @@ export interface SettingsServiceSubscriber {
...
+	private readonly debounceBroadcast: () => void
 
 	constructor(private $rootScope: IRootScopeService, private $timeout: ITimeoutService, private loadingGifService: LoadingGifService) {
 		this.settings = this.getDefaultSettings()
-		this.debounceBroadcast = _.debounce(
-			(update: RecursivePartial<Settings>) =>
-				this.$rootScope.$broadcast(SettingsService.SETTINGS_CHANGED_EVENT, { settings: this.settings, update: update }),
-			400
-		)
+		this.debounceBroadcast = _.debounce(() => {
```
alschmut: Do we wanna separate the two broadcast lines into an extra function?

visualization/app/codeCharta/state/settings.service.ts
```diff
@@ -13,41 +13,39 @@ export interface SettingsServiceSubscriber {
...
-		})
-		this.updateSettings({
-			fileSettings: this.getNewFileSettings(fileStates)
-		})
+		this.resetDynamicAndFileSettings(fileStates)
 	}
 
-	public onImportedFilesChanged(fileStates: FileState[], event: angular.IAngularEvent) {}
+	public onImportedFilesChanged(fileStates: FileState[], event: angular.IAngularEvent) {
+		this.resetDynamicAndFileSettings(fileStates)
```
alschmut: Is this necessary, as the `onFileSelectionStatesChanged`-event always gets called after new files were imported?

visualization/app/codeCharta/ui/metricType/metricType.component.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+			expect(CodeMapMouseEventService.subscribe).toHaveBeenCalledWith($rootScope, metricTypeController)
+		})
+	})
+
+	describe("onSettingsChanged", () => {
+		it("should set the areaMetricType to absolute", () => {
+			metricTypeController.onSettingsChanged(settings, { dynamicSettings: { areaMetric: "rloc" } }, undefined)
+
+			expect(metricTypeController["_viewModel"].areaMetricType).toBe("absolute")
```
alschmut: Use of interface `AttributeType.absolute`?

visualization/app/codeCharta/util/settingsMerger.spec.ts
```diff
@@ -126,106 +126,40 @@ describe("SettingsMerger", () => {
 		})
 	})
 
-	describe("MarkedPackages merge", () => {
```
alschmut: Why were all these MarkedPackages tests deleted?

visualization/app/codeCharta/state/metric.service.ts
NearW: refactored to
```
public getAttributeTypeByMetric(metricName: string, settings: Settings): AttributeType {
		const attributeType = settings.fileSettings.attributeTypes.nodes.find(x => _.findKey(x) === metricName)

		if (attributeType) {
			return attributeType[metricName]
		}
		return null
	}
```

visualization/app/codeCharta/state/settings.service.ts
NearW: If that is the case, no. removed.

visualization/app/codeCharta/ui/metricType/metricType.component.spec.ts
NearW: done

visualization/app/codeCharta/util/settingsMerger.spec.ts
NearW: Good catch. I can't explain why this happened. I re-added them

visualization/app/codeCharta/state/settings.service.ts
NearW: Moved the broadcast and reset of `this.update` into a function `notifySubscribers`. I couldn't find a better fitting name and I think `notifySubscribersAndResetUpdateState` is too much

Issue #547, opened 2019-05-28
Improve Reliability of merge filter


alexhunziker:
# Feature request

The MergeFilter relies on the hierarchies of the files merged. But these hierarchies are often not matching up (e.g. because one cc.json file has the root folder one hierarchy below or above the other one).
At the moment we just merge them anyways, the only check is that the project name adds up (which is never the case for parsed SVNLogs btw, since they don't obtain the project name). We should provide some better checks or merging mechanisms for the filter to:
1.  Inform the user about potential project structure mismatches
2.  If possible correct mismatches (?)

Richargh:
1. is a great idea. The merge should be aborted with a helpful message. But what can we suggest to the user instead?
2. I wonder if we should do this automatically at all. How about we create a new CC filter called `ccsh moveHierarchy` that allows us to move complete hierarchies inside one cc.json.

alexhunziker:
I like the idea of a separate filter, it is probably more transparent than if we try to fix something automatically. What that filter would actually do, is extract a subproject from a cc.json, right? Do you think  it should also able to insert additional levels of hierarchy into a project?

Richargh:
I think both ways are necessary. If you have 10 files and 9 of them have the same hierarchy, you want some filter that can fix the hierarchy of that 10th file. 

After what tool should we model this? Is it a `grep`, a `sed`, a `mv`, a ... Or perhaps it's something more visual, that open some command-line menu where you can move stuff up and down? It seems this tool is something you use to fix something in your cc.json, not something you would use in your build pipeline all the time. What do you guys think?

alexhunziker:
For the merge filter i have added some informative warnings about merges that did not go how the user might have intended them to go.

Regarding the modification of project structures, I have implemented a new filter that can 
- extract a subproject from the json structure (e.g. making /root/foo/bar -> /root and discarding the rest of the project) -OR- 
- move nodes, that are not leaves, aroung (e.g. /root/foo/bar -> /root/bar)
- additionally it can print the first x levels of the hierarchy of a cc.json file to the console, so that users can inspect the structure

Interface wise I kept it in line with our other ccsh tools, so everything is done by specifying command line paramters (e.g. --extract=/foo/bar). What do you think of that approach, and how do you think this filter should be called? (Personally i think we should model the tool in such a way that it can be part of a pipeline, since you may have two different data sources which will always produce missmatching project structures that one might want to correct automatically)

Richargh:
[...]

Issue #276, opened 2018-11-12
MergeFilter is too slow with big files


alschmut:
# Feature request

## Description

As I have talked to @ukinimod, we should really add the option to delete nodes from a cc.json file, because merging huge files takes much too long.

## Question
- Analysis: How could an analysis filter look like to input the nodepaths?
- Visualization: 
    - Do we gonna introduce another *hard deletion* `ExcludeType`, which is also stored in the blacklist to make the deletion repeatable on newer versions of the cc.json map (with another new analysis filter) or 
    - should we give the user the option to delete excluded nodes when downloading the code map file?

ukinimod:
@alschmut I'm a bit at a loss what the FR is about. 
For the mergefilter there is also the solution to use not the recursive node merging but the leaf node merging. Probably the former should be converted entirely in the later. The performance for the non-traversing strategy is much faster and was enough for my examples.

I still would propose to be able to do white and blacklisting of nodes using the backend as it fits better into my personal workflow. The use case I have in mind is, getting a c# codebase, there are certain generated files I always want to exclude (Reference.cs,...). Even when I get two snapshots of a codebase I want to include / exclude the same things in both cc.json. When I do this on the backend I am able to do this repeatably without additional manual effort. When I am able to do hard delete it speeds up every process including merging.

It works now, with a bit of additional manual effort, so for me the priority of the issue is rather low.

Richargh:
This relates to #181 which I closed too early.

Richargh:
@alexhunziker did #547 also improve the speed of the merge filter or why did you reference this issue?

alexhunziker:
I think @alschmut added this issue to the pull request of #547 since in the description of this issue it is suggested to address this by providing a hard-delete option, which the new filter does.

Regarding the actual speed of the merge-filter no measures were taken. So it depends wha the issue is actually about; if it is about the merge-filter being slow due to unnecessarily large .cc.json files then it is addressed, if it is about the actual speed of the merging it is not.

Richargh:
Works for me. Should we close this feature for now and see if we get different feedback?

PR #567, opened 2019-06-13
Feature/547/reliability merge filter, StrucureModifier

### LINKED ISSUE #181: A filter to remove unneeded parts from the cc.json

alexhunziker:
# StructureModifier added to modify structure of cc.json and improved logging regarding the success of merges

closes #181 #547 #276

## Description
- Added logging, that provides informative statements about the merge process
- Refactored some methods of MergeFilter
- Added StructureModifier that can remove/rename nodes, set a node as new root of a project and print the hierarchy of the project file to the console.
- Added possibility to pipe input project into StructureModifier instead of using intermediate files

## Definition of Done

- [x] **All tests pass**


analysis/filter/MergeFilter/README.md
```diff
@@ -13,9 +15,16 @@ specific strategy. Currently there are two main strategies:
 - recursive (`--recursive`) (default): leave structure of additional files. This will also merge optional edges.
 - leaf (`--leaf`) (beta):  fit leaf nodes into reference structure according to their name (and tail of their path), 
 either adding missing leaves (`--add-missing`) or ignoring them (default)
+- leaf (`--leaf`) (beta):  fit leaf nodes into reference structure according to their name (and tail of their path)
```
alschmut: This line is duplicated isn't it?

analysis/filter/MergeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/mergefilter/ProjectMerger.kt
```diff
@@ -63,19 +66,28 @@ class ProjectMerger(private val projects: List<Project>, private val nodeMerger:
...
+            getMergedEdges()
         } else {
-            return mutableListOf()
+            getEdgesOfMasterAndWarnIfDiscards()
+        }
+    }
+
+    private fun getEdgesOfMasterAndWarnIfDiscards(): MutableList<Edge> {
+        projects.forEachIndexed { i, project ->
+            if (project.edges.isNotEmpty() && i > 0) logger.warn("Edges of ${project.projectName} were ignored. Use recursive strtegy to merge edges.")
```
alschmut: typo: strtegy -> str**A**tegy

analysis/filter/StructureModifier/README.md
```diff
@@ -0,0 +1,52 @@
...
+                             show first x layers of project hierarchy
+  -r, --remove=<remove>...   node(s) to be removed
+  -s, --setRoot=<setRoot>    path within project to be extracted
+  -t, --moveTo=<moveTo>      ... move nodes to destination folder
+```
+
+
+## Examples
+
+> sh ccsh modify foo.cc.json -p=2
```
alschmut: Do we wanna remove the `sh` to be consistent with all other ccsh readme's, where it currently only refers to `ccsh` or `./ccsh`

analysis/filter/StructureModifier/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+> sh ccsh modify foo.cc.json --moveFrom=/root/foo --moveTo=/root/bar -outputFile=project.cc.json
+
+> sh ccsh modify foo.cc.json --setRoot=/root/foo/ --projectName=NewName
+
+ ## Piped input
+
+ Instead of providing a cc.json file as input, a project can also be piped to the filter:
+ ```
+ cat demo.cc.json | sh ccsh modify -p=2
```
alschmut: What stand this `sh`?

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/StructureModifier.kt
```diff
@@ -0,0 +1,105 @@
...
+        return null
+    }
+
+    private fun readProject(): Project? {
+        val bufferedReader = source?.bufferedReader() ?: return readPipedProject()
+        return try {
+            ProjectDeserializer.deserializeProject(bufferedReader)
+        } catch (e: Exception) {
+            val input = source!!.name
+            logger.error("$input is not a valid project file and is therefore skipped.")
```
alschmut: Where do you validate the project? Doesn't the `deserializeProject(...)` just check if its a valid json content? Also see line 87 in `readPipedProject()`

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
```diff
@@ -0,0 +1,52 @@
...
+        val filteredChildren = node.children.filter { it.name != path.firstOrNull() || path.size > 1 }
+        node.children = filteredChildren.map {
+            filterNodes(path.drop(1), it)
+        }.toMutableList()
+        return node
+    }
+
+    private fun removeNodes(paths: List<List<String>>): MutableList<MutableNode> {
+        var root = project.rootNode.toMutableNode()
+        paths.forEach { root = filterNodes(it.drop(1), root) }
```
alschmut: What do you think about a method called similar to this? Feel free to adjust the method name, but I think it's easier to read.
```
pathWithoutRootName(path) {
    return path.drop(1)
}
```

alschmut: Okey I just realised, the method is used more often... what about `pathWithoutFirstFolder()`?

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
```diff
@@ -0,0 +1,52 @@
...
+        return mutableListOf(root)
+    }
+
+    private fun removeEdges(removePatterns: Array<String>): MutableList<Edge> {
+        var edges = project.edges
+        removePatterns.forEach { path -> edges = edges.filter { !it.fromNodeName.contains(path) && !it.toNodeName.contains(path) } }
```
alschmut: Actually the `edge.fromNodeName` and `edge.toNodeName` is possibly a gitignore pattern, so matching the path-string doesn't cover all options. Also see line 48 in `removeBlacklistItems`

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/FolderMover.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private fun moveNodes(moveFrom: String, moveTo: String): List<MutableNode> {
+        val originPath = getPathSegments(moveFrom)
+        val destinationPath = getPathSegments(moveTo)
+        val rootNode = project.rootNode.toMutableNode()
+
+        val newStructure = listOf(removeMovedNodeFromStructure(originPath, rootNode)!!)
+
+        return if (toMove == null) {
```
alschmut: How about:

```
if (toMove == null) {
    logger.warn("Path to move was not found in project. No nodes are therefore moved")
} else {
    insertInNewStructure(destinationPath.drop(1), rootNode)
}
return newStructure
```

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/FolderMover.kt
```diff
@@ -0,0 +1,97 @@
...
+            null
+        } else {
+            node.children = node.children.mapNotNull { child -> removeMovedNodeFromStructure(originPath.drop(1), child) }.toMutableList()
+            node
+        }
+    }
+
+    private fun insertInNewStructure(destinationPath: List<String>, node: MutableNode) {
+        if (destinationPath.isEmpty()) {
+            node.children.addAll(toMove!!)
```
alschmut: What happens with children that that have the same name and the same type? Because we can not import cc.json files with invalid children into the visualisation.

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractor.kt
```diff
@@ -0,0 +1,69 @@
...
+                copyBlacklist()
+        ).build()
+    }
+
+    private fun extractNodes(extractionPattern: List<String>, node: MutableNode): MutableList<MutableNode> {
+        val children: List<MutableNode> = node.children
+        val extractedNodes: MutableList<MutableNode> = mutableListOf()
+
+        val currentSearchPattern = extractionPattern.firstOrNull()
+        if (currentSearchPattern == node.name && extractionPattern.size == 1) return node.children.toMutableList()
```
alschmut: How about:
```
if (currentSearchPattern == node.name) {
    return if (extractionPattern.size == 1) node.children.toMutableList()
    else children.map { child -> extractNodes(extractionPattern.drop(1), child)}.flatten().toMutableList()
}
```

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
alschmut: The more appropriate approach for the `NodeRemover` were the following simplified code. `newGitignoreList` in this case can be either edge-paths or the blacklist-paths.
```
newGitignoreList = gitignorePathList.filter { gitignorePathMatchesAnyNode(it, newRootNode) }
```

I admit its still very tricky for the `SubProjectExtractor` and `FolderMover`...

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemoverTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    fun `Affected edges are removed`() {
+        val subProjectExtractor = NodeRemover(sampleProject, null)
+
+        val result = subProjectExtractor.remove(arrayOf("/root/foo"))
+
+        Assertions.assertThat(result.edges.size).isEqualTo(1)
+        Assertions.assertThat(result.edges.first()).isEqualToComparingFieldByFieldRecursively(sampleProject.edges.last())
+        Assertions.assertThat(result.edges.map { it.fromNodeName }).doesNotContain("/roo/foo")
```
alschmut: Typo: "/roo/foo" -> "/root/foo". See next line as well

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemoverTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    fun `Affected edges are removed for multiple removals`() {
+        val subProjectExtractor = NodeRemover(sampleProject, null)
+        val toExclude = arrayOf("/root/foo/file1", "root/else/")
+
+        val result = subProjectExtractor.remove(toExclude)
+
+        val shouldBeExcluded = listOf("/root/foo/file1", "root/else/file1")
```
alschmut: I think testing against `listOf("/root/foo/file1", "root/else")`makes more sense, as the test removes exactly those two folders

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/ProjectStructurePrinterTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun `Nodes on desired max hierarchy level are printed`() {
+        val projectStructurePrinter = ProjectStructurePrinter(sampleProject)
+
+        projectStructurePrinter.printProjectStructure(2)
+
+        Assertions.assertThat(outContent.toString()).contains(listOf("root", "src", "main", "test"))
+    }
+
+    @Test
+    fun `Nodes deeper that max hierarchy level are not printed`() {
```
alschmut: Typo: that -> than

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractor.kt
```diff
@@ -0,0 +1,69 @@
...
+    private fun extractRenamedEdgesForPattern(pattern: String): List<Edge> {
+        return project.edges.filter { it.fromNodeName.startsWith(pattern) && it.toNodeName.startsWith(pattern) }
+                .map { edge ->
+                    edge.fromNodeName = "/root" + edge.fromNodeName.removePrefix(pattern)
+                    edge.toNodeName = "/root" + edge.toNodeName.removePrefix(pattern)
+                    edge
+                }
+    }
+
+    private fun copyAttributeTypes(): MutableMap<String, MutableList<Map<String, AttributeType>>> {
```
alschmut: I think this were also the correct way to collect attributeTypes, but I don't know, if that might have a high impact to the performance. `attributeList` can be `attribute types.edges` or `attributeTypes.nodes`
```
newAttributeList = attributeList.filter { attributeExistsOnAnyNode(it, newRootNode) }
```

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractorTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        val extractedNodeChild = result.rootNode.children.first()
+        Assertions.assertThat(extractedNodeChild.attributes).containsKey("nloc")
+    }
+
+    @Test
+    fun `Project name is changed if provided`() {
+        val subProjectExtractor = SubProjectExtractor(sampleProject, "foo")
+
+        val result = subProjectExtractor.extract("/root/somethig")
```
alschmut: Typo: "/root/somethig" -> "/root/something"
I know its not important here ;)
again in line 65

alschmut:
@alexhunziker All in all I really like the code for this filter!! 👍 its a the simple structure with one public function in each file, which is only called from within the main class `StructureModifier`. Some of the requested changes, are super trivial, some of them require a quite complex implementation and should more be like questions and reminders :)

btw: I didn't manually test it yet

analysis/filter/MergeFilter/README.md
alexhunziker: it most certainly is, i don't know what my intention was here. removed it.

analysis/filter/StructureModifier/README.md
alexhunziker: i put it here, since when you do the thing with the piped input it only works with the sh in front of it (otherwise it will overwrite the contents of the bash file), so in order to keep it consistent within the file, i have added the sh everywhere.

analysis/filter/StructureModifier/README.md
alexhunziker: without the sh, the piped input will be written to the ccsh file, but what we want is the ccsh file to be executed, therefore we tell the command line to execute it with the `sh`

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/StructureModifier.kt
alexhunziker: good point, we have a Validation tool in ccsh (no idea how powerful it is), maybe it would be a good idea to test the input file with it. I will check that

alschmut: it validates against a json schema file `cc.json`

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/FolderMover.kt
alexhunziker: that sounds like a case that could occur; i added a filter to sort the duplicate nodes out and give a warning

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/FolderMover.kt
```diff
@@ -58,7 +57,17 @@ class FolderMover(private val project: Project, private val projectName: String?
...
+            val destinationNodeNamesAndType: HashMap<String, NodeType?> = hashMapOf()
+            node.children.forEach { destinationNodeNamesAndType[it.name] = it.type }
+
+            val filteredNodesToMove = toMove!!.filter {
+                !(destinationNodeNamesAndType.containsKey(it.name) && destinationNodeNamesAndType[it.name] == it.type)
+            }
+            if (filteredNodesToMove.size < toMove!!.size) {
+                logger.warn("Some nodes are already available in the target location, they were not moved and discarded instead.")
+            }
+            node.children.addAll(filteredNodesToMove)
```
alschmut: I think this is a better solution 👍 but, what if the children with the same name+type have a different content? Should there be an option to rename the new child, instead of deleting it?

alexhunziker: Well here is what I thought when implementing this:

This filter aims mostly at correcting non-matching hierarchies, so that the .cc.json files can be merged, therefore the major use case will be something like moving nodes from /root/ to /root/foobar/ or from /root/foo/ to /root/bar (in case the folder has two different names) or vice versa. In most cases I suppose the target folder will not exist in the original structure.

Renaming the new child (foo -> foo_copy or similar) does, in my opinion not make sense in most use cases, since then we do not get matching hierarchies either, because foo_copy cannot be merged with foo.

If there is really the case where a user wants to move /something to /else and else and have matching children, merging them would probably be the most useful way. However I am reluctant to implement something like this, because I don't want to duplicate the merge-filter here, also i think the use case is very rare. If the use case occurs however, the two filters combined allow to solve the issue.

Therefore I think that discarding & warning is probably the best way to implement this here

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractor.kt
alexhunziker: Theoretically you are right, we could filter the new Attribute list. However, in the visualization non-available attributes are not shown and the list of attributes will never be that long, that it significantly bloats the file size. Therefore I think filtering the attribut types (or not) has little practical impact. Or am I missing something here?

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemoverTest.kt
alexhunziker: good catch, test does not make a lot of sense otherwise :)

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
alexhunziker: do you know any java/kotlin library that supports gitignore pathmatching? I was searching for this in another context and did not find anything promising then

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractor.kt
alschmut: Yea you're right ;) and the attributes are almost always the same for every node.

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/StructureModifier.kt
alexhunziker: i noticed, that the projectDeserializer actually checks for the structure of the input file, a further check is therefore not necessary. I added a testcase to verify that we are discarding jsons that are not valid project files.

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
alschmut: Unfortunately not.

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/FolderMover.kt
alschmut: Alright, yea that makes sense 👍

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/StructureModifier.kt
alschmut: Ah nice, so this `GSON.fromJson(reader, Project::class.java)` actually compares the json input to the `Project`'s class structure 👍

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
alexhunziker: In which case is the edge a gitignore pattern?

alschmut: True, the edge is currently never created as a gitignore pattern, but the blacklisted nodes can be

alexhunziker:
@aleschmut I noticed that when I put in a parameter as --moveFrom=/root/foo/ GitBash automatically converts that to C:/Users/AlexH/AppData/Local/Programs/Git/root/foo ... however, without the first slash it works correctly. Could anybody check how the behaviour is on a mac?

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemover.kt
alexhunziker: In this case I would suggest we leave it for now, since potentially irrelevant gitignore patterns can be removed in the visualization afterwards. I think implementing a Gitignore matcher for that may be a bit of an overkill, since as a user I would not necessarily expect the blacklist entries to be checked for relevance if i remove some nodes. If we find a library for Gitignore matching we can add it later. What do you think?

alexhunziker:
There seems to be an issue with identifying paths & setSource, i will fix that

alschmut:
Generally (wit one exception) it works 👍 Just a few things to mention, after I run some commands:

- Executing `ccsh modify --moveFrom=root/foo --moveTo=root/bar` moves the nodes inside the folder, not the folder itself. Probably thats the use case, I was just confused, because I expected the folder will be like `root/bar/foo` afterwards.
- Executing `ccsh modify --remove=/root/ codecharta.cc.json -o removeFolder.cc.json` throws the following error. I tried with different folders and checked the hierarchy with `--printLevels`. The file is fine.
    ```
    [main] ERROR de.maibornwolff.codecharta.filter.structuremodifier.StructureModifier - The piped input is not a valid project and no input file was specified
    ```

alexhunziker:
> Executing ccsh modify --moveFrom=root/foo --moveTo=root/bar moves the nodes inside the folder, not the folder itself. Probably thats the use case, I was just confused, because I expected the folder will be like root/bar/foo afterwards.

Yes that is the expected behaviour. Do you think we could/should change the naming of the parametersto make it more clear what it does? like `moveInto` or something?

> Executing ccsh modify --remove=/root/ codecharta.cc.json -o removeFolder.cc.json throws the following error. I tried with different folders and checked the hierarchy with --printLevels. The file is fine. 

Ok, so what happens here is the following: remove acccepts a number of arguments, so `--remove=root/foo root/bar` removes both foo and bar, therefore picocli thought the file was another path to remove and did not recognize it as file. If you flip the arguments it should work. If we want to address this, we have to disable the option to remove multiple paths at a time.

alschmut:
> Do you think we could/should change the naming of the parametersto make it more clear what it does? like `moveInto` or something?

I think we can keep it as there is probably no short name for renaming the folder and moving and its children

>  If you flip the arguments it should work

yes it did 👍

Issue #627, opened 2019-07-05
Move File extension bar between file bar and metric bar


Richargh:
# Feature request

## Description

The file extension bar is currently below the metric bar. Please move it above the metric bar to clear some space for our future search panel and avoid overlay between the two.

PR #632, opened 2019-07-11
Feature/526/new search panel

### LINKED ISSUE #526: New Search Bar
# Feature Request

Let's incrementally change our UI. Next stop, move the search bar next of the metrics, so it is immediately obvious to our users where to search for stuff. Next to the search field are three buttons. They behave like tabs and allow us to switch between tree view, hidden items and exclude items. The search bar can also be opened independantly from the metrics.

grafik

The search + scenario + metrics bar might become a bit wide. That's ok. But please also add a helpful little `...` or something to the right side of the bar if some elements are hidden.

Please change the symbol in the hide and exclude list that you use to remove something from that list. The users think that the trash can symbol completely removes the excluded or hidden buildings from the cc.json. What if we instead present the user with a `regular eye` icon on the **left** side of the hidden/excluded entry? On the left is normally the hide or exclude symbol but that is redundant in the hide/exclude list anyway.

alexhunziker:
# New search panel

closes #526 #627 

## Description

- New Search Panel (As part of RibbonBar)
- Splited Search Panel into components
- Separate Views for Hides and Excludes
- Moved FileExtensionBar above RibbonBar
- Added some tests
- Removed Burger Menu / SideNav

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/codeCharta.model.ts
```diff
@@ -80,6 +87,7 @@ export interface DynamicSettings {
 	searchPattern: string
 	margin: number
 	colorRange: ColorRange
+	searchPanelMode: SearchPanelMode
```
alschmut: This should actually be inside appSettings, as it is not related to the map. But as we wanna split the settings object step by step, should we add a separate service called `searchPanelModeService`, which would have a similar structure to the `loadingGifService`?

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -33,6 +34,8 @@ export class CodeChartaController implements SettingsServiceSubscriber {
 		private settingsService: SettingsService,
 		private codeChartaService: CodeChartaService,
 		private fileStateService: FileStateService,
+		// tslint:disable-next-line
+		private nodeSearchService: NodeSearchService, // We have to inject it somewhere
```
alschmut: I don't get yet, why this has to be injected here, as it's not even used. Isn't there a nicer solution?

visualization/app/codeCharta/state/settings.service.ts
```diff
@@ -13,7 +22,7 @@ export interface SettingsServiceSubscriber {
 
 export class SettingsService implements FileStateServiceSubscriber {
 	private static SETTINGS_CHANGED_EVENT = "settings-changed"
-	private static DEBOUNCE_TIME = 400
+	private static DEBOUNCE_TIME = 100
```
alschmut: We should actually stick to a higher debounce time, as the sliders start lagging with larger maps. I think you reduced the debounce time, because otherwise it takes some time to open the searchPanel, but also here a separate service for storing the mode information helps (see other comment)

visualization/app/codeCharta/ui/fileExtensionBar/fileExtensionBar.component.scss
```diff
@@ -37,7 +34,15 @@ file-extension-bar-component {
...
 		text-align: center;
 		background-color: white;
-		overflow: auto;
+		height: 0px;
+		overflow: hidden;
+		transition: all linear 0.2s;
+
+		&.expanded {
+			border-top: 1px solid #cdcdcd;
+			overflow: auto !important;
```
alschmut: Can we add `overflow` and `border-top` to the `.bar-details` instead of having it inside the `&.expanded` CSS class?

visualization/app/codeCharta/ui/matchingFilesCounter/matchingFilesCounter.component.ts
```diff
@@ -0,0 +1,61 @@
...
+			SettingsService.subscribe($rootScope, this)
+		}
+
+	public onNodeSearchComplete(searchedNodes: CodeMapNode[], event: angular.IAngularEvent){
+		this.searchedNodeLeaves = this.getSearchedNodeLeaves(searchedNodes)
+		this.updateViewModel(this.searchedNodeLeaves, this._viewModel.blacklist)
+	}
+
+	public onSettingsChanged(settings: Settings, update: RecursivePartial<Settings>, event: IAngularEvent){
+		this._viewModel.blacklist = settings.fileSettings.blacklist
```
alschmut: I think we should check if the blacklist is in the update object, as we might run this code too often, when its not necessary

visualization/app/codeCharta/ui/matchingFilesCounter/matchingFilesCounter.component.ts
```diff
@@ -0,0 +1,61 @@
...
+		blacklist: BlacklistItem[]
+	} = {
+		fileCount: 0,
+		hideCount: 0,
+		excludeCount: 0,
+		searchPattern: "",
+		blacklist: []
+	}
+
+	private searchedNodeLeaves: CodeMapNode[] = [] // TODO: geht das in die DynamicSettings oder service
```
alschmut: TODO:
This `private searchedNodeLeaves` stays there and as it is not updated here, we don't need to notify/call anybody else. or did I miss something?

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.html
```diff
@@ -1,9 +1,11 @@
-<md-card id="header">
+<search-panel-component></search-panel-component>
+
+<md-card class="header" id="header">
```
alschmut: Think about class and id name *header*. @alexhunziker and me would rather rename it to metrics-panel. Can we possibly remove the id or is it used somewhere necessarily?

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.scss
```diff
@@ -124,8 +147,19 @@ ribbon-bar-component {
 				vertical-align: top;
 
 				reset-settings-button-component .md-button {
-					top: ($heightExpanded - 45) + px;
+					top: 0 + px;
```
alschmut: only write `0`

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.scss
```diff
@@ -124,8 +147,19 @@ ribbon-bar-component {
...
 					right: -5px;
+					opacity: 0;
+					height: 0px;
```
alschmut: Only write `0`

visualization/app/codeCharta/ui/matchingFilesCounter/matchingFilesCounter.component.ts
alexhunziker: the todo is a leftover and outdated, i will remove it.

visualization/app/codeCharta/ui/fileExtensionBar/fileExtensionBar.component.scss
alexhunziker: the border-top is in expanded, since otherwise we have a too strong border to the ribbon bar when not extended; the overflow is here i think because otherwise it would not hide the labels but show them over the ribbon bar

visualization/app/codeCharta/state/settings.service.ts
alexhunziker: I have reset that to 400ms, since we now have the SearchPanelService

visualization/app/codeCharta/ui/matchingFilesCounter/matchingFilesCounter.component.ts
alexhunziker: I don't think that this has an impact on performance, but i changed it anyways in order to be consistent with other components

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.html
alexhunziker: it is used for selecting which part of the ribbon bar is expanded when clicking on the section titles/the dedicated expansion button. Furthermore it is used for the e2e tests. Therefore i renamed the id instead

PR #638, opened 2019-07-16
Feature/576/spotlight hovering

### LINKED ISSUE #576: Change hover effect to a spotlight
# Feature request

Currently CodeCharta slightly lightens the color of buildings when the mouse hovers over them. It works but when doing a presentation the people further in the back can't see the hovered building so clearly. 

## Description

How about we change the effect to something like this (this picture is just an example and no longer covers the intent completely, see below for details):

The idea is to darken all buildings that are not hovered, instead of highlighting the one building that is hovered. It works like this:

* buildings have regular brightness but then
* mouse enters some perimeter and
* buildings far away from the mouse are now very dark
* They get progressively lighter the closer they are to the mouse pointer
* the building right under the mouse pointer is significantly brighter than the buildings sorrounding it but still only regular brightness

The result is somewhat like a **spotlight**. We don't need that light circle or different levels of brightness for a single building though. It's probably not necessary to darken the whole background because it might be visually tiring.

## In combination with the new select

The nice thing is we can also combine this rather well with a different way of *selecting* buildings (not part of this issue):

NearW:
# Hovering of Buildings without shader

closes #576

## Description
- Removed Selection and Highlighting from Fragmentshader
- Added a Switch to enable Presentation Mode for Spotlight Hovering
- Implemented Spotlight-Hovering based on distance
- Refactored codeMapMesh and CodeMapBuilding to work as a view/model
- Inverting Delta Colors will now actually swap the two entries in the settings
- Split up CodeMapMouseEvent-Events
- Update-Loop will check for mouse-movement before executing raycast

Before we used uniforms in the fragment shader to set the delta color. Uniforms bundle fixed values. Since we apply highlighting without the shader by setting a color, we need to set a deltaColor as well. This is done by removing the delta color from the uniform and adding it as an attribute to the geometry. Attributes can be altered and injected into the shader. That's why the codeMapBuilding has an deltaColor field now.

## Definition of Done

- [x] **All tests pass**


alschmut:
While having a pair review with @NearW we realised that hovering and selecting buildings currently doesn't work in delta mode

PR #656, opened 2019-07-29
Spike: CCSH script generator

mrpatpat:
This tool uses inquirer.js to provide a CLI for less technical users. It focusses on guiding the user through the process of correctly using ccsh and using more descriptive and less technical terms.

The idea is to add a layer on top of ccsh. The user gets  guided through the process of adding projects to the analysis. 
They come from different sources like the file system, SVN or Git. By adding them the script knows where they are located and what there names are. In case of git projects the script knows where they will be located at runtime after cloning and will use this path instead. 
After adding one or more named projectes we can start adding analysis specific to the given projects. The script knows where to run the analysis and how the name the file because the project was already added as a defined entity.
Adding a filter consumes an analysis and results in an analysis. Since all analysis and their source project are defined in the tool, we can easily select them before filtering.

Currently the resulting script can be generated or run. In the future it would be nice if it also could visualize it directly.

render1564386563457

This is only a rough implementation to show what is possible with inquirer.js and shell.js.

The current features are: 
* adding projects from git, the tool manages cloning and remembering the path by itself
* adding projects from file system (untested)
* analyzing added projects with sourcecodeparser
* analyzing added projects with scmlogparser, the script handles generating git logs
* analyzing added projects with sonarqube
* merging added analysis
* generating a commented shell script to produce the defined maps
* running the script from within the tool and reporting errors for each line
* the generated script cleans up after itself (currently only when no error happens)
* context sensitivity e.g. does a project support git ? If yes, then enable git log parsing

Known bugs / problems that need to get solved:
* adding a fs project with the path . (current directory) makes the scmlogparser add wrong commands
* Handling of paths in general should be addressed more
* most commands do not ask for their parameters. They are hardcoded. Only the sonarimporter asks for every parameter and provides default values
* Input validation is missing completely
* Deleting/Listing projects/analysis is not possible 
* file naming is very basic. The script ensures uniqueness by combining project names and indices. It can propably be done in a more readable way
* state object is global and does more than state handling
* propably many other things. Typescript would be nice, too


Richargh:
What's happening here?

> This is only a rough implementation to show what is possible with inquirer.js and shell.js.

Are we investing time into this? Do we clean it up / is it a spike that we make available for everyone? How do we ensure the wizard works? Is it a new project sepearate from analysis and visualization? 

Please don't just move stuff into Review like this.

alschmut:
@Richargh sorry, I was moving the PR from zenhubs column Backlog to Review, because I thought PR's should be placed there, since zenhub doesn't do it automatically (other than waffle.io) and some of us might forget about it, if it's hidden somewhere in the Backlog. Btw: GitHub says 
> Pull requests let you tell others about changes you've pushed to a branch in a repository on GitHub

So PR's are not only for reviews. It's mainly about collaboration.

Regarding what to do:
- either @fatihi / @Richargh prioritise it and the team will work on it,
- somebody convicts you to prioritise it or
- if you don't, @mrpatpat respectively whoever who is interested can work on it in their spare time (potentially until its ready for production and @fatihi / @Richargh are convinced to integrate it)

In my opinion this is a pretty cool start to integrate users, who do not want to use the 
`ccsh [subcommand] --help`
every time in order to find out which parameters to use. A simple potentially filtered selection of option makes it a lot easier to go through the process. Analysing multiple sources and merging them sounds easier as well.

Richargh:
> * somebody convicts you to prioritise it or

not a good solution. I don't like being convicted. :)

I am fine with releasing a prototype to users who are interested. Better than let this useful code start carrying dust. We should just mark it explicitely als alpha or something.

Before that though I would like to know what is the code quality? Do we need to rewrite it, can we give it to our users, does it have bugs, does it delete code by mistake etc.

NearW:
In my opinion, this feature needs some more time to fix bugs and improve code quality.
- Use Typescript
- Work on a better project-structure in ccshg
- Should we move this to it's own project on root level?
- Fix all bugs

We can focus on the functionality that the ccshg provides right now and ship it as soon as it's production ready and wait for user feedback.

PR #669, opened 2019-08-07
added missing delays, to make e2e tests more reliable

alexhunziker:
# Added Delay to e2e tests

## Description

Added Delays to RibbonBar and SearchPanel Page objects, since the tests using therir methods were unreliable and delivered unpredictable results.

## Definition of Done

- [x] **All tests pass**


NearW:
If we do it, do it consistently. Add a delay to every page-object function.
Also, let's change the signature of the delay method to this:
```
function delay(ms: number = 1000) {}
```

If we need higher delay at some point, we can pass a value, otherwise a fixed delay seems good.

alexhunziker:
So I have added two delays to functions with trigger annimations; they were called by the two tests i noticed failing. The delays represent the annimation duration. I think adding a delay to everything would be a bit of an overkill.

PR #671, opened 2019-08-09
Tech/split up settings events

NearW:
# Split up some settings-service-events

## Description
Split up the subscribers. We now have:
- BlacklistSubscriber (did blacklist change?)
- MetricSubscriber (did the currently selected metrics change?)
- SearchPatternSubscriber (did the current search-pattern change?)

As you can see, we can pretty much do this for all entries in the settings-object. For future refactoring, we should try to achieve:
- Move fields of appSettings / dynamicSettings to it's fitting location. AppSettings can be exported, dynamicSettings are just temporarily. (Currently selected metrics should be in appSettings?)
- Split up the sub-settings into it's own services, creating `DynamicSettingsService`, `AppSettingsService` etc.
- Achieve a linear dataflow when loading a new map.

## Definition of Done

- [x] **All tests pass**


alschmut:
I think the general goal is quite good, as split events make it more readable and easier to subscribe. I just had a quick look at the changed files and would say, that everything is good, except for the two settingsService files, which increased by more than 230 lines. Why do we need the settingsService as it was, if we split all events? My suggestions were to create individual services like `blacklist.service.ts` or `areaMetric.service.ts`, which only stores this one object. I guess this is close to the principle of [blacklist-as-a-class](https://github.com/MaibornWolff/codecharta/blob/master/TECHNICAL_DEBT.md#blacklist-as-a-class) mentioned by @NearW. 

If we scale that for all settings, we will have around 20-25 services, each storing one property. This also allows us to move all functionality from [codeMapHelper.ts](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/util/codeMapHelper.ts) and [codeMap.actions.service.ts](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/ui/codeMap/codeMap.actions.service.ts) into its corresponding service (blacklist, edge etc.)

We could possibly also combine some trivial appSettings into one service, but I think splitting the events also means splitting the service.

What do you guys think?

alschmut:
I just saw following bug on this branch (not on master):
When importing two files and selecting [Multiple](url) or `Delta` Mode
then the map is not being updated and the settings-loading-gif is spinning forever

NearW:
This is a great idea. I thought about splitting the settingsService into `AppSettingsService`, `DynamicSettingsService` and `FileSettingsService` for now, but I really like your approach. The question is: Can you as a reviewer handle that refactoring in this PR? Otherwise we'd merge this and do a new one.

> I guess this is close to the principle of blacklist-as-a-class mentioned by @NearW.
Not exactly, since we really want to have the functionality of the blacklist (adding or removing entries) in a class `Blacklist`. You can instantiate that class multiple times. You can't do that with the service. But think about storing multiple blacklists now, if we actually build a `BlacklistService`. This would be so easy to do! I'm hyped!

> When importing two files and selecting Multiple or Delta Mode then the map is not being updated and the settings-loading-gif is spinning forever
I'll look into this on friday. Thank you!

Issue #476, opened 2019-04-12
markedPackages blacklist behaviour onFileSelectionChanged


alschmut:
# Feature request

## Description

Currently the behaviour when changing the file selection including changing the mode (single, delta, multiple), doesn't adapt the backlist or markedPackages to the new map nodes, as there is just one settings-object, so its mostly useless as different nodes exist. In #178 the behaviour is not yet implemented, which means the backlist and markedPackages get deleted, when selecting another map.

# Question
- How about writing each blacklist or markedPackage entry to the imported `CCFile.settings.fileSettings`, when the path of the entry can be resolved?
- That means for every fileSelectionChange it merges the defined fileSettings of all visible files and sets those merged fileSettings into the settingsService, which blacklistPanel accesses.
- Don't forget about deleting entries in all visible files, when the user wants to delete those

alschmut:
I was just shortly thinking about what the expected behaviour for the fileSettings (inkl. blacklist, markedPackages) is and came up with some ideas. It looks like a lot, but the options are actually quite similar. Just make it a bit easier to understand I separated everything.
- `fileSettings` include for example the blacklist and markedPackages
- `file<X>.fileSettings` is the fileSettings object from the imported file, without any changes
- `fileSettingsList.<NAME>` represents the idea of a list of objects (okay it's actually a map) `Map<listName, fileSettingsObject>` which stores lots of separated fileSettings for each different single, delta, multiple visible file-combinations

### Single
- Single fileA
  - Copy fileA.fileSettings to fileSettingsList.SingleA
  - Show fileSettingsList.SingleA
  - Adding an item adds it to fileSettingsList.SingleA
- Single fileB
  - Copy fileB.fileSettings to fileSettingsList.SingleB
  - Show fileSettingsList.SingleB
  -  Adding an item adds it to fileSettingsList.SingleB
- Single fileC 
  - Copy fileC.fileSettings to fileSettingsList.SingleC
  - Show fileSettingsList.SingleC
  - Adding an item adds it to fileSettingsList.SingleC

### Delta fileA and fileB
- Option 1
  - Show `mergeOf(fileA.fileSettings, fileB.fileSettings, fileSettingsList.SingleA, fileSettingsList.SingleB)`
  - Adding an item adds it to fileSettingsList.SingleA and fileSettingsList.SingleB
  - Question: But then those added items will also exist in Single mode showing fileA e.x. afterwards
- Option 2
[...]

PR #724, opened 2019-09-11
Feature/685/reset camera when changing map

### LINKED ISSUE #685: Add "Reset Camera when changing map" toggle to the settings
# Feature request

CodeCharta resets the camera whenever the user changes the map. This is a useful feature if the new map is much smaller or larger than the current map. It is inconvenient when the users wants the camera perspective to stay the same, to compare similar maps from the same perspective. 

## Description

Please add a checkbox (default is on) to "Reset Camera when changing map". The user can deselect this checkbox to prevent CodeCharta from resetting the perspective when chaning the map.

JaiJoPa:
# Reset Camera when changing map

closes #685 

## Description

A checkbox is added to the global settings, which disables the camera reset if deselected,  when a new map is loaded into CodeCharta. 

Checkbox_Camera

## Definition of Done

- [x] **All tests pass**


alschmut:
I am just having the idea, that this Issue might not solve the actual problem. The intention was to give the user the option to use multiple maps in one browser tab (instead of having codecharta opened in multiple tabs with multiple maps). I think the issue should actually intent a similar approach to #476, because the user might want to set an individual camera angle for each map and when switching to another map it uses the camera angle of the newly selected map.

Using this PR, all maps use the same camera angle. When moving the camera angle, it will "kind of update" the camera angle for every map.

fatihi:
@alschmut I would have thought the same, but the intent on the feature was, that the user can compare multiple states of the same codebase from the same angle by loading in the different states. Saving individual camera angles for each map would not help in this case

PR #735, opened 2019-09-22
Feature/527/attribute side bar

### LINKED ISSUE #527: Metric Attribute list side bar
# Feature request
Let's incrementally change our UI. Next stop, move the metric attribute list to the right side. 

Currently the metric attribute list can be opened by clicking on the file path indicator. In the settings this is called the detail panel. This issue replaces this detail panel and it can be removed.

## Description
The new metric side bar is useful when users want to check out other metrics beside the three selected ones for size, height and color: "this building has high complexity, bad line coverage but how often is it actually changed?". The features are:

- [x] Is on the right side
- [x] opens if a user has selected a building (orangy), closes if she deselects a buildings or if he presses the close symbol
- [x] will only show metrics for the selected building
- [x] it can be scrolled, just like before, to check out all the other metrics.
- [x] the name and basic metrics are "frozen", so we don't lose this information while looking at other metrics
- [x] the window should be horizontally very slim
- [ ] ~If the user clicks on the file path it opens vertically, like a tree view. Unlike a tree view it can only be expanded fully and minimized fully. And unlike our regular tree view on the upper left side it does not~
- [x] Show the URL in a tooltip when hovering the nodeLink icon
- [x] Hide AttributeSideBar onFileSelectionStateChanged and onBlacklistChanged
- [x] the full path requires a lot of horizontal space. So we won't show the complete file path and instead only the fitting parts. The path is now word-wrap and shown over multiple lines.
- [x] Fix scrolling behaviour when fixed header is getting longer due to long strings visible over multiple lines.
- [x] Move the viewCube component when the sideBar is opened to be on the left of it (see comment below)
- [x] Move the legendPanel when the sideBar is opened to be on the left of it (see comment below)
- [x] For the above metrics table: Show two metrics in a row (see comment below)
- [x] Move the link icon next to the selected nodename
- [x] Move the close icon to the top right corner
- [x] Color the delta value background box according to the `node.deltas["theMetric"]` and `settings.appSettings.invertColors` in light green or red. This would currently mean 4 times duplicated code inside the html file.

 show the complete structure, but only the path to this specific file.

The result looks something like this:

Metric-Bar

While I do like the radial buttons proposed in the screen shot I think we should implement these buttons with a later story.

## Inspiration

I was inspired by this UI mockup:

UI-Refactoring

alschmut:
# Feature/527/attribute side bar

closes #527

## Description

This PR includes quite some changes split into the following three sections. See screenshots below:

#### DetailPanel
- Removed DetailPanel
- Removed option `settings.appSettings.maximizeDetailPanel`
- Added nodePath component which contains the remaining part of the detailPanel to show the path of the hovered building

#### AttributeSideBar
- Added AttributeSideBar component to show selected building information
  - Opens (slides in) when a building was selected
  - Closes (slides out) when building was deselected (click somewhere else, load a new file or change of blacklist)
  - The AttributeSideBar shows
    - close icon (top right)
    - node name
    - link icon
    - full path
    - origin filename
    - selected metrics with its values (plus delta value in delta mode in green/red)
    - all other metrics with its values (plus delta value in delta mode in green/red) in a scrollable box that adjusts to the height of the remaining window space
- Extract green/red delta boxes in a separate component
- Added AttributeSideBarService to update and broadcast information regarding the state of the sideBar (open/closed). This enables other components to move accordingly (viewCube, legendPanel)
- Changed animation to show or hide the legendPanel

#### Refactoring
- Move centerMapButton component into viewCubeComponent so it moves with the viewCube
- Refactor CodeMapMouseEventService and BuildingSelectedEventSubscriber to broadcast two separate events (instead of one combined event)
  - onBuildingSelected with building as a parameter
  - onBuildingDeselected
- Remove unneeded comments in threeViewer classes
- Update and add tests for threeUpdateCycleService
- Simplify metricChooser component with storing the hovered node object instead of every single hovered attribute value
- Refactor outdated `.bind()` usages into arrow functions `() => {}`

Screenshot 2019-09-14 at 20 09 13
Screenshot 2019-09-14 at 20 09 05

## Definition of Done

- [ ] **All tests pass**


alexhunziker:
I like the side bar, it makes the app feel much more interactive & the information more accessible.

While manual testing i noticed two things 

- The rounding of metric values is inconsistent, sometimes metric values are rounded to 1 decimal place, sometimes to 3 decimal places. I think we should display a maximum of one decimal figure.
rounding

- When selecting a building and changing the metric afterwards (while the building is still selected) the attribute side bar stays open, but the building is unselected. Either the side bar should be closed, or the building should stay highlighted
side bar wo selection

fatihi:
Thanks @alexhunziker , I agree. Regarding your second point I think the building should stay selected.

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
```diff
@@ -6,17 +6,29 @@ legend-panel-component {
 		bottom: ($fixedHeight + 20) + px;
 		right: -35px;
 		transform: rotate(90deg);
-		z-index: 2;
+		border-radius: 0 0 5px 5px;
+
+		&.expanded {
```
alexhunziker: Is this expanded for the Attribute side panel, or the legendPanel?

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.ts
```diff
@@ -172,47 +169,32 @@ export class MetricChooserController
...
+			return "green"
+		} else if (heightDelta < 0) {
+			return "red"
 		} else {
 			return "inherit"
 		}
 	}
+
+	private synchronizeAngularTwoWayBinding() {
+		this.$timeout(() => {})
```
alexhunziker: What is this doing?

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.ts
```diff
@@ -133,16 +142,13 @@ export class LegendPanelController implements SettingsServiceSubscriber {
...
-		$(document).ready(() => {
-			let start = 40
-			let target = -500
-			let visible = false
-			$("legend-panel-component .panel-button").click(() => {
-				$("legend-panel-component .block-wrapper").animate({ right: visible ? target + "px" : start + "px" }, "fast")
-				visible = !visible
-			})
-		})
+	public toggle() {
```
alexhunziker: Is there a reason we don't use ng-class or similar instead of this?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.ts
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	public onClickCloseSideBar() {
+		this.attributeSideBarService.closeSideBar()
+	}
+
+	private updateSortedMetricKeysWithoutPrimaryMetrics() {
+		if (this._viewModel.node) {
+			this._viewModel.secondaryMetricKeys = _.keys(this._viewModel.node.attributes)
+				.filter(x => !_.values(this._viewModel.primaryMetricKeys.node).includes(x))
```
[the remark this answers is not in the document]
alschmut: Until now the selected metrics were always shown twice and I really didn't like it. But to be host it also looked different, as it was all in one list (in the detail panel) and now its rather separated. So If we decide democratically, I am fine with changing that. Its also gonna be slightly less code then.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.deltas.component.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+		})
+	})
+
+	describe("onBuildingSelected", () => {
+		beforeEach(() => {
+			attributeSideBarDeltasController["setDeltaColorClass"] = jest.fn()
+			attributeSideBarDeltasController["setDeltaValue"] = jest.fn()
+		})
+		it("should call function setDeltaValue", () => {
+			attributeSideBarDeltasController.onBuildingSelected("mySelectedBuilding" as CodeMapBuilding)
```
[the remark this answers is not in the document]
alschmut: yea It was intended, but I changed it now with a real CodeMapBuilding, which was already used in this file.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.deltas.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+	constructor(
+		private $rootScope: IRootScopeService,
+		private $timeout: ITimeoutService,
+		private threeSceneService: ThreeSceneService,
+		private settingsService: SettingsService
+	) {
+		CodeMapMouseEventService.subscribeToBuildingSelectedEvents(this.$rootScope, this)
+		SettingsService.subscribe(this.$rootScope, this)
+		this.$timeout(() => {
```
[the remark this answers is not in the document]
alschmut: The component uses an angular binding `metricDeltaSelectedComponent.bindings.attributekey` to give each component the needed metricName from within the attributeSideBar.component.html (see line 25 e.x.). The components get created when the delta is activated and when the secondaryMetrics change. Until the binding `attributekey` is set and accessible inside the `metricDeltaSelectedComponent` it takes a few ms. Thats why the timeout is there.

(the code is now inside the new metricDeltaSelectedComponent).

visualization/app/codeCharta/ui/codeMap/codeMap.component.ts
```diff
@@ -41,6 +45,14 @@ export class CodeMapController implements BuildingRightClickedEventSubscriber, L
 		}
 	}
 
+	public onAttributeSideBarVisibilityChanged(isAttributeSideBarVisible: boolean) {
+		if (isAttributeSideBarVisible) {
+			$("view-cube-component").addClass("expanded")
```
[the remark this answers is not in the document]
alschmut: with name do you suggest? `moved-left`?

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
alschmut: this moves the legendPanel to the left when the attributeSideBar is opened. same as with the viewCube. Should we rename the CSS class?

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.ts
alschmut: you will find this function `synchronizeAngularTwoWayBinding` in some components. The angular lifecycle handles the twoWayBinding of data between the controller and the html template. This means angular decides, when it is a good moment to update the html data. For some reason, angular does not always want update the data, especially after events are received and data was changed.

The timeout function (with 0 seconds by default) is a workaround, which indirectly triggers updating the html data.

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.ts
alschmut: What exactly do you think of?

alschmut:
Regarding the first short manual review by @alexhunziker 
- Having the building selected, after a new map was build (e.x. due to any settings change except for the "height scale") is slightly tricky. We have to select the building again if there was one selected (see `threeSceneScerice.getSelectedBuilding()`) after the map was rerendered, because the building, which "knew" it is selected, doesn't exist anymore, because a completely new map was build with the updated settings.
- Rounding the numbers is a good point! I would suggest to round it to 0 decimals.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.scss
```diff
@@ -0,0 +1,163 @@
...
+				.metric-delta {
+					border-radius: 3px;
+					padding: 1px 3px;
+				}
+
+				.metric-value {
+					background: #d5d5d5;
+				}
+
+				.metric-delta {
```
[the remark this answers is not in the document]
alexhunziker: I also moved the margin-left to the new component

visualization/app/codeCharta/ui/codeMap/codeMap.component.ts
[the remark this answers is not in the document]
alexhunziker: yes I think `moved-left` would be fine, or something `like side-bar-expanded` would be fine too imho.

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
alexhunziker: Yes, I would use the same name as used for the viewCube CSS class

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.ts
alexhunziker: essentially instead of using JQuery, i used angularjs ng-class to add/remove the class based on an attribute in the viewModel. Personally I think it is nicer, if you don't like it you can revert the two commits in question ;)

alexhunziker:
> Having the building selected, after a new map was build (e.x. due to any settings change except for the "height scale") is slightly tricky. We have to select the building again if there was one selected (see threeSceneScerice.getSelectedBuilding()) after the map was rerendered, because the building, which "knew" it is selected, doesn't exist anymore, because a completely new map was build with the updated settings.

I am also fine with just closing the SideBar whenever a new Map is built. (Especially since there may be the case that the selected building is not visible in the new map anymore anyways (e.g. new AreaMetric is 0 for the selected building)...) @fatihi what do you think?

fatihi:
I would prefer if the building stays selected. Although if that is too complicated, I'm fine with the sidebar closing

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.ts
alschmut: actually I like it a lot! Thanks :)

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
alschmut: Done

visualization/app/codeCharta/ui/codeMap/codeMap.component.ts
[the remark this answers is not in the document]
alschmut: Done

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.ts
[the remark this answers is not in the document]
alschmut: after talking to @fatihi and @NearW we decided to keep it as is for now.

alschmut:
> I would prefer if the building stays selected. Although if that is too complicated, I'm fine with the sidebar closing

I agree and I would suggest to open a new bug-issue for that, as its from a tech side not related to this PR.

alschmut:
@alexhunziker do you wanna check the last commits again, have a short manual look, give some feedback and then we can potentially merge this (:

alexhunziker:
@alschmut the code looks fine, however if we postpone keeping the building selected, we should close the attribute side bar when a new map is built.

alschmut:
> @alschmut the code looks fine, however if we postpone keeping the building selected, we should close the attribute side bar when a new map is built.

Even although this were the expected rational behaviour, I would not do so, because then the attributeSideBar will be closed every time you change some settings, which probably means it will barely be visible. The bug is not about the attributeSideBar, but with the building selection.

alexhunziker:
Yes, the bug is not with the attribute side bar. However I would still close it, whenever the building on the map is unselected, and then change the behaviour once the underlying bug is fixed, since otherwise we are introducing inconsistent behaviour.

PR #739, opened 2019-09-29
Tech/add tests

alschmut:
# Tech/add tests

## Description

- Refactor dialogDownload component functions into multiple smaller functions
- Add tests for dialog component
- Add test for MapTreeView component

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+			expect(dialogGlobalSettingsController["_viewModel"].resetCameraIfNewFileIsLoaded).toBeFalsy()
+		})
+	})
+
+	describe("applySettings", () => {
+		it("should call settingsService.updateSettings", () => {
+			settingsService.updateSettings = jest.fn()
+			const viewModel = { hideFlatBuildings: false, resetCameraIfNewFileIsLoaded: false }
+			dialogGlobalSettingsController["_viewModel"] = viewModel as any
```
NearW: Instead of doing these things, add a function `setViewModel` to the test, that takes in one parameter for each field and sets it. This should shorten the code and remove the anys

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.ts
```diff
@@ -23,6 +23,7 @@ export class DialogGlobalSettingsController implements SettingsServiceSubscriber
 	}
 
 	public onSettingsChanged(settings: Settings, update: RecursivePartial<Settings>) {
+		//TODO: dead code. Subscription does not work with "controllerAs $ctrl"
```
NearW: What's up with this TODO?

visualization/app/codeCharta/ui/dialog/dialog.download.component.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+			return {
+				getFileStates: jest.fn().mockReturnValue(FILE_STATES)
+			}
+		})()
+	}
+
+	function getFilteredFileContent(name: DownloadCheckboxNames) {
+		return dialogDownloadController["_viewModel"].fileContent.find(x => x.name == name)
+	}
+
```
NearW: Can we get a wrapper here that all these tests happen on construction? You can spot that by seeing that only `rebuildController` is called as an act, but it would be great to know this by reading it from top to bottom.

```
describe("constructor", () => {
    describe("amoutOfAttributeTypes", () => {
        it("should be set correctly", () => {}
    }
}
```

etc

alschmut: Do you think of one describe-block with the string "constructor" that wraps all tests?

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.ts
alschmut: Removed the TODO as the subscription works. Side note, which is also mentioned in the commit description:
- the constructor is called every time the dialog is opened
- and events only arrive when the dialog is opened

visualization/app/codeCharta/ui/dialog/dialog.download.component.spec.ts
NearW: Yes, exactly. And rename the other describe blocks to the field of the viewmodel you're currently testing.

alschmut: Did it.

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
NearW: The viewModel is empty per default.

Change
```
function setEmptyViewModel() {
		dialogGlobalSettingsController["_viewModel"] = {
			hideFlatBuildings: null,
			maximizeDetailPanel: null,
			isWhiteBackground: null,
			resetCameraIfNewFileIsLoaded: null
		}
	}
```
to
```
function setViewModel(hideFlatBuildings: boolean, maximizeDetailPanel: boolean, isWhiteBackground: boolean, resetCameraIfNewFileIsLoaded: boolean) {
		dialogGlobalSettingsController["_viewModel"] = {
			hideFlatBuildings,
			maximizeDetailPanel,
			isWhiteBackground,
			resetCameraIfNewFileIsLoaded
		}
	}
```

so that this

```
it("should call settingsService.updateSettings", () => {
			settingsService.updateSettings = jest.fn()
			setEmptyViewModel()
			dialogGlobalSettingsController["_viewModel"].hideFlatBuildings = false
			dialogGlobalSettingsController["_viewModel"].resetCameraIfNewFileIsLoaded = false

			dialogGlobalSettingsController.applySettings()

			expect(settingsService.updateSettings).toHaveBeenCalledWith({ appSettings: dialogGlobalSettingsController["_viewModel"] })
		})
```

will become

```
it("should call settingsService.updateSettings", () => {
			settingsService.updateSettings = jest.fn()
			setViewModel(false, null, null, false)

			dialogGlobalSettingsController.applySettings()

			expect(settingsService.updateSettings).toHaveBeenCalledWith({ appSettings: dialogGlobalSettingsController["_viewModel"] })
		})
```

Another approach would be to make this function more abstract, so that the first parameter sets the value of the first key in the viewmodel. We could use that function in all controllers to setup the viewmodel in one line for a test.

Something I'm concerned about though:
It shortens code, if you need to setup more than one field of the viewModel, but if we only need one field of the viewModel for our test to work, it's probably easier to read when we write `myController.viewModel["key] = foo`
What do you think?

alschmut: I actually prefer it the other way around. Your approach might be less code, but it takes longer to understand and change the parameters if needed.

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+	function withMockedSettingsService(mockSettings: Settings = settings) {
+		settingsService = jest.fn<SettingsService>(() => {
+			return {
+				getSettings: jest.fn().mockReturnValue(mockSettings)
+			}
+		})()
+	}
+
+	function setEmptyViewModel() {
```
NearW: We don't need to set an empty view model because it's empty on construction anyway.

Remove this function and it's calls.

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
NearW: Alright. But at least see my latest comment and remove the `setEmptyViewModel` function

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
alschmut: True now we don't need it anymore.

alschmut:
```
ERROR in /home/travis/build/MaibornWolff/codecharta/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.spec.ts
2647(28,27): error TS2554: Expected 2 arguments, but got 1.
```

I don't really get why this error occurs. The parameters look fine to me.

PR #759, opened 2019-10-24
Feature/607/move path panel

### LINKED ISSUE #607: Move path indicator into the file bar
# Feature request

## Description

The file path indicator takes a lot of space and risks being overlayed by other controls. How about we place it where the file bar is normally? We don't need the file bar while hovering over buildings.

The result could look like this, with a clear separation between the folders and files:

file-path-bar

* Please ignore the fancy icons. We don't need them. Regular file and folder icons are perfect.
* If the path is wider than the available space we'll wrap at the end and use two or more lines to display the path. That's the same behavior as our current path indicator.

alexhunziker:
# Move Path Panel to ToolBar

Depends on #735 which needs to be merged first
closes #607

## Description

- Show File Path when file is hovered, and file panel if nothing is hovered.
- Move File Path to ToolBar
- Use folder/file icon for hovered node

cc_file_path

## Definition of Done

- [x] **All tests pass**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Changed
 
 - Animation to show or hide the legend panel #527
+- Moved nodePathPanel to toolBar and updated style #607
 
 ### Removed
 
--- a/visualization/app/codeCharta/codeCharta.component.html
+++ b/visualization/app/codeCharta/codeCharta.component.html
@@ -20,8 +20,6 @@
 
 <legend-panel-component></legend-panel-component>
 
-<node-path-panel-component></node-path-panel-component>
-
 <loading-gif-file-component></loading-gif-file-component>
 
 <div id="mw-logo">
--- a/visualization/app/codeCharta/ui/nodePathPanel/nodePathPanel.component.html
+++ b/visualization/app/codeCharta/ui/nodePathPanel/nodePathPanel.component.html
@@ -1,3 +1,9 @@
 <div id="node-path">
-	{{ $ctrl._viewModel.hoveredNodePath }}
+	<span ng-repeat="folder in $ctrl._viewModel.hoveredNodePath track by $index">
+		<span role="img" class="fa fa-folder"></span>
+		{{ folder }} >
+	</span>
+
+	<span role="img" class="fa fa-file-o"></span>
+	{{ $ctrl._viewModel.hoveredNodeName }}
 </div>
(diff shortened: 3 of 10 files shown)


alschmut:
Thank at @alexhunziker! ;) 

After a first visual look at the gif, I would maybe try to check out the following:
- Make the folder/file icons less dark (meaning rather dark grey instead ion black). Maybe the same color as the area/height/... icons next to the metricChooser
- Same with the font-color. I suggest a less dark color looks more seamless
- Instead of the arrow `>` maybe use an fa icons. Maybe this one?
  `<i class="fa fa-angle-right"></I>`

alschmut:
I was adding my feedback from above myself 😃 I think this looks cleaner, especially with large paths.


- - -

Although in my opinion we still need to figure out a way to present long paths, as I don't really like the toolbar to grow in height. What might look nice is the following. So the path is vertically centered and can show at least two lines in a nice way, without changing the toolbar height.

**The following two screenshots are not implemented!** I did not find out how to center the text vertically), maybe someone else has an idea.

Screenshot 2019-10-27 at 18 06 52
Screenshot 2019-10-27 at 18 06 40

fatihi:
I agree with @alschmut that using the icon looks cleaner 👍 

Regarding long paths: I'm also not a fan of the height of the toolbar growing for long paths. I suggested doing it like IntelliJ, cutting the left part of the path

alexhunziker:
> Regarding long paths: I'm also not a fan of the height of the toolbar growing for long paths. I suggested doing it like IntelliJ, cutting the left part of the path

Just implemented it

alexhunziker:
Regarding the last commit f7c1608 (refactoring): When I implemented it first I also thought about whether or not I should move the if statements to the html. I decided to keep them in the controller (or rather the distinction between the name and the path), since I thought the html is easier to understand that way, and the number of if satements to be evaluated is lower (2 per total path vs. 3 per path component) (Not that that would matter for the performance though). What were your thoughts when you moved the logic to the ng-if's, respectively, why do you prefer this?

alschmut:
@alexhunziker Yea, I mostly thought about the controller to be less full of logic, and in my opinion if one knows what the $last in the html ng-repeat does, it is pretty understandable. Maybe we get a third opinion about this? If you are convinced to revert the commit, I am fine with that as well.

alexhunziker:
I guess it is a matter of preference... Personally, I would still put it in the controller to reduce the number of if statements in the html. However, in the end it doesn't make that much difference here imo, so we can keep it and merge the branch as it is from my point of view.

Issue #350, opened 2019-02-15
CodeCharta json 2.0


Richargh:
A place to design the new **codecharta.cc.json**. This issue description should always contain all commented ideas and be up to date!!!

- **`analysis`** is data from the ccsh importer. This will never be updated!
- **`settings`** can be updated by the user. Enriches the analysisData
- The `color`'s value can either be a ID number, which matches to our default color scheme or a hex-string value

### In short
``` json
{
  "fileMeta": {},
  "analysis": {
    "map": [],
    "edges": [],
    "attributeTypes": {},
    "timeTravel": [],
    "attributeDescription": [],
  },
  "settings": {
    "fileSettings": {
      "excludedNodes": [],
      "flattenedNodes": [],
      "markedPackages": []
    }
  },
  "scenarios": [{
    "name": "Complexity",
    "settings": {
      "dynamicSettings": {},
      "appSettings": {}
    }
  }]
}
```

### In detail
``` json
{
  "fileMeta": {
    "apiVersion": "1.1",
    "createdOn": "2019-02-21T19:50:10+00:00",
    "updatedOn": "2019-03-01T19:50:10+00:00"
  },
  "analysis": {
    "map": [{
      "name": "root",
      "type": "Folder",
      "attributes": {},
      "link": "",
      "children": []
    }],
    "edges": [{
      "fromNodeName": "/root/src/de/Foo.java",
      "toNodeName": "/root/de/maibornwolff/codecharta/Bar.java",
      "attributes": {
        "averageRevs": 9,
        "pairingRate": 100
      }
    }],
    "attributeTypes": {
      "edges": [{
        "name": "pairingRate",
        "type": "average"
      }, {
        "name": "averageRevs",
        "type": "sum"
      }],
      "nodes": [{
        "name": "loc",
        "type": "sum"
      }, {
        "name": "testCoverage",
        "type": "average"
      }]
    },
    "timeTravel": [{
      "nodeName": "root/src/de/maibornwolff/Blub.java",
      "previousNodeNames": [{
        "timestamp": "2018-10-12T12:10:15+00:00",
        "name": "CoolStuff.java"
      }]
    }],
    "attributeDescription": [{
      "name": "mcc",
      "displayName": "Code Complexity",
      "longDisplayName": "Complexity by McCabe",
      "description": "The code coverage measures blah blah and is calculated by blah blah"
    }],
  },
  "settings": {
    "fileSettings": {
[...]

PR #784, opened 2019-11-13
Feature/691/redesign blacklist panel

### LINKED ISSUE #691: Make the purpose of hide/exclude clear
# Feature request

I am pretty sure that the purpose of hide an exclude is not clear to our users. Let's change that by adding some descriptive text to the hide and exclude dropdowns. I propose three things as seen below.

**Intended** as a quick-fix: if this is not simple to do within **a couple of hours** then please 1) don't do it, 2) comment why it's difficult (tech issues, test issues, unclear goals, etc.) 3) let's figure out together if we should do it.

## Description

1) I think the hide and exclude Dropdowns should start with a simple Headline (please comment or change the wording for clarity):

--------

### Hidden (or should we change this to "Flattened"?)
Hidden buildings are flattened and are included in folder metrics.

### Excluded
Excluded buildings are not shown in the map, but still stored in the file and not included in folder metrics.

--------

2) Below that is an underlined description "add with search or node context menu"
I think the underline makes it look like this is a search field, so we should probably remove that line. Also I think this message should only be displayed, when the list is empty.

3) When an item is exclude or hidden it has the exclude or hide symbol on the left and trash icon on the right: 

We don't need the exclude or hide symbols any more, because we split the backlist into hide and exclude. The trash symbol is confusing because our users are afraid, that clicking it might lead to data loss. Also the trash symbol is on the right side, which means we need to scroll all the way to the right to reshow an excluded or hidden item with a long name.

My idea is to remove trash, exclude and hide symbols. The latter two are not needed. The effect of the trash first one is to **reshow** whatever was excluded/hidden. So instead we'll place left of the hidden/excluded text an eye symbol to indicate we can show the item again.

alexhunziker:
# Update Blacklist Panel

closes #691

## Description

- Optical clean up of the blacklist panel
  -  Remove the hide/exclude icon on the left side of the pattern as it convayed no useful information, since the types are separated in two lists
  - Change the trash can icon to a minus icon and move it to the left side of the pattern
  - Crop the pattern visibility on the left side if the pattern is too long
- Rename Hide to Flatten, and also change the behaviour accordingly. The behaviour is therefore consistent with the flattend buildings during search, and the option to flatten buildings with no edges. This also makes the difference between hide and exclude clearer.
- Hide flattened buildings in the settings-menu is turned off by default now

## Definition of Done

- [x] **All tests pass**


alexhunziker:
just fyi: I also changed the BlacklistType: "hide" to "flatten" --> "hides" stored in old downloaded files will not show after this PR

Richargh:
> just fyi: I also changed the BlacklistType: "hide" to "flatten" --> "hides" stored in old downloaded files will not show after this PR

Sorry for chiming in but this topic is very dear to me. Please do not make such a breaking change lightly. Evil as I am I made some promises about our [cc.json versioning](https://maibornwolff.github.io/codecharta/docs/versioning/). The correct thing to do would be to incrememt our version from 1.38 to 2.0 due to this breaking change. Do you want to do that or do you suggest that this is not a breaking change? 

An alternative is to read the old "hide", convert it to "flatten" and when saving store it as "flatten". That code would stay in CC for a long time though until we remove it. I assume more than a year until all users have reopened all their `cc.json` maps again.

alschmut:
Although I would like to keep everything consistent, I have to agree with @Richargh. This is a breaking change. I think for now we should keep the old names `exclude` and `hide` for the downloaded cc.json and consider changing the names, when we are willing to do the breaking change potentially with a cc.json 2.0 in #350.

Nevertheless for the future, I am actually a fan of adding a layer to translate old versions to new ones. I am not sure though, if this should be placed in the visualization automatically or the ccsh as a separate filter.

alexhunziker:
Fair point, this issue should not cause a breaking change.

So which one do you prefer, that BlacklistType.flatten is stored in the cc.json as "hide" as @alschmut suggests or convert "hide" to "flatten" when loading a new map?

alschmut:
I was working again a bit on the design and:
- added a thick header, showing which tab we are in
- added a question mark icon, which displays what flattened/excluded means on hover 
- split the html of the blacklistPanel into two separate html files (tech). I wasn't sure, if we should split the component completely, as they still share some similarities (e.x. the style sheet).

What do you think about it? In my opinion it now looks nicer and less like a paragraph of text :)

alexhunziker:
I like the headers and the tooltip :) also the splitting into two HTML files makes sense. Personally i woud keep it this way and not split the component completely, as they share the appearance and logic part almost completely.

Also I have added a function that converts old blacklist types when .cc.json files are loaded

alschmut:
Are we sure, we want to store the word `flatten` as the new type in the cc.json, when we export a map? In my opinion, we should translate the type from `flatten` to `hide` before we download it as well.

alexhunziker:
So the renaming in the downloaded files would mean that you could not see flattened buildings exported from a new version of the visualization in an old visualization. While this is probably a really rare, or maybe even never occuring use case, in my opinion, you are technically correct that this should not happen. I therefore pushed an update, which takes care of that.

Issue #698, opened 2019-08-17
Split event management into multiple services


alschmut:
# Feature request

## Description

Since we merged #671 we want to split the settingsService with its events into multiple services, each storing only one property or a few combined ones.

### Ideas
- Use the current Settings interface structure to adapt folder structure
  - Rename the `settingsService` folder into `settings`
  - Add the following folders into the `settings` folder:
    - appSettings
    - dynamicSettings
    - fileSettings
    - treeMapSettings? Maybe we find a better solution where to store the `mapSize` property
- Each property, which already has its own event will be separated into a separate service inside the correct settings sub folder. The Service contains its event subscriber interface and its static event name.
- The `SettingsService.updateSettings()` will not be called from within a component. It only listens to the other new services and updates its settings bundle.

### Before starting, some Questions:
- Think about how we could implement #476 with the new settings services. How do we store multiple settings, if we have multiple services?
- We should finally check, if RxJS makes our life easier with lots of events.

alschmut:
I was taking some moments to rethink our intent to split the settingService together with ideas to store multiple settings version for different maps or even a history of settings for each file.

Instead of:
- splitting every single property into a separate service.
- collecting data from every service into a all encompassing settingService (in case many of the properties are needed)

we could:
- keep the settingsService as a central and structured data storage
- create separate services for each settings-property* (without storing the data itself!)
- each service will 
  - provide a method to update the property and store the data inside the settingsService, 
  - collect the property data from the settingsService
  - broadcast a separate event for only the one property
- the settingsService will ideally not broadcast any event with the complete settings object

=> this means the UI components and other services only access the property-services
=> this also makes it easier to manage multiple settings, as only the settingsService could provide a blackBox storage of map or time-dependant settings, without having to update every of the separate services

An example of such a service is shown below:

###### * a property is for example the `areaMetric` or the `amountOfEdgePreviews`

```
import { SettingsService } from "../settings.service";
import { IRootScopeService } from "angular";

export interface AmountOfEdgePreviewsSubscriber {
	onAmountOfEdgePreviewsChanged(amountOfEdgePreviews: number)
}

export class AmountOfEdgePreviewsService {

	private static AMOUNT_OF_EDGE_PREVIEWS_CHANGED_EVENT = "amount-of-edge-previews-changed-event"

	constructor(private settingsService: SettingsService) {
	}

	public getAmountOfEdgePreviews(): number {
[...]

PR #792, opened 2019-11-16
Tech/790/decorate node path earlier

### LINKED ISSUE #790: Set nodepath when importing files
# Feature request

## Description

Currently we add the `path` attribute every time when a new map is rendered (after every change of settings). The path should be predecorated just when new files are imported. Two reasons apply:
1. The path of a node never changes, so It makes sense to add it only once
1. Currently when a new map is generated, some code is always executed twice after a new map is shown, because it needs the path, which is not available with the first round (in detail: inside the edgeMetric.service the metrics get updated, using the CodeMapHelper, which needs the nodes path, `onFileSelectionChanged` and `onBlacklistChanged`. The first event always triggers the second, which means only onBlacklistChanged the code can actually run as intended)
1. Even if the impact might not be huge, it probably speeds up the selection process when showing different loaded maps.

alschmut:
# Tech/790/decorate node path earlier

closes #790

## Description

- Decorate node path in codeCharta.service when importing files
- Remove path decoration in preRenderService every time a new map was rendered
- Deep Copy File states on getFileStates() to prevent pass by reference

## Definition of Done

- [x] **All tests pass**


alexhunziker:
Generally only create the paths when they could have changed seems like a good idea :)

grafik

However, for some reason the file name does appear now twice within the path.

alschmut:
So while trying to find a solution for the wrong path, I uncovered a much bigger issue:
- We think, we store only the most original data we get from the imported files, inside the `fileState.service`, but it turns out, as JavaScript allows to mutate passed objects (kind of like call by reference), we overwrite those fileStates now and then (e.x. now the path, when we use the `multiple` mode)

Background regarding this path-issue:
- when transforming multiple maps into one, we add another top node and ADD the fileName into the path. This overwrites the fileStates.
- Showing a file in single mode afterwards, shows all nodes with the same path (including the fileName), instead of only the pure node-path 
- And when we use the multiple mode afterwards again, it ADDs one more fileName to the path

So the options we have:
1. Forget about the issue and not merge this PR, as paths are always set (again) just before building the map
1. Implement a quick fix solution to the path problem.
1. always use `_.cloneDeep` for the fileStates to pass object by value. This probably slows down performance, but it's the safe option to not mutate the objects accidentally 
1. Check the whole project to not mutate objects (meaning reassign internals of an object), but always use `_.cloneDeep` for the parts, where necessary.

My favourite is probably a mix of number 3 and 4, maybe to provide two methods, one which returns the fileStates by value and the other by reference.

alexhunziker:
Yes I agree, it probably makes sense to clean this up and use _.cloneDeep where appropriate. Offering the two functions could be useful, to get a clear whether we are getting a reference or a copy.

fatihi:
I'm also in favor of cleaning up properly. Performance is an issue though, that's why I would like to see how much performance it really costs. How long would it take to find every part where you mutate an object and change it?

btw: You don't have to use lodash for that, you can also just use the destructuring notation:

`javascript
const newObject = {
  changedProperty: changedValue,
  ...oldObject
}
`

This is also used a lot in redux when reducers create a new state as shown here:
https://redux.js.org/recipes/using-object-spread-operator

alschmut:
Thanks @fatihi truely, lodash is not needed, but I would personally suggest, to first focus on #698 as @NearW started implementing a working approach using redux.

fatihi:
Would the implementation of redux also take care of this problem? I thought that they were two independant problems

alschmut:
They are two independent problems, but they both will modify multiple ui-components.

fatihi:
I see. If that is your recommendation, I'd say we put this issue on hold

visualization/app/codeCharta/state/fileState.service.ts
```diff
@@ -27,7 +29,12 @@ export class FileStateService {
 	}
 
 	public getFileStates(): FileState[] {
-		return this.fileStates
+		return _.cloneDeep(this.fileStates)
```
NearW: remove deepClone. DeepCloning can cost some time and we should only mutate the fileStates with the methods provided by this service (or later with the actions defined for this part of the store)

alexhunziker: I just tested it with one of the largest maps we have, and i could not tell a performance difference compared to the current master. There also only remain 3 cases in which the FileState needs to be obtained; one of which needs a deep copy anwyays (see bug above). Personally i kind of like that with the deep clone we cannot alter file states by accident. If you dislike it conceptually, we can remove it from this method and deep clone on the recieving end where necessary.

alexhunziker: ...well changed my mind; in the event we get it by reference anyways, and there is no point in preventing pass by reference when getting it while we pass it by reference with the action. I will remove deep clone and add it in the preRenderService, where we actually need it.

visualization/app/codeCharta/ui/codeMap/codeMap.preRender.service.ts
```diff
@@ -110,11 +110,8 @@ export class CodeMapPreRenderService implements StoreSubscriber, MetricServiceSu
 	}
 
 	private getSelectedFilesAsUnifiedMap(): CCFile {
-		const fileStates: FileState[] = this.fileStateService.getFileStates()
+		const fileStates: FileState[] = _.cloneDeep(this.fileStateService.getFileStates())
```
NearW: why do we clone here?

alschmut: there was a reason. Is there some useful commit message?

alexhunziker: Well, we mutate the FileState that was passed by reference here. I guess whoever implemented this did not realize that this will also change the object in the FileStateService. Until this PR it did not really matter that we did that, since the map was redecorated completely anyways. However, if we predecorate the Paths earlier, it starts to matter. What happens without the deepClone can be seen in my first comment to this PR.

NearW: Alright. We should address this issue later. I think it makes sense to not return value from the decoration-methods and just mutate the parameter indicating that there are obviously side effects to it.

For now, let's keep the cloneDeep. Be careful with that in production code though. :)

PR #827, opened 2020-01-15
Feature/660/raw text parser (indentation level parser)

### LINKED ISSUE #660: New Importer: IndentationLevelImporter
# Feature request

## Description

Create an importer that counts the indentation level of each line and generates an cc.json output with attributes like this:
```
{
    "il1+" : 1,
    "il2+" : 3
}
```
where il1 is the number of lines in the file with `indentation level 1 or higher`.

The `indentation level` is calculated by
```
(positionOfTheFirstCharacterInLine / tabWidth) + (positionOfTheFirstCharacterInLine % tabWidth == 0 ? 0 : 1)
```

The `tabWidth` needs to be estimated before.

The importer should be executable with:
`ccsh indentationlevelimporter FOLDER/FILE -p PROJECT_NAME -o OUPUT.CC.JSON`

alexhunziker:
# Created RawTextParser

closes #660 

## Description

- Added a new parser; name currently RawTextParser (if you have objections to the name and have a better idea: let me know)
- This new parser can parse files/projects and parses them irrespective of their file extension/programming language. It comes with a number of options (excludes, fileExtensions, pipe support, metric selection ...)
- Added metric indentationLevel (as described in the issue) to this new parser. It either guesses the tabWidth, or it can be specified. Furthermore the max indentationLevel can be altered via the CLI.
-  Some minor refactorings

{Descriptive pull request text}

## Definition of Done

- [x] **All tests pass**


analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/model/FileMetrics.kt
```diff
@@ -0,0 +1,10 @@
+package de.maibornwolff.codecharta.parser.rawtextparser.model
+
+class FileMetrics {
+    val metricMap = mutableMapOf<String, Double>()
+
+    fun addMetric(name: String, value: Int): FileMetrics {
```
NearW: Why is the map of type double when you can only add ints to it and convert them to double later?

NearW: If I understand this correctly this is due to indentationLevel being an integer 100% of the time and we want to export doubles in the cc.json.
However, if we decide to add another metric that cna be a double, you'd have to change the method signature. I'd suggest to change the method signature to `value: Double`

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/metrics/Metric.kt
```diff
@@ -0,0 +1,18 @@
+package de.maibornwolff.codecharta.parser.rawtextparser.metrics
+
+import de.maibornwolff.codecharta.parser.rawtextparser.model.FileMetrics
+
+interface Metric {
+
+    val name: String
+    val description: String
+
+    fun setParameters(parameters: Map<String, Int>) {
```
NearW: are the { } required here?

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/metrics/Metric.kt
```diff
@@ -0,0 +1,18 @@
+package de.maibornwolff.codecharta.parser.rawtextparser.metrics
+
+import de.maibornwolff.codecharta.parser.rawtextparser.model.FileMetrics
+
+interface Metric {
+
+    val name: String
+    val description: String
+
```
NearW: Why doesn't the interface declare `value` as well here?

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
```diff
@@ -0,0 +1,60 @@
...
+                    projectMetrics[standardizedPath] = parseFile(it)
+                }
+            }
+        }
+        return projectMetrics
+    }
+
+    private fun parseFile(file: File): FileMetrics {
+        val metrics = MetricsFactory.create(metrics, parameters)
+        file.readLines().forEach { line -> metrics.forEach { it.parseLine(line) } }
```
NearW: I don't think a metric should parse a line. Maybe a MetricVisitor can do that and then set a value in the actual metrics once it decided what the final metric for the file will be. If I remember correctly, sonar did it similar.

This basically means to split up the Metric-Interface into an actual Metric-Interface and a Metric-Visitor that supports methods such as `parseLine` or `setParameters`

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/metrics/Metric.kt
alexhunziker: Why should it be? of course you could do declare `value` and then override the get() method for each metric. But I don't think that this would make it easier as the value property may (as in the case at hand) never hold a value and just contain the current getValue() function as an overridden getter.

(Conceptually you should not write trivial getters/setters in Kotlin and instead refer to the property value directly. However, getValue() does not simply access the state, but instead actually performs an action (e.g. aggregation of data to a value), therefore I implemented it as a function. If you don't like getValue() we may change it to computeValue()/retrieveValue())

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/metrics/Metric.kt
alexhunziker: The { } provides a default implementation of the function. If you have a metric that does not require to set parameters, you are not forced to override the function with an empty function.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/model/FileMetrics.kt
alexhunziker: You are right, it is strange to only provide addMetric with Int, when in fact fileMetrics contain double values. I will solve this by providing an addMetric(name: String, value: Double) function in addition to the current one.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
alexhunziker: Personally I think it is unnecessary to split the Interface, since the resulting Metric classes will do nothing to very little. However, we can rename the metric classes to reflect better reflect what it is doing. Something like MetricCalculator or similar.

Or how would you have suggested to split the Interface

PR #840, opened 2020-01-24
Tech/performance remove deep clone

NearW:
# Improve performance when loading a new map

## Description

- Remove deep cloning of file states when decorating a map
- Decorating a map now decorates the input without cloning the input and returning the decorated clone
- Use rfdc instead of lodash deepClone in production code

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.reducer.ts
```diff
@@ -10,7 +9,7 @@ export function blacklist(state: BlacklistItem[] = setBlacklist().payload, actio
 		case BlacklistActions.REMOVE_BLACKLIST_ITEM:
 			return removeItemFromArray(state, action.payload)
 		case BlacklistActions.SET_BLACKLIST:
-			return _.cloneDeep(action.payload)
+			return [...action.payload]
```
BridgeAR: This results in a shallow clone. Is that intentional?

NearW: This is intentional. redux uses immutable update patterns for performance. 

https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns/

BridgeAR: Sorry, I can't follow. The payload does not seem to be changed in this function. Is the shallow clone still necessary? I would expect no cloning to be necessary?

visualization/app/codeCharta/state/store/dynamicSettings/colorRange/colorRange.reducer.ts
```diff
@@ -1,11 +1,11 @@
 import { ColorRangeAction, ColorRangeActions, setColorRange } from "./colorRange.actions"
 import { ColorRange } from "../../../../codeCharta.model"
-import _ from "lodash"
+const clone = require("rfdc")()
 
 export function colorRange(state: ColorRange = setColorRange().payload, action: ColorRangeAction): ColorRange {
 	switch (action.type) {
 		case ColorRangeActions.SET_COLOR_RANGE:
-			return _.cloneDeep(action.payload)
+			return clone(action.payload)
```
BridgeAR: It would be great to document for what reason the cloning is done. The code itself does not provide any hint why it's necessary. In most cases I would expect a cloning to happen right before code that mutates parts of the object. So documenting it would be good (I recommend to remove the cloning of an individual function and to run the test suite afterwards to identify all places that do not work properly anymore. Otherwise it might be difficult to identify the actual reason retrospectively).

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.reducer.ts
NearW: Redux uses pure functions as reducers. Therefore the state object itself should not be mutated under any circumstances and instead return a new state.

Not shallow copying the previous state would result in mutating it and breaking the pure function. We use a shallow copy of the previous state and simply add the new item aka the payload there.

visualization/app/codeCharta/state/store/dynamicSettings/colorRange/colorRange.reducer.ts
NearW: This is something, that I need to look into again. Setting a new state by using the whole payload should not require any cloning. However, if there still exists another reference to the colorRange object, you can mutate the state without dispatching an action into the storeService.

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.reducer.ts
NearW: Oh god, ignore my previous comment. I thought this was the ADD_BLACKLIST_ITEM action.

Shallowcopying the blacklist here doesn't make any sense. It may make sense in cloning it though as done previously (see comment below)

This change needs to be reverted.

PR #856, opened 2020-02-04
Bug/798/marked packages not imported

### LINKED ISSUE #798: MarkedPackages do not get saved when loading files
# Feature request or Bug

## Description

Currently, when we load files, which contain previously marked packages, into codeCharta the `markedPackages` are not saved in the `CCFile` (see [codeCharta.service.ts#L42](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/codeCharta.service.ts#L42)).

As the markedPackages have not been exported yet, if we decide to import them, I would suggest to clean the interface first in order to not export the `markedPackages.attributes`:

- Lets show the full path inside the legend panel and crop the left side using some CSS. Then we do not need to create a second cut path (see `markedPackages.attributes["name"]` in the `legendPanel.ts`). 
- This will decrease some code inside the legendPanel, but also remove the 3 dots inside the legenPanel, when a path is too long.
- Afterwards we can remove the `markedPackages.attributes`

alexhunziker:
# Bug/798/marked packages not imported

closes #798 

## Description
- Removed attributes field from MarkedPackage
- Shorten name of marked packages by CSS in Legend panel
- Load marked packages into CCFiles

## Definition of Done

- [x] **All tests pass**


alschmut:
Code looks good 👍 unfortunately the CSS style `direction: rtl;` makes problems in Safari (see screenshot). This seems like a common issue (but for some reason within the nodePathPanel in works)

alexhunziker:
could you send me the map you tested with, I can't reproduce the problem...

alschmut:
@alexhunziker did you check with safari? I just used the sample1.cc.json map:
- marked any package
- open the legendPanel
- get the added markedPackage path element from the browser tools
- changed the path to make it longer
- and the same problem appeared (the ellipsis comes first and the slash last, while the first part of the path is shown instead of the last)

I say its not an issue with the map, its pure css for the path string. 

@NearW can you check with your laptop using Safari?

NearW:
I can confirm what @alschmut said. In safari, the start of the path is displayed instead of the end.

NearW:
I just merged the master here and a bug occured:
1. mark ParentLeaf in sample1 and download
2. mark ParentLeaf in sample2 and download
3. open both downloaded files
4. switch single file from sample1 to sample2
=> map buggy

alschmut:
@NearW it actually works fine for me. What exactly is buggy?

NearW:
@alschmut oh boy, I spent way too much time trying to figure out, why this was the case. It somehow only happened, when I was loading the amazon map inbetween.
I found out, that the map was black and empty, because the areaMetric selected is 0 on every building. Therefore nothing was shown. It only happened when loading the amazon map inbetween, because the map has other metrics and therefore the default metric would change on a map change.

The behaviour is currently like that: When you select from a metric > 0 to a metric that is 0, the map stays the same. When a map is loaded into codecharta however and selects the map with 0 per default, only the floor of the map is visible.

Maybe we try to catch that case and display a message, that no map can be shown, if a 0-metric is selected and also never select a 0-metric as a default.

alschmut:
@NearW I am not sure I can follow your bug description 😃 
- First, do you see the bug on master as well? Then I would rather create some new issue.
- Second, when import other files, where the current selected metrics are not available it doesn't use metrics that are zero (see e.x. [areaMetric.service.ts#L37](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/state/store/dynamicSettings/areaMetric/areaMetric.service.ts#L37) and the exact check in [metricHelper.ts#L13](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/util/metricHelper.ts#L13))

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
```diff
@@ -65,6 +66,16 @@ legend-panel-component {
 			.set-left-margin {
 				margin-left: 18px;
 			}
+
+			.marked-package {
+				overflow: hidden;
+				vertical-align: middle;
+				direction: rtl;
```
BridgeAR: The direction should probably not be changed. This is normally only necessary for languages that are read right to left instead of left to right. This should not be the case here.

https://developer.mozilla.org/en-US/docs/Web/CSS/direction

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
```diff
@@ -65,6 +66,16 @@ legend-panel-component {
 			.set-left-margin {
 				margin-left: 18px;
 			}
+
+			.marked-package {
+				overflow: hidden;
+				vertical-align: middle;
```
BridgeAR: Is this required?

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
NearW: There is no css property for text-overflow: ellipsis left. People on stackoverflow suggest the "direction-hack" as well. https://stackoverflow.com/a/9793669

Issue #610, opened 2019-06-28
Display a helpful message when schema of loaded cc.json does not match visualization schema


Richargh:
# Feature request

## Description

The CodeCharta [schema](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/util/schema.json) prevents loading files that do not confirm to it. Unfortunately the error message is rather cryptical, sometimes does not help at all and the text not copyable. 

With the advent of cc.json 2.0 we need to change this because cc.json will change in fundamental ways.

### Major version difference

* If CodeCharta detects that the major version of the cc.json is higher than the schema version that CC knows about, it displays an error notification that urges the user to update CC and does not load the map.

### Minor version difference

* If CodeCharta detects that the minor version of the cc.json is higher than the schema version that CC knows about, it displays an info notification that urges the user to update CC.
* If CodeCharta detects that the cc.json contains items that it does not know from the schema it also displays helpful information but continuous working as normal.
* If CodeCharta detects that the cc.json contains items that are in conflict with the schema it displays an error information with concrete details and does not load the map further.

I think we can display the warning as a helpful triangle next to loaded file in the dropdown. No idea about the error notification though.

A attached a badly behaving cc.json:
[https://app.zenhub.com/files/84911142/61779d35-7f10-4660-bf66-234e48a008c4/download](https://app.zenhub.com/files/84911142/61779d35-7f10-4660-bf66-234e48a008c4/download)

PR #555, opened 2019-05-30
Feature/523/download dialog

alschmut:
# Feature/523/download dialog

closes #523 

## Description

- **When** clicking the downloadFile-Button
- **then** a pop-up dialog appears and shows following content
    - the fileName including the current timestamp
    - a list of checkboxes to indicate which data is available to download (the user can also un/select available data-checkboxes)
    - The *Save*-button triggers the actual download of the *cc.json*-file including the selected data-objects

image

## Definition of Done

A task/pull request will not be considered to be complete until all these items can be checked off.

- [x] **All** requirements mentioned in the issue are implemented
- [x] Does match the Code of Conduct and the Contribution file
- [x] Task has its own **GitHub issue** (something it is solving)
  - [x] Issue number is **included in the commit messages** for traceability
- [x] **Update the README.md** with any changes/additions made
- [x] **Update the CHANGELOG.md** with any changes/additions made
- [ ] **Enough test coverage to ensure that uncovered changes do not break functionality**
- [x] **All tests pass**
- [x] **Descriptive pull request text**, answering:
  - What problem/issue are you fixing?
  - What does this PR implement and how?
- [ ] **Assign your PR to someone** for a code review
  - This person _will be contacted **first**_ if a bug is introduced into `master`
- [x] **Manual testing** did not fail

PR #860, opened 2020-02-07
Tech/rewrite file validation

alschmut:
# Tech/rewrite file validation

## Issue
As a developer I want to guarantee that the imported files contain the expected data right before pushing their file-content into the application state, in order to be able to use it appropriately afterwards.

## Current situation
CodeCharta uses a self written json Schema (find app/codeCharta/util/schema.json) to validate the incoming imported files. The issue with this is, the json schema is very difficult to read, hard to maintain, when the corresponding interface changes and therefore it's outdated and doesn't validate, what we actually want it to validate.

## Solution
Use [typescript-json-schema](https://github.com/YousefED/typescript-json-schema) to generate the json schema from the well defined interface `ExportCCFile` inside the `codeCharta.model.ts` file. This ensures that: 
- nobody needs to understand the internals of the json schema
- everybody is able to easily update the schema using the command `npm run schema:generate`
- The schema is up to date and correct IN DETAIL:
  - It reports errors with wrong enum parameters (like with `AttributeType`)
  - It knows which properties are required and which are optional from the interface `?` character.

Tech: The following command generates the file. It uses the following parameters:
- `--required` makes properties without the `?` required
- `--topRef` makes it easier to read the json schema with a top level reference, which references to the definition for the `ExportCCFile` schema, which then recursively calls other definitions.
```
typescript-json-schema ./app/codeCharta/codeCharta.model.ts ExportCCFile --required --topRef --out app/codeCharta/util/generatedSchema.json
```

Further necessary changes:
- update dataMocks and remove outdated properties like `fileName` and `id`
- update interface `ExportCCFile` to reflect the actual expected content with optional properties and the different BlacklistType `hide`

## How to test
You can use the online [JSON Schema validator](https://www.jsonschemavalidator.net), copy paste the generatedSchema and a cc.json of your choice. Then you can change the cc.json data and check if it reports the errors or not. Of course you can also import files into the running visualization of this branch, as it uses the new generatedSchema.

## TODO
- [x] Move HTML generation out of the dialogService (undo the changes to create a new dialog pop up)
- [x] Validate apiVersion and migrate afterwards (see [comment](https://github.com/MaibornWolff/codecharta/pull/860#issuecomment-586589507))
- [x] Fix resetting fileStates to early when importing invalid files, which results in a state without any files
- [x] Copy or generate the same validation file for analysis/tools/ValidationTool/src/main/resources/cc.json

## Definition of Done

- [ ] **All tests pass**


alschmut:
Can somebody check the failing e2e test2? They don't wanna pass on my machine, for some reason not even on master..

NearW:
Code looks good. We should try to get rid of the stringified html code inside the controller though somehow.

I'll look into the e2e tests.

I'd suggest that once we have this branch cleaned up and ready to merge, we split the `codeCharta.model.ts` file into `codeCharta.model.ts` and `codeCharta.export.model.ts`

NearW:
The amazonaws map is invalid with this scheme. It's a 0.1 version and has a fileName specified in the cc.json. 

I don't have any information about how the versions of the cc.json changed. We could write a converter, that just removes the attribute from the file before importing it, so that we still support < 1.0 maps

NearW:
I just added a migration that removes the fileName from the cc.json.

Seems like the fileName was set from downloading a file in an older version of codecharta though. It's not part of the amazonaws.cc.json we have in our analysis. However, I know get errors regarding missing attributes on nodes.
We really need to find out, what the difference between every version is and migrate them in the visualization. We could also do that with a new filter in the analysis, but as a user it would be easier for me to just open any file and it magically works.

alschmut:
@NearW thanks for contributing (: 

Regarding the apiVersion: If I remember right, we changed that from 0.1 to 1.0 when I implemented the first version of edges in June 2018. The additional `fileName` property is 100% a leftover from a downloaded file and actually not related to the apiVersion. That means, it could also have happened that people downloaded a file with fileName with apiVersion 1.0, as we removed the fileName property not before our released codecharta-visualization version 1.26.0 within #555 on July 13th 2019. 

The CodeMap interface had the `attributes` property required, although all parent nodes, usually do not have an `attributes` property, as they get summed up from its children in the visualization. Therefore I made the property optional.

alschmut:
@NearW are you sure to migrate before we can even guarantee we have some useful data? One test fails, because we can not guarantee the file has an apiVersion. I think there are three solutions:
1. either we just check if the `apiVersion`property is available and keep everything else you implemented, but for more complicated migrations, we can not guarantee we have the needed properties,
1. we validate with an interface that covers all apiVersions, mark the outdated properties optional and migrate those properties afterwards
1. or we do following:
    - check if the `apiVersion` property is available
    - validate with the appropriate interface (schema) corresponding to its apiVersion (that means we have multiple schemas for each possible exportCCFile interface)
    - migrate afterwards, as we now know its some valid file

alschmut:
In order to only export valid cc.jsons we could actually validate exported cc.jsons either on runtime or with a unit test.

README.md
```diff
@@ -11,7 +11,7 @@
...
 
 - [analysis](https://maibornwolff.github.io/codecharta/docs/analysis/): Command-Line-Tool for generating visualization data in the form of `.cc.json` files. It includes some pre-defined importers for e.g. SonarQube, SourceMonitor, SCM log information, generic csv data, as well as a command to validate and merge multiple data files.
-- [visualization](https://maibornwolff.github.io/codecharta/docs/visualization/): GUI for visualizing the code metrics inside the `.cc.json` files. The file schema is located [here](/visualization/app/codeCharta/util/schema.json) and we are using the [json-schema v4](https://tools.ietf.org/html/draft-zyp-json-schema-04) format.
+- [visualization](https://maibornwolff.github.io/codecharta/docs/visualization/): GUI for visualizing the code metrics inside the `.cc.json` files. Imported files are validated with the [generatedSchema.json](/visualization/app/codeCharta/util/generatedSchema.json), which is generated using the tool [typescript-json-schema](https://github.com/YousefED/typescript-json-schema).
```
BridgeAR:
```suggestion
- [visualization](https://maibornwolff.github.io/codecharta/docs/visualization/): GUI to visualize code metrics from `.cc.json` files. Imported files are validated using JSON Schema as defined in [generatedSchema.json](/visualization/app/codeCharta/util/generatedSchema.json).
```

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -29,7 +31,7 @@ describe("codeChartaService", () => {
 
 	describe("loadFiles", () => {
 		const expected: CCFile = {
-			fileMeta: { apiVersion: "1.1", fileName: "noFileName", projectName: "Sample Map" },
+			fileMeta: { apiVersion: "1.1", fileName: fileName, projectName: "Sample Map" },
```
BridgeAR:
```suggestion
			fileMeta: { apiVersion: "1.1", fileName, projectName: "Sample Map" },
```

It might be good to use a linting rule to automatically use the shorthand notation. It is nicer to read and to grasp the code that way.

visualization/app/codeCharta/ui/dialog/dialog.service.ts
```diff
@@ -45,4 +55,34 @@ export class DialogService {
...
+		)
+	}
+
+	private createDialogContent(errors: ErrorObject[]) {
+		let html = ""
+		errors.forEach((error: ErrorObject, index: number) => {
+			const key = _.keys(error.params)[0]
+			const object = " " + JSON.stringify(error.params[key])
+
+			html += "<p><b>Error #" + (index + 1) + "</b> " + key + " <i>" + object + "</i></p>"
```
BridgeAR: Is the error index useful for anyone? I suggest to remove that.

The error objects seem somewhat uncommon though. How and where are these errors generated?

visualization/app/codeCharta/ui/dialog/dialog.service.ts
```diff
@@ -45,4 +55,34 @@ export class DialogService {
...
+			html += '<p><i class="fa fa-comment-o" title="Message"></i> ' + _.lowerCase(error.message) + "</p>"
```
BridgeAR: How come the error message should be lower cased?

BridgeAR:
This needs a rebase with master.

visualization/app/codeCharta/ui/dialog/dialog.service.ts
alschmut: They are generated by the `ajv` json schema validator inside our [fileValidator.ts](https://github.com/MaibornWolff/codecharta/blob/master/visualization/app/codeCharta/util/fileValidator.ts#L30)

visualization/app/codeCharta/ui/dialog/dialog.service.ts
alschmut: Okey I admit, that's not a good idea.

visualization/app/codeCharta/ui/dialog/dialog.service.ts
alschmut: And yea we can remove the index. Officially we get an array of error objects, in reality I always just got one.

alschmut:
I was just undoing the changes, which created the new importErrorDialog with JS. Refactoring the dialog components will be done in a separate PR. That means now we just have to fix the failing e2e test. Can somebody have a look into there? :)

alschmut:
The same validation file exists in [analysis/tools/ValidationTool/src/main/resources/cc.json](https://github.com/MaibornWolff/codecharta/blob/master/analysis/tools/ValidationTool/src/main/resources/cc.json), which we should generate as well.

NearW:
This PR will be on hold until #884 is implemented.

NearW:
@alschmut btw is this related to #610 ?

alschmut:
@NearW oh yea it was, but not anymore since we reverted the html creation inside the service

alschmut:
With the above two commits I was just undoing what I proposed in [this comment](https://github.com/MaibornWolff/codecharta/pull/860#discussion_r392386132). Which means now this PR is mostly a tech refactoring:
- Now it allows every imported file, that has the necessary required properties (such as `nodes`, `apiVersion` or the node's `name` and `type`, etc), and doesn't check for additional or misspelled properties.
- This in turn means I was removing all migration that was done in order to accept some legacy properties (like the `filename`) that were exported in the past.

BridgeAR:
@alschmut this needs a rebase :)

NearW:
@alschmut We might have some conflicts here... 👼

NearW:
@alschmut needs another rebase. Will review this next week then. :)

alschmut:
We are having an issue here with the changes of #888. Meaning the `attributeType` type changed from `array` to `object`. So there exist exported files with both types, the old and the new ones. The released CC accepts both, because the validation file is not validating correctly. When fixing the validation with this PR the old `array` will be invalid though. See below the different json schema parts.

Released invalid schema:
```
"attributeTypes": {
	"type": "object",
	"minItems": 1,
	"items": {
		"type": "array",
		"minItems": 1,
		"items": {
			"$ref": "#/definitions/attributesType"
		}
	}
},
"attributesType": {
	"type": "object",
	"properties": {
		"sum": {
			"type": "number"
		},
	"average": {
		"type": "number"
		}
	}
}
```

New valid generated schema:
```
"attributeTypes": {
	"$ref": "#/definitions/AttributeTypes"
},
"AttributeTypes": {
	"properties": {
		"edges": {
			"additionalProperties": {
				"enum": ["absolute", "relative"],
				"type": "string"
			},
			"type": "object"
		},
		"nodes": {
			"additionalProperties": {
				"enum": ["absolute", "relative"],
				"type": "string"
			},
			"type": "object"
		}
	},
	"type": "object"
}
```

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -47,6 +48,8 @@ export class FileChooserController {
 
 		this.codeChartaService.loadFiles([nameDataPair]).catch((validationResult: CCValidationResult) => {
 			this.storeService.dispatch(setIsLoadingFile(false))
+			this.storeService.dispatch(setIsLoadingMap(false))
+			console.error(validationResult)
```
BridgeAR, line 52: This should probably not be kept?

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -15,119 +15,117 @@ export interface CCValidationResult {
...
-
-			if (!valid) {
-				result.error = validate.errors.map((error: Ajv.ErrorObject) => this.getValidationMessage(error))
-				result.title = this.ERROR_MESSAGES.validationError.title
-			} else if (!FileValidator.hasUniqueChildren(file.nodes[0])) {
-				result.error.push(this.ERROR_MESSAGES.nodesNotUnique.message)
-				result.title = this.ERROR_MESSAGES.nodesNotUnique.title
-			}
+	if (result.error.length === 0) {
+		const ajv = new Ajv({ allErrors: true })
```
BridgeAR, line 67: Ideally, the validation is separated from the schema generation.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -15,15 +16,12 @@ export class CodeChartaService {
...
-				const validationResult: CCValidationResult = FileValidator.validate(nameDataPair.content)
-				if (validationResult.error.length === 0) {
-					const ccFile = this.getCCFile(nameDataPair.fileName, nameDataPair.content)
-					NodeDecorator.preDecorateFile(ccFile)
-					this.storeService.dispatch(addFile(ccFile))
-					if (validationResult.warning.length !== 0) {
-						reject(validationResult)
-					}
-				} else {
+				const validationResult: CCValidationResult = validate(nameDataPair.content)
```
BridgeAR, line 19: Ideally, the `validate` function would just throw in case of an error. If that's the case, the promise would immediately be rejected with the error that the function threw.

A validation function should ideally also not return anything. It's just there to synchronously validate that everything works as expected and throws an error if not.

In this specific case it would also have the benefit of stopping the forEach loop. Right now it will just continue to run, while with throwing an error it would stop right away.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -15,15 +16,12 @@ export class CodeChartaService {
...
+				if (validationResult.error.length > 0) {
+					reject(validationResult)
+				}
+				this.addFile(nameDataPair.fileName, nameDataPair.content)
+				if (validationResult.warning.length > 0) {
```
BridgeAR, line 24: I suggest to "emit" warnings or to handle them as errors. Right now errors and warnings are handled identically in this code. Do we ever handle them differently? And if so, why does a warning result in the same behavior here?

BridgeAR:
@alschmut we discussed to release a new major soon. This might be a good candidate to include in that major version. @NearW @RomanenkoVladimir @JaiJoPa what do you think?

visualization/app/codeCharta/util/fileValidator.ts
NearW: What are the benefits of doing so?
in our case we could create a file `SchemaGenerator` which exports a function 

```ts
export function generateSchema() {
	const ajv = new Ajv({ allErrors: true })
	return ajv.validate(jsonSchema)
}
```

visualization/app/codeCharta/codeCharta.service.ts
NearW: I was confused by this as well. I think the intention here is to reject errors and display them. We want to resolve and add the file to our state on warnings but still display the warnings.

I committed something that should make this intention clear and also resolves your other discussion regarding returning a validationResult @BridgeAR

NearW:
I spent some time today and worked on this PR including all the discussions involved here.

Besides adding proper resolve and rejections to the promises, I also updated the schema slightly so that we can also read and validate older attributeTypes.

I think with these changes this PR can be merged, since we still support all previous cc.json versions.

Someone PTAL

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -75,10 +76,10 @@ export class CodeChartaController {
 			.then(() => {
 				this.storeService.dispatch(setState(ScenarioHelper.getDefaultScenarioSetting()))
 			})
-			.catch(e => {
+			.catch((errors: ErrorObject[]) => {
```
BridgeAR, line 79: It is not a good idea to reject arrays. Please always only reject / throw actual error.

Note: while working with promises it is always recommended to use async functions. Those do not only improve readability but also provide good stack traces.

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -2,16 +2,20 @@ import "./codeCharta.module"
...
 import { getCCFiles, isSingleState } from "./model/files/files.helper"
+import { DialogService } from "./ui/dialog/dialog.service"
 
 describe("codeChartaService", () => {
 	let codeChartaService: CodeChartaService
 	let storeService: StoreService
-	let validFileContent
+	let dialogService: DialogService
+	let validFileContent: ExportCCFile
+	const fileName: string = "someFileName"
```
BridgeAR, line 18: Note: it is not required to define the type in situations like these. The TypeScript compiler infers the type right away.

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -152,28 +156,28 @@ describe("codeChartaService", () => {
...
 			delete invalidFileContent.projectName
 			codeChartaService
-				.loadFiles([{ fileName: validFileContent.fileName, content: null }])
+				.loadFiles([{ fileName, content: invalidFileContent }])
 				.then(() => {
 					letTestFail()
 				})
 				.catch(err => {
-					expect(err.error).toEqual(["file is empty or invalid"])
+					expect(err.error).toEqual(["Required error:  should have required property 'projectName'"])
```
BridgeAR, line 168: Nice change!

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -183,5 +187,28 @@ describe("codeChartaService", () => {
...
+						content: validFileContent
+					}
+				])
+				.then(() => {
+					const attributeTypes: AttributeTypes = {
+						nodes: { rloc: AttributeTypeValue.absolute },
+						edges: { pairing: AttributeTypeValue.relative }
+					}
+					expect(getCCFiles(storeService.getState().files)[0].settings.fileSettings.attributeTypes).toEqual(attributeTypes)
+					done()
```
BridgeAR, line 210: Instead of using `done`, it should be possible to return a promise. Using an async function would be ideal.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,46 @@
...
 import { getCCFiles } from "./model/files/files.helper"
+import { DialogService } from "./ui/dialog/dialog.service"
 
 export class CodeChartaService {
 	public static ROOT_NAME = "root"
 	public static ROOT_PATH = "/" + CodeChartaService.ROOT_NAME
 	public static readonly CC_FILE_EXTENSION = ".cc.json"
 
-	constructor(private storeService: StoreService) {}
+	constructor(private storeService: StoreService, private dialogService: DialogService) {}
```
BridgeAR, line 16: While it's correct TypeScript, I recommend to use as much JavaScript notation instead of TypeScript specific notation. JS does support private fields (`#privateProperty = "privateValue"`). It would then be transpiled for older browsers.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +59,53 @@ export class CodeChartaService {
...
+			nodes: !attributeTypes.nodes ? {} : attributeTypes.nodes,
+			edges: !attributeTypes.edges ? {} : attributeTypes.edges
 		}
-		return newAttributeTypes
 	}
 
 	private potentiallyUpdateBlacklistTypes(blacklist): BlacklistItem[] {
 		blacklist.forEach(x => {
-			if (x.type === "hide") {
+			if (x.type === ExportBlacklistType.hide) {
```
BridgeAR, line 82: ❤️

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +59,53 @@ export class CodeChartaService {
...
 				x.type = BlacklistType.flatten
 			}
 		})
 		return blacklist
 	}
+
+	private migrateAttributeTypes(attributeTypes: OldAttributeTypes): AttributeTypes {
+		const result = { nodes: {}, edges: {} }
+		attributeTypes.nodes.forEach(x => {
+			result.nodes[Object.keys(x)[0]] = Object.values(x)[0]
```
BridgeAR, line 92:
```suggestion
			const [[key, value]] = Object.entries(x)
			result.nodes[key] = value
```

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +59,53 @@ export class CodeChartaService {
...
+		})
+		attributeTypes.edges.forEach(x => {
+			result.edges[Object.keys(x)[0]] = Object.values(x)[0]
```
BridgeAR, line 95:
```suggestion
			const [[key, value]] = Object.entries(x)
			result.nodes[key] = value
```

visualization/app/codeCharta/util/fileDownloader.ts
```diff
@@ -101,7 +91,7 @@ export class FileDownloader {
 	}
 
 	private static downloadData(data: ExportCCFile, fileName: string) {
-		let dataJson = stringify(data)
+		let dataJson = JSON.stringify(data)
```
BridgeAR, line 94: 👍

visualization/app/codeCharta/util/fileValidator.spec.ts
```diff
@@ -24,143 +24,181 @@ describe("FileValidator", () => {
...
-		expectFileToBeInvalid(errors)
-		expect(errors.title).toEqual("Error Loading File")
+	it("should throw on null", () => {
+		expectFileToThrowResult(invalidFile)
+		try {
+			validate(null)
+		} catch (e) {
+			expectFileToBeInvalid(e)
+			expect(e.title).toEqual("Error Loading File")
+		}
```
BridgeAR, line 50: Using `try / catch` in tests is almost always not the ideal way to write a test. Instead, rely upon e.g., `assert.throws()` or `expect(foo).toThrow(error)`

The reason is that we never validate that the function did indeed throw. If it would not throw, we'd miss that! We could in theory add another variable to test for that as well but that's not nice to read.

visualization/app/codeCharta/util/fileValidator.ts
BridgeAR: What I meant is that the schema is currently generated inside of a `validate` function (https://github.com/MaibornWolff/codecharta/pull/860/files#diff-0c4adf7ab75be4463a0b0279057f0fe4R44) and that it's best to keep validation function to validating things only :)

I would split this function into a separate step.

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -75,10 +76,10 @@ export class CodeChartaController {
...
 				this.storeService.dispatch(setIsLoadingFile(false))
-				console.error(e)
-				this.printErrors(e)
+				console.error(errors)
+				this.dialogService.showErrorDialog(JSON.stringify(errors))
```
BridgeAR, line 82: Stringifying errors is going to loose some information depending on how the object looks like. I do not know where this is logged though and what tools you otherwise use to inspect objects.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,46 @@
...
-					const ccFile = this.getCCFile(nameDataPair.fileName, nameDataPair.content)
-					NodeDecorator.preDecorateFile(ccFile)
-					this.storeService.dispatch(addFile(ccFile))
-					if (validationResult.warning.length !== 0) {
-						reject(validationResult)
+				try {
+					validate(nameDataPair.content)
+				} catch (e) {
+					if (e.warning.length > 0) {
+						this.printWarnings(e)
```
BridgeAR, line 25: Using `try / catch` here seems the wrong approach.

There are three ways I see how to handle this: either return `warnings` only from the validation function and `throw` inside the validation function in case of actual errors. That way the try / catch is not necessary.
An alternative would be to pass through the `printWarnings` function to the validate function, so that it would be triggered from inside the validate function.
The third one is to split the validation function up into pure `errors` and then check for warnings in a separate step afterwards.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,46 @@
...
 		return new Promise((resolve, reject) => {
 			nameDataPairs.forEach((nameDataPair: NameDataPair) => {
-				const validationResult: CCValidationResult = FileValidator.validate(nameDataPair.content)
-				if (validationResult.error.length === 0) {
-					const ccFile = this.getCCFile(nameDataPair.fileName, nameDataPair.content)
-					NodeDecorator.preDecorateFile(ccFile)
-					this.storeService.dispatch(addFile(ccFile))
-					if (validationResult.warning.length !== 0) {
-						reject(validationResult)
+				try {
```
BridgeAR, line 21: This is independent from the change: this function is purely sync code or promises get a new chain. Thus, wrapping the code in `new Promise` seems wrong.

visualization/app/codeCharta/codeCharta.service.spec.ts
NearW: That is very true. Enabling this rule will solve this issue once and for all. I'll adress this in another PR though.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-inferrable-types.md

visualization/app/codeCharta/codeCharta.service.ts
NearW: Webpack should compile our Typescript code to JS and then transpile it with babel.
Correct me if I'm wrong here

visualization/app/codeCharta/codeCharta.service.ts
NearW: This is beautiful.

visualization/app/codeCharta/util/fileValidator.spec.ts
NearW: We actually check if this test throws an error. I added two functions checking that:

```
function expectFileToThrowResult(file) {
	expect(() => {
		validate(file)
	}).toThrow()
}

function expectFileNotToThrowResult(file) {
	expect(() => {
		validate(file)
	}).not.toThrow()
}
```

They are called before doing something in the test to ensure, that we actually execute the part in the catch-block.

visualization/app/codeCharta/codeCharta.component.ts
NearW: mdDialog expects it's content as a string

NearW:
@BridgeAR Thanks for the in depth code review. Most of the requested changes are connected with how I handled Promises and error throwing here. I'll try to give it another try and come back to you, if I need further assistance next week.

visualization/app/codeCharta/codeCharta.service.ts
NearW: Your first option is implies to return a validation-result containing the warnings. In another comment you said, that returning something in a validate-function is unexpected, since it should just validate and throw errors.

I think we should get rid of the warnings in the validation and handle them separately after we know that our file is valid.

visualization/app/codeCharta/codeCharta.service.ts
NearW: The promise chain in this one here is a mess. I'm not even sure if we need promises at all. I'll revisit that for sure.

visualization/app/codeCharta/codeCharta.service.ts
BridgeAR: There's nothing wrong. TypeScript itself is for me just mostly a very powerful linter. It provides no actual benefit language wise besides that type declarations that I would really use (abstract classes are nice though).

It's just that more people know how to write JavaScript than TypeScript and TypeScript is always just a super set of JavaScript. Thus, if something already works with JavaScript itself, I would not use a different TypeScript notation (you also have more control over the transpiled code that way).

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: That is fine :-) what I meant is that it's best to use something else than `JSON.stringify()` to get the string representation of the value.

visualization/app/codeCharta/util/fileValidator.spec.ts
BridgeAR: After checking the Jest `expect` API, I now see why you implemented it this way.
Seems like their `toThrow` check is not that powerful and is only able to match the class instance or the error message, not error properties.

Are these tests run in a Node.js context? If so, I recommend to use `assert`. But I guess it's run in the browser?

What error message do these errors have by the way? We always just check for the title property?

NearW: Tests are run in a node.js context.

You're right. We should add tests for the messages as well. Done.

NearW:
@BridgeAR Promises are gone now. Would be great if you could take another quick look at [this commit](https://github.com/MaibornWolff/codecharta/pull/860/commits/c2ad44f79f885ca742137993aa30356f6d64622d). 
Branch looks clean to me now. Will test some older cc.jsons and update the schema in the analysis when you approve those changes.

NearW:
I tested some cc.jsons and it seems like the validation works as it should.
When I updated the schema in the analysis, the JasomeImporter-Tests were failing. The JasomeImporter was never really in use and really created some weird cc.jsons. I fixed that.

This PR is ready to merge now. Please give this another quick review (reviewing the latest 4 commits is enough)

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -31,21 +30,23 @@ export class FileChooserController {
...
 				reader.onloadstart = () => {
 					this.storeService.dispatch(setIsLoadingFile(true))
 				}
+
 				reader.onload = event => {
 					if (isCompressed) {
 						const zlib = require("zlib")
 
-						content = zlib.unzipSync(Buffer.from((<any>event.target).result))
+						content = zlib.unzipSync(Buffer.from(event.target.result))
```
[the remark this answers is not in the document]
NearW: No, I tried removing the buffer. Did not work :<

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -92,96 +114,98 @@ describe("codeChartaService", () => {
...
-				.loadFiles([{ fileName: validFileContent.fileName, content: null }])
-				.then(() => {
-					letTestFail()
-				})
-				.catch(err => {
-					expect(err.error).toEqual(["file is empty or invalid"])
-					expect(err.warning).toEqual([])
-					done()
-				})
+		it("should not load the default scenario after loading a valid file, that does not have the required metrics", () => {
```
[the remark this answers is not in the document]
NearW: We select the first metric in the list. List is sorted alphabetically

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,56 @@
...
+				if (!_.isEmpty(e.warning)) {
+					this.dialogService.showValidationWarningDialog(e)
+				}
+
+				console.error(e)
+				this.storeService.dispatch(setIsLoadingFile(false))
+			}
 		})
+
+		if (!_.isEmpty(this.storeService.getState().files)) {
```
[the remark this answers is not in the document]
NearW: Do we gain anything from that? I think that the lodash function is more readable.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,56 @@
...
-				} else {
-					reject(validationResult)
+	public loadFiles(nameDataPairs: NameDataPair[]) {
+		nameDataPairs.forEach((nameDataPair: NameDataPair) => {
+			try {
+				validate(nameDataPair.content)
+				this.addFile(nameDataPair.fileName, nameDataPair.content)
+			} catch (e) {
+				if (!_.isEmpty(e.error)) {
+					this.dialogService.showValidationErrorDialog(e)
```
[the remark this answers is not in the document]
NearW: `validate()` throws ONE validationResult at the end that contains an array of errors and warnings.
These errors are then mapped in the `showValidationErrorDialog`-function. So we display all the encountered errors. :)

visualization/app/codeCharta/codeCharta.service.ts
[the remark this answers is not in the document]
BridgeAR: Lodash's isEmpty function works on many different data types. Thus it's unclear that it's an array, if you use Lodash. It's of course also extra cpu work (while not being expensive here). I always recommend against using Lodash. It's historical that it has such a broad usage but JavaScript has lots of new functions and sugar that makes most parts of Lodash obsolete. 

Check https://www.npmjs.com/package/eslint-plugin-you-dont-need-lodash-underscore for further readings.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +69,38 @@ export class CodeChartaService {
...
+			if (x.type === ExportBlacklistType.hide) {
 				x.type = BlacklistType.flatten
 			}
 		})
 		return blacklist
 	}
+
+	private setDefaultScenario() {
+		const metricData = this.metricService.getMetricData()
+		const { areaMetric, heightMetric, colorMetric } = ScenarioHelper.getDefaultScenarioSetting().dynamicSettings
```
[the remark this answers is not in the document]
NearW: The default scenario is complexity and only includes these three metrics. The metrics are rloc, mcc and mcc (area, height, color).

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +69,38 @@ export class CodeChartaService {
...
+		const set = new Set([areaMetric, heightMetric, colorMetric])
+
+		metricData.forEach(x => set.delete(x.name))
```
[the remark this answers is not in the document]
NearW: We have to find out, if we can actually apply the default scenarios. When we don't have rloc or mcc available in the cc.json we can't.
The set adds these metric names in a set (resulting in a set containing ["rloc", "mcc"]. If we encounter that metric name in our metricData from the cc.json, we can remove it from the set.
If the set is empty, the scenario can be applied.

visualization/app/codeCharta/codeCharta.service.ts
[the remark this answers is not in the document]
NearW: note to myself: break the for loop once an error is encountered

visualization/app/codeCharta/codeCharta.service.ts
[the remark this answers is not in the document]
NearW: Your suggestion wasn't correct. Instead of checking if every metric in metricData exists in the metrics from the scenario, we have to actually check if the metrics from the scenario exists in the metricData. :)

visualization/app/codeCharta/ui/dialog/dialog.service.ts
```diff
@@ -31,4 +32,26 @@ export class DialogService {
...
+	public showValidationErrorDialog(validationResult: CCValidationResult) {
+		const errorSymbol = '<i class="fa fa-exclamation-circle"></i> '
+		const lineBreak = "<br>"
+
+		const errorMessage = validationResult.error.map(message => errorSymbol + message).join(lineBreak)
+
+		const htmlMessage = "<p>" + errorMessage + "</p>"
+
+		this.showErrorDialog(htmlMessage, validationResult.title)
+	}
```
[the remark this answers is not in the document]
NearW: Done anyway.

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -92,96 +114,98 @@ describe("codeChartaService", () => {
...
-					letTestFail()
-				})
-				.catch(() => {
-					done()
-				})
+		it("should show error on invalid file", () => {
+			codeChartaService.loadFiles([{ fileName, content: null }])
+
+			expect(storeService.getState().files).toHaveLength(0)
+			expect(dialogService.showValidationErrorDialog).toHaveBeenCalled()
```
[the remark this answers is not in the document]
NearW: Please take a look at [this commit](https://github.com/MaibornWolff/codecharta/pull/860/commits/6c0e80041629ff0694d7bda19cf1803089ce5d35)

I hope this is what you asked for. :)

NearW:
@BridgeAR Everything is adressed again. PTAL

visualization/app/codeCharta/ui/dialog/dialog.service.ts
```diff
@@ -31,4 +32,28 @@ export class DialogService {
...
+
+	private buildHtmlMessage(symbol: string, validationResult: string[]): string {
+		const lineBreak = "<br>"
+
+		const errorMessage = validationResult.map(message => symbol + message).join(lineBreak)
+
+		return "<p>" + errorMessage + "</p>"
+	}
```
BridgeAR, line 58: Note: this could be simplified a lot but I do not want to block this PR any longer, so it's nothing to address here. Most of this code is duplicated and it's possible to inline everything into a single function that is probably just five lines long instead of three lines.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -52,36 +76,38 @@ export class CodeChartaService {
...
 		})
 		return blacklist
 	}
+
+	private setDefaultScenario() {
+		const metricData = this.metricService.getMetricData()
+		const { areaMetric, heightMetric, colorMetric } = ScenarioHelper.getDefaultScenarioSetting().dynamicSettings
+		const names = [areaMetric, heightMetric, colorMetric]
+
+		const existsInMetricData = (metric: string) => metricData.map(x => x.name).includes(metric)
```
BridgeAR, line 107: The map increases the work significantly, since we do not only have to do the lookup but also have to go through all entries each time. It's therefore `O(n * m * 2)` and the map operation is more expensive then the includes due to allocating a new array and transforming the content. There are not a lot of entries, it won't hurt much due to that but it's a good practice not to code like that in general.

```suggestion
		const metricNames = new Set(metricData.map(x => x.name))
		const existsInMetricData = (metricName: string) => metricNames.has(metricName)
```

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,63 @@
...
 				}
-			})
-			this.storeService.dispatch(setSingle(getCCFiles(this.storeService.getState().files)[0]))
-			resolve()
+
+				if (!_.isEmpty(e.warning)) {
+					this.dialogService.showValidationWarningDialog(e)
+				}
+
+				console.error(e)
```
BridgeAR, line 43:
```suggestion
```

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -1,39 +1,63 @@
...
+				return
+			}
+
+			try {
+				validate(nameDataPair.content)
+				this.addFile(nameDataPair.fileName, nameDataPair.content)
+			} catch (e) {
+				if (!_.isEmpty(e.error)) {
+					this.dialogService.showValidationErrorDialog(e)
+					printErrors = false
```
BridgeAR, line 36: Should this only stop in case of errors, not warnings?

We miss to remove already added files in case the first file is valid but the second one is not. This might cause issues when trying to load files again after having validation errors.

NearW: > Should this only stop in case of errors, not warnings?

Yes. Warnings are totally fine. Warnings happen when a file is outdated. But since we're backward compatible, we can read the file.

> We miss to remove already added files in case the first file is valid but the second one is not. This might cause issues when trying to load files again after having validation errors.

That is the case in this PR, but is adressed in #1082 . We wait until all files are validated before setting them. Let's fix this issue in #1082 then

NearW:
@BridgeAR Done

PR #866, opened 2020-02-23
Tech/remove dependencies

alschmut:
# Tech/remove dependencies

## Description

- Remove dependency [roboto-fontface](https://github.com/choffmeister/roboto-fontface-bower), as it doesn't really make a big difference (see gif with the minimal difference to the default font)
- Remove dependency [angular-md-expansion-panel](https://github.com/B-3PO/angular-material-expansion-panel), as we are not using it anymore (since the old sidebar on the right hand side with the expandable panels is gone)

## Questions
- can we also remove the dependency [angular-messages](https://github.com/angular/bower-angular-messages)? I think it's not in use, but wanted to ask you again to be sure.
- Is it recommended to use the [typedoc](https://github.com/TypeStrong/typedoc) documentation by default for other potentially interested developers or could we remove it, as I actually never saw that documentation until now? It is created within the travis build and can be created manually with `npm run doc`. Is there any way to see these docs from our deployed version on github.io? It's also currently our 2nd largest dependency with 47MB.

## Screenshots
The bolder version is without the Roboto font.

## Definition of Done

- [ ] **All tests pass**


BridgeAR:
I just checked for `angular-messages` and could not find any usage either. It should therefore be fine to remove it.
I did not check typedoc. It's just a dev dependency, so it's not urgent to remove it.

What might as well be removed are the following modules:

* `path`: it is a direct copy of the Node.js path module. Is it not possible to use the core one instead?
* `rimraf`: Node.js core has this functionality built-in in versions >= 12.x (`fs.rmdir` with the recursive option).
* `es6-shim`: is it required to support old browsers? I would expect this not to be required.
* Maybe some babel related dependencies? We might at least re-evaluate some of the settings similar to the es6 shim.

alschmut:
- `rimraf` is only used to create the typedoc documentation files and currently our travis CI uses node v9.11.2
- for `es6-shim` regarding browser compatibility, @fatihi or @Richargh might have some plans, ideas, wishes etc. ;)

NearW:
For some odd reason we don't really distinguish between dev dependency and regular dependency. A lot of frameworks such as specific angular-components are listed as dev dependency.

We don't use typedoc right now and I'm not a fan of using it. In most cases it just adds redundant documentation to a function.

NearW:
> * `rimraf` is only used to create the typedoc documentation files and currently our travis CI uses node v9.11.2

Should we update our node CI version? :)

alschmut:
@NearW As a first step I am fine with removing typedoc and rimraf. This might in fact also enhance our build performance ✌️ Updating the node version might be more tricky. So I would suggest to do that with a separate branch.

PR #867, opened 2020-02-23
Tech/replace intersection result interface

alschmut:
# Tech/replace intersection result interface

## Description

If you share the same opinion as me, that the interface `IntersectionResult {intersectionFound: boolean, building?: CodeMapBuilding}` is not necessary, here is some small PR to remove it and just return the intersectedBuilding, if one was found while hovering.

## Definition of Done

- [x] **All tests pass**


alschmut:
Thanks for your review @BridgeAR. As this PR mentiones its just a tech change (refactoring) and doesn't impact any user interactions with the visualisation. I basically just changed the signature of a [function](https://github.com/MaibornWolff/codecharta/pull/867/files#diff-84a1e28e59f87ee486835fb66710ee5bR30-R59) which checks if the mouse cursor hoveres a CodeMapBuilding:
- Before, it returned this `IntersectionResult`, containing a boolean if a building is hovered and another object containing null or the hovered building
- Now, it returns a `CodeMapBuilding`, containing null, if there is no building hovered or the building itself, if one is hovered

In my opinion this is clearer, having only one return object (the CodeMapBuilding), than having the IntersectionResult with two properties, where one depends on the other.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -46,7 +45,6 @@ export class CodeMapMouseEventService
 
 	private mouse: Coordinates = { x: 0, y: 0 }
 	private oldMouse: Coordinates = { x: 0, y: 0 }
-	private intersectionResult: IntersectionResult
```
NearW: We need to track the intersectionResult for this line below `if (highlightedBuilding && this.intersectionResult.intersectionFound)`.  This change was required, because we can now have highlighted buildings without an intersection by hovering over the distribution-bar.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -112,22 +110,15 @@ export class CodeMapMouseEventService
...
-					to = this.intersectionResult.building
-				} else {
-					to = this.highlightedInTreeView
-				}
+				const to = intersectedBuilding ? intersectedBuilding : this.highlightedInTreeView
 
 				if (from !== to) {
 					this.unhoverBuilding()
-					if (to) {
+					if (to !== null) {
```
NearW: `intersectedBuilding` can be undefined. This new if-statement would always be true, because `undefined !== null` and `A_BUILDING !== null` is both true, isn't it?

alschmut: True, this was not a useful change. I undid it. But still, highlighting worked as expected also with the distribution bar.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
alschmut: That still works as in fact we use the same resource `highlighted` (buildings) inside the threeSceneService. When hovering the fileExtenseionBar it adds the buildings into that list and this list is used from the mouseEventService. So we don't have to duplicate and store the highlightedBuildings inside the mousEventService.

NearW: My bad. This was related to another issue that reoccurs now: When you select a building all buildings that are connected to it get highlighted. When you click on an empty space to deselect, there are still buildings hovered. That's why we have to check the intersectionResult later on again.

Reproduce:
- Load a map with edges
- Select an edge metric
- Select a building
- Try deselecting it by clicking into the void

alschmut: uhh that is weird. I'll have a look

alschmut: I was now adapting the functionality as before. I think this should be good enough for now until the hovered/selected buildings are somehow stored in our redux state.

NearW:
Tried to fix a bug that I assumed didn't exist on master. Found out it also exists on master so I created #890 .
This is approved.

PR #877, opened 2020-03-11
Bug/876/wrong max value calculation

### LINKED ISSUE #876: Wrong max value of metric calculation
# Bug

## Expected Behavior
I'd assume that the maxValue of a metric shown in parenthesis behind the metric name is calculated only by the files that are shown in that map. Therefore in my gif the value for avgCommits should be 0.

## Actual Behavior
The maxValue is calculated with ALL files loaded. It doesn't matter if they're currently selected or not. This results in a wrong max value that is used to determine what metric we will use as a default. This default can result in an area metric selected, that is 0 meaning it show's a blank map.

## Additional Infos
This happens with downloaded maps, because when you open two maps A and B, where A has metric "rloc: 10" and "mcc: 5" and B only has "rloc: 12" then the downloaded map B_DOWNLOADED has metrics "rloc: 12" and "mcc : 0". When the metric isnt available at all, it can't be selected. That's why this bug mostly occurs with downloaded maps.

## Screenshots

wrong_metric_selected
Shows that the wrong max value is shown. It should be 0 for the second map.

wrong_default_metric_selected
Shows that the wrong metric is selected by default.

  1. Open two downloaded maps
 2. Select a metric that is actually 0 but is shows a higher max value

NearW:
# Max value calculation should only use visible files 

closes #876 

## Description

- Max value of a metric is now calculated by only used the visible files
- Prevents wrong default metric selection for downloaded maps

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/util/codeMapHelper.ts
```diff
@@ -40,7 +40,9 @@ export class CodeMapHelper {
 		const ignoredNodePaths = ignore()
 			.add(CodeMapHelper.transformPath(gitignorePath))
 			.filter(nodes.map(n => CodeMapHelper.transformPath(n.path)))
-		return nodes.filter(n => !ignoredNodePaths.includes(CodeMapHelper.transformPath(n.path)))
+		//TODO: Review again once we use a isBlacklisted attribute in our CodeMapNodes
+		const set = new Set(ignoredNodePaths)
+		return nodes.filter(n => !set.has(CodeMapHelper.transformPath(n.path)))
```
NearW: This commit got here somehow, weird. It's a small performance improvement though. Just keep it. :b

alschmut:
This is truly a small change, but i wouldn't call it a bug. It has been implemented like this forever. The metric chooser always showed every metric of all imported files, and it also shows the maxMetric for all imported files next by. I think one can discuss about what might be the better feature, but I think the actual issue with the metricChooser is another. 

I would say the user (how has not been developing the metricChooser and knows who it behaves) wants to know which metrics with its maxValues are available in which map: Meaning: maybe we would need to finally show which metrics are available in the visible maps and which aren't and then It makes more sense to only calculate the maxValue for the visible maps

NearW:
The user can select all metrics that are 0, because they actually only exist in another map? That sounds like bad design to me. The maxValue and the selectable metric should only depend on visibleFileStates in my opinion. I will discuss this tomorrow in our weekly and post a statement here.

NearW:
@alschmut we decided that the maxValue and the selectable metrics should only depend on the visible file states. I will fix the selectable metrics in another PR.

Feel free to merge. :)

PR #880, opened 2020-03-12
Tech/refactor node context menu

alschmut:
# Tech/refactor node context menu

## Description

- Move `onBuildingRightClicked()` from codeMapComponent to nodeContextMenu as it only called methods inside the nodeContextMenu
- Rename viewModel variable `contextMenuBuilding` to `codeMapNode` in nodeContextMenu
- Rename function `currentFolderIsMarkedWithColor()`to `isNodeOrParentMarked()`
- Rename function `packageIsMarked()` to `isNodeMarked()`
- Tidy up nodeContextMenu tests to call mock servicesonly in one `beforeEach()` function at the top of the file
- Remove event `onHideNodeContextMenu` as it's only needed within the component
- Init mapColors `onMapColorsChanged` event instead of once from the constructor
- The `$timeout` function is only needed to synchronise angular bindings and so it's named now. Without the timeout of 50ms, calling the function to hide the contextMenu had to be moved after the action, as otherwise the selected codeMapNode is set to null too early

## Definition of Done

- [ ] **All tests pass**


NearW:
Your solution was buggy (hide, exclude and focus was not working). I refactored some more and did the following things:
- Only trigger an onBuildingRightClicked event, when there is actually a building right clicked. Before it would broadcast on every right click with null
- Change the event listener for node context menu hide from "click" to "mousedown" so that it also tracks the right clicks, that are no longer broadcasted

In order to fix the final problem (when you click on on exclude, the event listener first hides the menu and then tries to exclude a null reference), we have to determine if we are clicking inside the node context menu or not. This is done with some query selectors now.

alschmut:
@NearW thanks for testing and fixing! Isn't there any other solution than calculating the elements offset etc? I would still say, overall its a good refactoring.

NearW:
It is a good refactoring with a good scope imo.

I tried a lot before I ended up working with reading css attributes. The main problem is, that we need an event listener on <body> to get the same behaviour as before. We also want to have the contextMenu pop up when we press mouse down, not up. The only way to really make this work was to determine whether or not we are clicking inside the nodeContextMenu component or outside of it. If we click inside of it, we don't want the other event listener to react.

NearW:
@alschmut , I think that the mouse events and therefore the nodeContextMenu will need some further refactorings. In my opinion moving everything to redux that involves user mouse actions will improve the data flow and understanding a lot. We have several click events, nodeContextMenu, highlighting, selecting... which can all be handled with well defined actions.

If you can't find a better solution for this problem right now either, please give it another manual test and approve it if everything works. I think, this can be merged.

alschmut:
Yea I will Testa new merge tomorrow then 👍 awesome, I already got the next small step refactoring for the hovered and selected nodes in my mind 💡

NearW:
These small PRs really don't take me a lot of time to review. Keep it up! 👍

PR #882, opened 2020-03-12
Tech/no unused locals ts compiler

NearW:
# Enable noUnusedLocals rule for typescript compiler

## Description

- Upgrade typescript for disabling rule checking on whole files
- Added actionHelper and added a wrapper function that checks if an action is part of an enum (notice that this method only works when key and value of an enum entry are the same. this is the case for our actions)
- Updated plop template

## Definition of Done

- [ ] **All tests pass**


visualization/app/codeCharta/util/actionHelper.ts
```diff
@@ -0,0 +1,3 @@
+export function isActionOfType(actionType: string, actions) {
+	return Object.keys(actions).includes(actionType)
```
alschmut: Why don't you use `Object.values()` kinda like before with `_.values()`?

visualization/plop-templates/redux/service.ts.hbs
```diff
@@ -15,7 +15,7 @@ export class {{properCase name}}Service implements StoreSubscriber {
 	}
 
 	public onStoreChanged(actionType: string) {
-		if (_.values({{properCase name}}Actions).includes(actionType)) {
+		if (isActionOfType(actionType, {{properName}}Actions)) {
```
alschmut: Use `{{properCase name}}` :)

visualization/app/codeCharta/util/actionHelper.ts
NearW: I might look like a rookie, if I say that I didnt know this function existed. That's why I will just resolve this conversation and act like this never happened. Will fix tomorrow.

NearW:
We have an issue with typedoc here it seems.

Let's merge #881 first and see if this solves our problem.

alschmut:
Really? I thought I saw some failing typescript linting rule: the unused variables inside the injector.service were not correctly ignored I guessed.

visualization/plop-templates/redux/service.ts.hbs
alschmut: done

visualization/app/codeCharta/util/actionHelper.ts
alschmut: done

alschmut:
@NearW Just one last question. What do you think about moving the function `isActionOfType` from actionHelper.ts to the reduxHelper.ts?

visualization/app/codeCharta/util/reduxHelper.ts
```diff
@@ -16,6 +17,10 @@ export function addItemToArray(array: any[], item: any): any[] {
 	return array
 }
 
+export function isActionOfType(actionType: string, actions) {
+	return Object.values(actions).includes(actionType)
```
BridgeAR: This seems very costly. These actions are all string based enums. AFAIK they all have a value that is identical to their key. It should therefore suffice to do:

 ```suggestion
	return actions[actionType]
```

This is a whole different complexity level: from `O(n)` to `O(1)`. I also wonder why these enums have to be string based? It would be more efficient to use number based enums.

NearW: https://redux.js.org/basics/actions/
Redux actions are strings.

I will apply your suggestion though. As stated before: keys and values are the same. Hopefully this won't change.

PR #886, opened 2020-03-13
Bug/search panel opens when importing new file

NearW:
# Search panel opens for 300ms when importing a new file 

## Description

- Use css transitions instead of $timeout
- Set viewModel to minimized when mounted
- Show treeView even though the panel is minimized. This will give the search-panel it's height, making transitions on height possible
- blacklist SearchPanelModeActions so that they don't trigger a rendering. This improves performance on larger maps a lot

## Future
We should work on a new redux-store structure. Blacklisting actions is not a good practice. We should split our store in properties that actually change the map and trigger a new rendering and properties, that just handle some booleans. `isPresentationMode` or `isLoadingMap` is not changing the map. `invertColorRange` or `searchPattern` trigger a re-rendering though.

We currently act like every action triggers a re-rendering and blacklist those who don't. Let's keep it that way, but group the non-rendering actions in a sub-reducers. This allows us to write a simple statement `isActionOfType(action.type, NON_RENDERING_ACTIONS)`

## Definition of Done

- [x] **All tests pass**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Changed
 
+- Selectable metrics will only contain metrics from the visible maps
+
 ### Removed 🗑
 
 ### Fixed 🐞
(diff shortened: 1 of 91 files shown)


NearW:
@alschmut ready for review.

alschmut:
So I found two minor visual bugs. 
- When you open the BlacklistFlattenPanel and close it again, the treeView is shortly visible.
- When you open the treeViewPanel and close it again, the matchingFilesCounter icons disappear too early

NearW:
@alschmut Thanks for your input.

I worked on a way better solution. We now use a boolean `isExpanded` and never set the `searchPanelMode` to minimized. This way, we can always render the previous state and give the div a height, that is required for a proper transition. When minimized, we simply update the boolean.

Issue #884, opened 2020-03-13
Filter: Migrating older cc.jsons


NearW:
# Feature request

## Description

As a user, I want to migrate my older versions of the cc.json to a newer or the newest available version, so that it can be loaded in the visualization for sure.

## Acceptance criteria
- After migrating, the cc.json can be correctly validated with `ccsh check myFile.cc.json` using the generated scheme.json of #860 
- The CLI should look like this `ccsh migrate myOldFile.cc.json -v 1.1 -o myNewFile.cc.json`. Per default, the -v should be the newest version. If -o is not provided, print it in the console.
- If the cc.json that should be migrated is missing the version-field or encounters a problem, print a meaningful error message

## Open questions
Should we generate scheme.json for each version so that we can validate the cc.json before migrating? If so, we should specify the version in the check tool like `ccsh check myFile.cc.json -v 1.0`

BridgeAR:
As an alternative solution for this:

What about implementing a strategy to auto-detect the version while loading it into the visualizer? That way it would be transformed internally and the user could get a pop-up notification about the ongoing transformation including an option to download the cc.json with the new structure.

That way it's super user friendly while probably being a bit more work implementation wise.

Richargh:
Older Versions? Does there exist a v2 of the cc.json? Also does the mergin of old/new files still work?

alschmut:
😂 nice Richard :D no don’t worry, there is no 2.0.  but when I was generating a correct Schema to validate our cc.json for #860 we found out that exported cc.json files before July 13th 2019 included the fileName property, that actually never existed on the cc.json schema, but was exported as we decorated the map within the visualization.

BridgeAR:
This is likely less of an issue as soon as we switch to the config based download functionality. Every change to the cc.json structure should be upwards compatible and thus, there should not be an actual need for this. If we ever encounter such need, let's do this on a case by case basis.

PR #888, opened 2020-03-16
Feature/365/attribute types in ccsh

alexhunziker:
# Attribute Types in TokeiImporter and SCMLogParser and consistent data structure

relates to issue #365

## Description

This PR adresses the issue concerning the Attribute types. It is the first step to properly solve #365. Namely it includes:

- Fixed inconsistencies concerning the AttributeTypes in the Analysis. This includes update and refactoring of the model sub-project
- New validation scheme for `ccsh check` which are consistent with the implemented AttributeTypes
- Update of merge and modify filters to correctly handle attributeTypes
- AttributeTypes for tokeiImporter and SCMLogParser
- Update of AttribiteTypes in the visualization to handle new (consistent) data structure.

This issue does *not* include the median calculation in the visualization nor the sum/median symbol for metrics and edgeMetrics.

# Is this a Bugfix / Feature or Change?

The remaining question is if the altered file schema is considered a bugfix, a new feature or a change.

If it was a change, the API version would need to be incremented to 1.2, also implementing an migration filter as suggested in #884 might make sense in this case. If this PR is a bugfix or a new feature, we are fine though imo.

The argument in favour of a change is that the file schema has been, strictly speaking, updated. On the other hand, since the file schema was inconsistent before this PR and did not work, it can be argued that this part of the PR is a bugfix. Lastly it could be argued that AttributeTypes as a whole is a new feature, since we effectivly did not make use of it previously.

Personally i would probably opt for either bugfix or new feature due to simplicity, but the decision is up to you.

## Definition of Done

- [ ] **All tests pass**


visualization/app/codeCharta/state/metric.service.ts
```diff
@@ -63,14 +62,8 @@ export class MetricService implements FilesSelectionSubscriber, BlacklistSubscri
...
 	public getAttributeTypeByMetric(metricName: string, state: State): AttributeTypeValue {
-		const attributeType = this.getMergedAttributeTypes(state.fileSettings.attributeTypes).find(x => {
-			return _.findKey(x) === metricName
-		})
-
-		if (attributeType) {
-			return attributeType[metricName]
-		}
-		return null
+		const attributeType = state.fileSettings.attributeTypes.nodes[metricName]
```
alschmut: I think we should for now check both the nodes and edges attributeTypes as before.

alexhunziker: I would disagree and would instead suggest to add a similar function to the edgeMetric.service instead. That way the metric service does only node metric stuff and leaves all edge things to the dedicated edgeMetric service

alschmut: oh that's a very good idea. Didn't think about it close enough 👍

analysis/import/SCMLogParser/src/main/kotlin/de/maibornwolff/codecharta/importer/scmlogparser/input/metrics/NumberOfAuthors.kt
```diff
@@ -20,4 +21,8 @@ class NumberOfAuthors: Metric {
     override fun value(): Number {
         return authors.size
     }
+
+    override fun attributeType(): AttributeType {
+        return AttributeType.relative
```
NearW: Should we set this to relative? This value neither does sense if sum it nor if we average/median it. This value would require to check for unique authors and sum them then.

example:
Folder A, contains File B and C.
A was commited by Alex and Ben at some point.
B was commited by Alex and Alexander at some point.

I would expect that when I hover over the folder, it would display that 3 authors worked on that package.

visualization/app/codeCharta/state/metric.service.ts
NearW: but am I missing something or is this functionality not implemented in the edgeMetricDataService?

visualization/app/codeCharta/state/metric.service.ts
```diff
@@ -63,14 +53,8 @@ export class MetricService implements FilesSelectionSubscriber, BlacklistSubscri
...
+		return !!attributeType ? attributeType : null
```
NearW: why do we return null here instead of undefined?

```
return attributeType
```

NearW:
I'll give this a big manual test. This will include
- MergeFilter
- EdgeFilter
- StructureModifier
- CodeMaatImporter
- SCMLogParser
- SourceCodeParser
- TokeiiImporter

Once this is done and my comments are resolved, this can be merged.

visualization/app/codeCharta/state/metric.service.ts
alexhunziker: You are right, this functionality is not implemented in this PR (because its functionality is not used currently). However the AttributeTypes II PR includes this method, since there attributeTypes for EdgeMetrics are used.

analysis/import/SCMLogParser/src/main/kotlin/de/maibornwolff/codecharta/importer/scmlogparser/input/metrics/NumberOfAuthors.kt
alexhunziker: I agree with you that neither aggregation mode is perfect. However, the behaviour you described (which I think makes sense) is beyond the scope of this PR/issue, and would need to be implemented as a new aggregation mode in the visualization.

So for now the question is: which aggregation mode is less wrong? Summing them up (in the current fashion) makes no sense obviously, while the median gives at least an idea on how many people worked on an average file within a package.

Therefore I suggest: Leave this relative for this PR & open a new issue for the better aggregation mode.

alschmut: Yea I agree with @alexhunziker ;)

visualization/app/codeCharta/state/metric.service.ts
alexhunziker: So when I was refactoring the method, i included this line because there are tests that fail if the return value is undefined instead of null.

However, i checked it just now, and there seems to be no good reason why undefined should be a less desirable return value than null (so the test case was specified too strict probably). I will therefore simplify the function and adjust the tests.

NearW: the only way to return null, if we actually read a cc.json that has `key: "rloc"`, `value: null`
and check for `state.fileSettings.attributeTypes.nodes["rloc"]

If I'm not mistaken, we can use this test instead:
```
it("should return undefined if attributeType not available", () => {
			const actual = metricService.getAttributeTypeByMetric("notfound", state)

			expect(actual).toBeUndefined()
		})
```

alexhunziker: Hmm this may get philosophical... so for our app we need that "actual" evaluates to false in an if statement. Therefore for any purposes in our app null or undefined are fine, which is why i have decided to test for !actual==true. (because the previous test, that required the result to be null made me think during refactoring that null is required when in fact it wasn't).

So from my point of view the question is: is there a value from defining the test as narrow as possible, even if the app does not require it?

PR #925, opened 2020-04-21
Feature/610/helpful message if scheme doesnt match ccjson

### LINKED ISSUE #610: Display a helpful message when schema of loaded cc.json does not match visualization schema

RomanenkoVladimir:
# Helpful Error Messages in Case of Validation Failure or API Version Discrepancy

closes #610

## Description
Changes Validation Scheme to jsconscheme, (ajv was used previously)
Checks API Version of CC against API Version used by Verification File
Provides Error Messages according to FeatureRequest 610
Added ngSanitize to sanitize inline HTML
Maybe changes to dialog.service are required, for a nicer and better scalable solution

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -4,18 +4,19 @@ import _ from "lodash"
...
 import { addFile, setSingle } from "./state/store/files/files.actions"
+import { DialogService } from "./ui/dialog/dialog.service"
 
 export class CodeChartaService {
 	public static ROOT_NAME = "root"
 	public static ROOT_PATH = "/" + CodeChartaService.ROOT_NAME
 	public static readonly CC_FILE_EXTENSION = ".cc.json"
 
-	constructor(private storeService: StoreService) {}
+	constructor(private storeService: StoreService, private dialogService?: DialogService) {}
```
alschmut, line 14: You can remove the `?` behind the dialogService

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -59,8 +58,12 @@ export class FileChooserController {
 		}
 	}
 
-	private printErrors(errors: Object) {
-		this.dialogService.showErrorDialog(JSON.stringify(errors, null, "\t"))
+	private printErrors(errors: string[]) {
+		let errorMessage = ""
+		for (let i = 0; i < errors.length; i++) {
+			errorMessage += errors[i] + "</br>"
+		}
```
alschmut, line 65: `let errorMessage = errors.join("<br>")` should replace the for look

visualization/package.json
```diff
@@ -85,13 +85,14 @@
 		}
 	},
 	"dependencies": {
-		"angularjs-nouislider": "^1.0.1",
 		"ajv": "^6.10.0",
```
alschmut, line 88: When `ajv` is not used anymore we can remove the dependency

visualization/app/codeCharta/util/schema.json
```diff
@@ -20,7 +20,8 @@
 		},
 		"apiVersion": {
 			"title": "version of schema",
-			"type": "string"
+			"type": "string",
+			"pattern": "[0-9][.][0-9]"
```
alschmut, line 24: I am not a regex expert, but isn't backslashing a dot (like `"[0-9]\.[0-9]"`) the more often used option of matching a dot?

alschmut: We didn't change our apiVersion often, but with this pattern a `1.10` would be be considered invalid

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -1,15 +1,62 @@
...
 
 export class FileValidator {
+	public static validate(file: { nodes: CodeMapNode[] }, dialogService?: DialogService): string[] {
+		let minorApiWrongMessage = ""
+
+		if (!file) {
+			return ['<i class="fa fa-exclamation-circle"></i>' + " file is empty or invalid"]
+		}
+
+		let Validator = require("jsonschema").Validator
```
alschmut, line 13: how about importing the `Validator` using new syntac. Does that work?
`import { Validator } from "jsonschema"`

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -22,23 +69,45 @@ export class FileValidator {
...
+		for (let key in schemaCodeCharta) {
+			if (key === "codecharta") {
+				schemaApiVersion = schemaCodeCharta[key]
+				for (let apiVersionKey in schemaApiVersion) {
+					return schemaApiVersion[apiVersionKey]
 				}
-			]
+			}
 		}
+	}
```
alschmut, line 104: This method is somehow super complex for the outcome. Isn't the returned value for an apiVersion just the major version like below?
- `1.2` -> `"1"`
- `3.4` -> `"3"`

Instead of checking if the apiVersion is inside the package.json, I would rather use a simple variable to get the major version ...
```
const latestMajorApiVersion = require("../../../package.json").codecharta.apiVersion[0]
```
... and add a unit test which checks that the property is available inside the package.json and matches a valid apiVersion.

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -22,23 +69,45 @@ export class FileValidator {
...
+		if (majorVersion > apiVersionMajor) {
+			apiVersionMajorMinor[0] = true
+		} else if (minorVersion > apiVersionMinor) {
+			apiVersionMajorMinor[1] = true
 		}
-		let ajv = require("ajv")()
-		let compare = ajv.compile(require("./schema.json"))
-		let valid = compare(file)
+		return apiVersionMajorMinor
+	}
```
alschmut, line 90: This function is quite hard to read. How about splitting this (and the functions below) into multiple smaller separated pieces like:
```
	private isValidApiVersion(file: ExportCCFile): boolean {
		const hasApiVersion = file.apiVersion !== undefined
		const versionRegExp = new RegExp("[0-9]\.[0-9]")
		const isValidVersion = versionRegExp.test(file.apiVersion)
		return hasApiVersion && isValidVersion
	}

	private static fileHasHigherMajorVersion(file: ExportCCFile): boolean {
		let fileMajorVersion = this.getMajorVersion(file.apiVersion)
		let latestMajorVersion = this.getMajorVersion(latestApiVersion)

		return fileMajorVersion > latestMajorVersion
	}

	private static fileHasHigherMinorVersion(file: ExportCCFile): boolean {
		let fileMinorVersion = this.getMinorVersion(file.apiVersion)
		let latestMinorVersion = this.getMinorVersion(latestApiVersion)

		return fileMinorVersion > latestMinorVersion
	}

	private static getMajorVersion(version: string) {
		return version.split(".")[0]
	}

	private static getMinorVersion(version: string) {
		return version.split(".")[1]
	}
```

- The function `isValidApiVersion()` didn't really exist, but I think it makes sense, when you wanna access it to first check if its even valid. If not you should straight show an error message.
- The variable `latestApiVersion` is imported
  ```
  const latestApiVersion = require("../../../package.json").codecharta.apiVersion
  ```

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -1,15 +1,62 @@
 import { CodeMapNode } from "../codeCharta.model"
+import * as codeCharta from "../../../package.json"
+import { DialogService } from "../ui/dialog/dialog.service"
 
 export class FileValidator {
+	public static validate(file: { nodes: CodeMapNode[] }, dialogService?: DialogService): string[] {
```
alschmut, line 6: Can you please refactor this method? Its really long. Also is there a significant difference of the two different exclamation icons? I would just use one and if it is always added before one array item, we could just add it together with the "<br>" inside the `fileChooser.component`

visualization/app/codeCharta/util/schema.json
RomanenkoVladimir: This was done another way now, I forgot to remove those changes, thank you

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: i found no way to do it like that: but i can shorten it to 
`let validationResult = require("jsonschema").validate(file, `require("./schema.json"))`

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: The problem is that in case of the users Minor API version is wrong we display a warning, but continue rendering, the current implementation of codeCharta.service does not support warnings

RomanenkoVladimir: we just stop rendering if any output other then 0 is returned
But maybe I should change that

visualization/app/codeCharta/util/fileValidator.ts
alschmut: Okey sad, then let’s keep it, but can you put the required statement to the imports?

visualization/app/codeCharta/util/schema.json
RomanenkoVladimir: ([0-9]+)\\.([0-9]+) should be the required solution

visualization/app/codeCharta/util/fileValidator.ts
alschmut: I am not getting the problem, sorry 😁 how about returning an object like the following and then the fileChooser component can call the correct dialogs? 
```
{
  warnings: [],
  errors: []
}
```

visualization/app/codeCharta/util/schema.json
alschmut: I think you can remove the round brackets and one backslash 😉

RomanenkoVladimir: double slash is needed or parsing will fail, wrong espace sequence

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: sounds good, will need to address codeCharta and its test accordingly, 
I will do it, thank you

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -22,23 +75,43 @@ export class FileValidator {
...
-		if (!FileValidator.hasUniqueChildren(file.nodes[0])) {
-			return [
-				{
-					message: "names or ids are not unique",
-					dataPath: "uniqueness"
-				}
-			]
-		}
+	private static isValidApiVersion(file: { nodes: CodeMapNode[] }): boolean {
+		const apiVersion = this.getFileApiVersion(file)
```
alschmut, line 87: Is that function necessary? You could just use `const apiVersion = file.apiVersion`

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -22,23 +75,43 @@ export class FileValidator {
...
+		const hasApiVersion = apiVersion !== undefined
+		const versionRegExp = new RegExp("[0-9]+.[0-9]+")
```
alschmut, line 89: A dot `.` can be any character, while `\.` is only a dot.
```suggestion
		const versionRegExp = new RegExp("[0-9]+\.[0-9]+")
```

alschmut:
@RomanenkoVladimir I couldn't help myself and describing the refactoring would have been a lot of words. I just committed some FileValidator refactoring. have a look, if that suits you. I tried to make it more readable, although I don't know how to make the `validate()` smaller or split it into multiple functions (as we need the many return statements..). I gonna have another look later on.

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: It seems the \ gets removed by the prehook

RomanenkoVladimir:
The refactoring seems good, much clearer now
I changed validate into a switch statement, that should make it clearer as well

RomanenkoVladimir:
Also I removed the Error Dialog in case of a failure to parse cc.json, as from my understanding after reading the specification, a failure to parse is always an illegal file. 
So we will not need to throw 2 error dialogs, one of which will get overwritten anyways, confusing a user.
But I would like your opinion here, maybe i missed something?

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: We need to use \\ to escape the escape character as the whole thing will be converted to string later, should be fixed now

visualization/app/codeCharta/util/schema.json
RomanenkoVladimir: We need to use \\ to escape the escape character as the whole thing will be converted to string later, should be fixed now

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -1,15 +1,80 @@
...
+
+export interface CCValidationResult {
+	error: string[]
+	warning: string[]
+	title: string
+}
 
 export class FileValidator {
-	private static hasUniqueChildren(node: CodeMapNode) {
+	private static FILE_IS_INVALID = ["file is empty or invalid", "Error Loading File"]
```
alschmut, line 19: I think having nice objects with readable properties is nicer: Something like this:
```
private static FILE_IS_INVALID = { message: "file is empty or invalid", title: "Error Loading File" }
```

Or what might be also nice, is one object like below:
```
private static ERROR_MESSAGES = {
  fileIsInvalid: {
    title: "some title",
    message: "some message"
  },
  ...
}
```

alschmut: And I think titles that are similar to each other, are easier to recognise as titles. Like always starting with `Error: abc`/`Warning: abc` or always ending with `ABC Error` / `ABC Warning`

RomanenkoVladimir: I really like the second idea, makes it much more readable
Should we export it as a constant then?

alschmut: cool :) nah, if we don't move the object into a separate file, we don't need to export it

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -16,35 +16,51 @@ export interface CCValidationResult {
...
+		fileIsInvalid: {
+			title: "Error Loading File",
+			message: "file is empty or invalid"
+		},
+		apiVersionIsInvalid: {
+			title: "Error File API Version",
+			message: "file API Version is empty or invalid"
+		},
+		apiVersionIsOutdated: {
+			title: "Error CodeCharta Major API Version",
```
alschmut, line 29: Do we wanna keep calling the api version `File API Version`? Also below.

alschmut: or in this case `File Major API Version`

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -16,35 +16,51 @@ export interface CCValidationResult {
...
+			message: "API Version Outdated: Update CodeCharta API Version to match cc.json"
+		},
+		minorApiVersionOutdated: {
+			title: "Warning CodeCharta Minor API Version",
+			message: "Minor API Version Outdated"
+		},
+		nodesNotUnique: {
+			title: "Error Name-Type Uniques",
```
alschmut, line 37: I think this is clearer for the user
```suggestion
			title: "Error Node Uniques",
```

visualization/app/codeCharta/util/fileValidator.ts
RomanenkoVladimir: you think we should call it api Version?

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -26,11 +26,11 @@ export class FileValidator {
 			message: "file API Version is empty or invalid"
 		},
 		apiVersionIsOutdated: {
```
alschmut, line 28: how about renaming the property to `majorApiVersionIsOutdated`?

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -1,15 +1,96 @@
...
 
 export class FileValidator {
-	private static hasUniqueChildren(node: CodeMapNode) {
+	private static ERROR_MESSAGES = {
+		fileIsInvalid: {
+			title: "Error Loading File",
+			message: "file is empty or invalid"
+		},
+		apiVersionIsInvalid: {
+			title: "Error File API Version",
```
alschmut, line 25: Lets also remove the word `File` here, so we call the `API Version` always the same like below (:

RomanenkoVladimir: But the other API Version Errors are related to the API Version of CodeCharta, e.g. when using the App 
Might be confusing?

RomanenkoVladimir: I guess we could remove it from the title and leave it in the text?

alschmut: Ah shit. Now I get it. I got confused as well. You are right

alschmut: mhm no I don't get it. Aren't we only talking about one API Version? which is the `1.1` at the moment?

RomanenkoVladimir: Minor/ Major Api Version wrong refer to CodeCharta
Richard asked for this feature if the CodeCharta Version is behind the version for the file we throw an error/a warning
apiVersionIsInvalid refers to an invalid file api, e.g. 1.a

alschmut: Well its the CodeCharta project. Everything refers to CodeCharta I guess.. Its always the "CodeCharta `cc.json` API Version" no matter if its valid, outdated or invalid

RomanenkoVladimir: ok, the first two refer to the app version
It is specifically asked for in the issue, due to a possible 2.0 cc.json with fundamental changes

alschmut: The app version is the latestApiVersion is guess (which we store in the package.json)?

RomanenkoVladimir: yes, but there is an app version of codecharta as well, right?

alschmut: eh no.. we have a codecharta version which is like `1.45.5` the one Ben changes when he does the releases every few weeks, and there is the cc.json API version, which is `1.1` since almost forever

RomanenkoVladimir: The issue asked to check whether the api Version in the  cc.json is lower than the api Version of the loaded file
If sich a case is impossible, maybe we should change the request accordingly?

alschmut: You sentence is misunderstanding, but I think I get what you mean.

That can happen when:
- somebody installs codecharta locally with `nom install codecharta-visualization` with CC Version 1.45.5 and a latest cc.json API Version of 1.1
- one year goes by
- we (the CC Team) updated the API Version to 2.0 Kollege
- the user imports a file (from a colleague) with API Version 1.1 cc.json into the old CodeCharta application
- now the loaded file has a newer version than the CC Application wants

RomanenkoVladimir: Ok, then I fail to see a reason to change the message

This message is thrown if the api Version of the loaded fail is not valid 

The other two are thrown in one of the case described above and are related to the cc.json only

RomanenkoVladimir: I can change it, but I believe it would result in more confusion

NearW: It's just about the usage of the word "file" twice here. We use it when the file context is actually invalid in the first part, but when we speak about the API Version, we also use the term `File API Version` even though the API Version is obviously part of the file.
The suggestion is to remove the word "file" from the API Version checks. I agree with that.

RomanenkoVladimir: Ok, I think I understand what you mean now

visualization/app/app.ts
```diff
@@ -4,9 +4,10 @@ import "angular-aria"
 import "angular-material"
 import "./codeCharta/codeCharta.module"
 import "./assets/icon.ico"
+import "angular-sanitize"
 import "./app.scss"
 
-angular.module("app", ["app.codeCharta", "ngMaterial"])
+angular.module("app", ["app.codeCharta", "ngMaterial", "ngSanitize"])
```
alschmut, line 10: does the `ngSanitize` do anything? I think it allows to use the `$sanitize` Service. If this isn't used yet, we should remove the dependency `angular-sanitize`

RomanenkoVladimir: It is needed to use htmlContext for md-dialog
Removing it will result in an error

NearW:
@alschmut @RomanenkoVladimir Is this mergable yet?

alschmut:
@NearW yes now it is mergable 🥳

PR #943, opened 2020-04-28
WIP: Tech/blacklist class

NearW:
# Add a blacklist class that provides all functionality to work with the blacklist

## Description

- Added class Blacklist that replaces `BlacklistItem[]`
- Removed unused parameter blacklist from `NodeDecorator.decorateParentNodesWithAggregatedAttributes` 

## TODO
- [ ] somehow setting the default state is not working correctly. Tests fail. Application works fine.

## Definition of Done

- [ ] **All tests pass**

### DIFF
--- a/visualization/app/codeCharta/codeCharta.model.ts
+++ b/visualization/app/codeCharta/codeCharta.model.ts
@@ -2,6 +2,7 @@ import { Vector3 } from "three"
 import { Action } from "redux"
 import { Files } from "./model/files"
 import { CodeMapBuilding } from "./ui/codeMap/rendering/codeMapBuilding"
+import { Blacklist } from "./model/blacklist"
 
 export interface NameDataPair {
 	fileName: string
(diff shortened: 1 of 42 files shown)


NearW:
So I did some research and found out why this PR might not be a good idea and also how this problem occurs on our master branch as well.

The major problem here is, that we store instances of classes in our redux store. Those instances have properties and functions. That means, that they're not serializable to a plain javascript object. Redux has a problem with that. We should only put plain javascript objects in our store.
I did this with the files in the past not knowing about this.
All methods on those classes that are mutating the state, can be defined as actions. However, methods that are simply calculating or retrieving a value from that state, can't be defined as actions.

I think we should return to the approach from before and create `helper-files` for each of these class-objects in our state and move that to the folder where the reducer is in as well.
In this specific case, we'd create a `BlacklistHelper.ts` which provides methods such as `isPathExcluded(blacklist: BlacklistItem[], path: string)`.

PR #949, opened 2020-05-01
Feature/948/move unfocus button to node context menu

### LINKED ISSUE #948: Move unfocus button to the node context menu
# Feature request

## Description

As a user, I want to easily find the unfocus button without it being placed on top of the 3d-scene.
A user would need to right-click a building and select focus. The user would assume that in order to revert this behaviour he can right-click the node again and instead of focusing, the user gets the option to unfocus the building again.

NearW:
# Moved unfocus button to node context menu

closes #948 

## Description

- Moved the unfocus button from on top of the 3d-scene to the node-context-menu
- Moved hideNodeContextMenu after each dispatch call to the subscription-methods

### Before:
focus_old

### After
focus_new

## Definition of Done

- [x] **All tests pass**


alschmut:
While I like some of the refactoring changes on this PR, as a user I think it got more complicated to unfocused the map, as I have to right click the exact node I focused. I guess you had a conversation about that yesterday, so I will leave the decision up to you, but IMO its now less user friendly.

Did you think about showing the unfocused option also on children nodes? Like showing focus and unfocus options for children nodes and only unfocused for the actually focused node?

NearW:
@BridgeAR What do you think?

BridgeAR:
I personally like the way this is handled but I agree that it might not be most intuitive for users. Some people might not find a way back anymore. We could in theory mark the focused elements a bit differently to indicate things changed.
I personally would like others to shim in as well though.

BridgeAR:
Ping @NearW

NearW:
Marking the focused building different is a difficult task, if we plan to not just change the color. Changing the color would be a bad idea as well, since we can mark packages, select a building and a building has a color by itself anyway. This just adds complexity for the user.

I present you another approach, which displays the unfocus-button as well as the focus-button for children of the focused building. Therefore any building right-click will have an option to unfocus the whole thing. The only problem here is, that we don't have a FA-Icon which displays a crossed bullseye. Right now we use the same icon for focusing and unfocusing.

unfocus-new

PR #984, opened 2020-06-06
Deactivate sonar on travis

BridgeAR:
# Deactivate sonar on travis and update Node.js version used to test to v10.x

## Definition of Done

- [x] **All tests pass**


.travis.yml
```diff
@@ -136,7 +136,7 @@ jobs:
...
 stages:
   - test
-  - name: sonar
-    if: branch = master AND type != pull_request
-  - name: deploy
-    if: tag IS present
+  # - name: sonar
+  #   if: branch = master AND type != pull_request
+  # - name: deploy
+  #   if: tag IS present
```
alschmut, line 142: @BridgeAR Mentioning a stage inside the stages only ensures the stages will run in order from top to bottom and each stage only runs if the conditions evaluate to true. So the `sonar` and `deploy` stages still run when commented out.

What you actually want is something like this, which only deactivates the sonar branch.
```
stages:
  - test
  - name: sonar
    if: false AND branch = master AND type != pull_request
  - name: deploy
    if: tag IS present
```

NearW:
I updated the node travis version to 12 in another commit, which is already pushed to master.
We needed at least version 10 for the new puppeteer update, but version 12 is working as well.

BridgeAR:
@NearW it would be great to test multiple Node.js versions (at least the LTS versions). And stopping support for the older Node.js version should be considered a breaking change. This is probably not a big concern right now due to the limited user scope but we might want to discuss our semver handling again.

NearW:
> @NearW it would be great to test multiple Node.js versions (at least the LTS versions). And stopping support for the older Node.js version should be considered a breaking change. This is probably not a big concern right now due to the limited user scope but we might want to discuss our semver handling again.

That is a very good point. We should extend our CI by at least checking our lowest and highest suported node version. We could use the highest in general and when we do a release, we could run the pipelines with both versions. This would only slow down our deploy-pipeline.
However I'm not sure if this is technically possible. We should discuss this tomorrow.

PR #992, opened 2020-06-14
ESLint: enable no var requires rule

NearW:
# Enable no var requires rule

## Description

- Updated all uses of require to an import statement
- Added file `clone.ts` and added a wrapper function to clone, since it would require a double method call every time we use it. This should be reverted once rfdc releases a new version.


visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -33,7 +35,7 @@ export class CodeChartaController {
 		// @ts-ignore
 		private injectorService: InjectorService // We have to inject it somewhere
 	) {
-		this._viewModel.version = codeCharta.version
+		this._viewModel.version = getVersion()
```
BridgeAR, line 38: I would not change this. The helper does not reduce the code and before it was immediately clear what version was read. Now it's an indirection and the helper function is just extra code that has to be maintained. Was there a specific reason why you changed this?

visualization/app/codeCharta/e2e/url.e2e.ts
```diff
@@ -52,13 +55,13 @@ describe("codecharta", () => {
 				request.respond({
 					contentType: "application/json",
 					headers: { "Access-Control-Allow-Origin": "*" },
-					body: JSON.stringify(require("../assets/sample2.cc.json"))
+					body: JSON.stringify(sample1)
```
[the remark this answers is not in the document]
NearW: That is true.
Just for my clarification: When importing a .json file, it's getting parsed.
When opening it with fs it's still stringified.

So in that case here we'd so something like `body: fs.readFileSync("../assets/sample2.cc.json2)`

visualization/app/codeCharta/codeCharta.component.ts
NearW: I felt like importing the package.json was a bit ugly and we do that a lot. Basically whenever we needed either the current version of CodeCharta or the current API version.
`codeCharta.version` reads well, however codeCharta is actually the package.json. 

I don't know what a good practice is here. I just though that importing the package.json in like 10 files was ugly.

visualization/app/codeCharta/util/clone.ts
```diff
@@ -0,0 +1,5 @@
+import rfdc from "rfdc"
+
+export function clone(content: any): any {
+	return rfdc()(content)
+}
```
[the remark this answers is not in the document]
NearW: Thanks for opening that issue.

Do I have to create a subfolder default in the project and just have to export the clone function there? Never really worked on a npm module before 0:)

visualization/app/puppeteer.helper.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from "path"
 import { Browser, Page } from "puppeteer"
+import * as puppeteer from "puppeteer"
```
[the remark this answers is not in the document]
NearW: I tried to find a solution here and tried to use `import puppeteer from "puppeteer"` which should be the same as `const puppeteer = require("puppeteer)". I even checked with some console.logs and they look identical.

However puppeteer seems to be undefined. Am I missing something here?

visualization/app/codeCharta/util/clone.ts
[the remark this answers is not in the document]
BridgeAR: See https://nodejs.org/api/esm.html#esm_subpath_exports :-) I hope that helps?

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: I am not aware of any best practice when it comes to this. It is not that common to read the current modules version. And the cases that did this just accessed the package json directly.

I suggest to use `packageJSON` as variable name instead of `codeCharta`. That way it's even clearer what the version corresponds to.

visualization/app/codeCharta/e2e/url.e2e.ts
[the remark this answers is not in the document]
BridgeAR: `fs` reads the data into a buffer. That is turned into a string if you pass through the `encoding` option and set it to `utf8` (`fs.readFileSync("../assets/sample2.cc.json2", "utf8")`). You might want to use the promise based function instead though. Using the sync versions is almost always not ideal.

visualization/app/codeCharta/util/clone.ts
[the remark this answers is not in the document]
NearW: Helped a lot. I can't open a PR for that though since I'm not authorized to push a branch. :(

visualization/app/codeCharta/e2e/url.e2e.ts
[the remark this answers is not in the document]
NearW:
```
let sample1: string
fs.readFile("../assets/sample1.cc.json", "utf8", (err, data) => {
	sample1 = data
})
```

This seems a bit long.

Are you fine with wrapping that in a `jsonReader.ts` which exports this function:
```
export function readJSON(path: string): string {
	let tmp: string

	fs.readFile(path, "utf8", (err, data) => {
		tmp = data
	})
		
	return tmp
}
```

This would give us way shorter calls such as
```
const sample1 = readJSON("../assets/sample1.cc.json")
```

visualization/app/codeCharta/util/clone.ts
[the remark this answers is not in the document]
BridgeAR: You first have to fork the repository and then you push to your fork. That way you are able to open a pull request from your fork to the actual repository.

image

visualization/app/codeCharta/e2e/url.e2e.ts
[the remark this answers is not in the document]
BridgeAR: Your example code would not work. It would always return `undefined`. It's an async call but the function is executed synchronously. Using the promise based version is much nicer:

```js
page.on("request", async (request) => {
	if (request.url().includes("/fileOne.json")) {
		const body = await fs.promises.readFile("../assets/sample2.cc.json2", "utf8")
		request.respond({
			contentType: "application/json",
			headers: { "Access-Control-Allow-Origin": "*" },
			body
			...
```

visualization/app/puppeteer.helper.ts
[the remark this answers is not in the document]
BridgeAR: You have to check the TypeScript options you use. It's transpiled and I guess you have to add compatibility to Node.js esm.

visualization/app/codeCharta/e2e/url.e2e.ts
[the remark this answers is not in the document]
NearW: Ich habe versucht es umzusetzen. Es sind wohl einige TS-Entwickler unzufrieden, weil es keinen ordentlichen import dafür gibt. 
Jedenfalls funktioniert `import { promises as fs } from "fs"` für mich nicht.

https://github.com/nodejs/node/issues/21014#issuecomment-529012253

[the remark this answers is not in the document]
BridgeAR: @NearW it's possible to import it with `import fsPromise from "fs/promises"`

See https://nodejs.org/api/fs.html#fs_fs_promises_api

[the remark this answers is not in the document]
NearW: https://github.com/nodejs/node/pull/31553

Das geht wohl leider erst ab Node 14

visualization/app/codeCharta/util/clone.ts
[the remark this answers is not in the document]
NearW: @BridgeAR PR was created. Thanks a lot. :)

visualization/app/codeCharta/e2e/url.e2e.ts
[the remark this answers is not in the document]
BridgeAR: Ha, I see. In that case it's possible to use:

```js
import fs from "fs";

fs.promises.readFile()
```

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -8,10 +8,13 @@ import { InjectorService } from "./state/injector.service"
 import { StoreService } from "./state/store.service"
 import { setAppSettings } from "./state/store/appSettings/appSettings.actions"
 import { setIsLoadingFile } from "./state/store/appSettings/isLoadingFile/isLoadingFile.actions"
-import * as codeCharta from "../../package.json"
+import packageJson from "../../package.json"
 import { setDelta, setMultiple, setSingle } from "./state/store/files/files.actions"
 import { getCCFiles } from "./model/files/files.helper"
+import sample1 from "./assets/sample1.cc.json"
+import sample2 from "./assets/sample2.cc.json"
```
BridgeAR, line 15: Note: what I personally do in situations like these is to safe the object in a `.ts` file and mark the object as e.g., `ExportCCFile`. That way you have the type safety while still being able to import it easily.

NearW: We'd have two files with the same content then though. We can't just remoe the `.cc.json` files. :(

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -39,9 +39,7 @@ export class FileChooserController {
 
 				reader.onload = event => {
 					if (isCompressed) {
-						const zlib = require("zlib")
-
-						content = zlib.unzipSync(Buffer.from(event.target.result))
+						content = require("zlib").unzipSync(Buffer.from(event.target.result))
```
NearW, line 42: I don't know, why this has to be an inlined-require block.
If I import zlib with es6-imports, the e2e tests are failing with an error, that require is not defined. The tests somehow passed in travis though.

visualization/app/codeCharta/codeCharta.component.ts
NearW: Yes, but right now we don't have two identical json files.

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
BridgeAR: You found an actual bug. I guess it's somewhat my fault that I did not see that properly before. This is a Node.js module but the frontend has no `zlib` module. I was originally mistaken where this code would run.

Thus, we have to use some kind of module that is able to handle the data.

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: Why would the TS Files be identical in that case?

NearW: Depending on the cc.json it would be quite long. The only difference would be something like:
```ts
const json : CCFile = {
...MY_REALLY_LONG_JSON...
}
```

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
NearW: Ehm, ok. I still don't understand. If the frontend has no zlib module, why can we require it though?
What is our context actually? Are we in some DOM-context or is angular actually node-application?

BridgeAR: I have no idea why it did not fail earlier. Do we actually reach this code in our test?

NearW: OH. That's why it's working. No, we never reach it.
But we'd reach an import on top. That makes sense

NearW: But still: Why is this working in production?

BridgeAR: I am not certain what you mean with it working in production?

I'd say you found another bug: we can't choose zipped files anymore :) Those should be available.

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: I'm confused... I am not certain why you would want to keep both?

NearW: I think we might use the `.json` files for testing purposes and for the url-parameters. But I have to take a deeper look.

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
NearW: Our angular tests run in the `jsdom` environment. [Puppeteer chooses it's environment ](https://github.com/smooth-code/jest-puppeteer#usage). Afaik the puppeteer environment extends the NodeEnvironment. I thought zlib would be available then.

Our [webpack config](https://github.com/MaibornWolff/codecharta/blob/main/visualization/conf/webpack.config.js) compils our code to two different targets. We can create a node environment for the standalone (this allows us to use the cool upcoming features for standalone only, e.g. opening files, when double clicking a building) and we can create a [browser-like environment](https://webpack.js.org/configuration/target/) using `web`.
I think you can open a zipped `cc.json` with both targets.

BridgeAR: No, the compressed detection checks for the file ending. `json` is not compressed.

We are indeed able to do more in the standalone version but that should not limit the web version. So maybe let's just fix that? :-)

BridgeAR: We load browserify at some point and browserify has a zlib implementation that it ships. This causes issues when loading files are not compressed gzipped:

```
Uncaught Error: incorrect header check
    at Zlib._handle.onerror (index.js:352)
    at Zlib._error (binding.js:283)
    at Zlib._checkError (binding.js:254)
    at Zlib._write (binding.js:116)
    at Zlib.writeSync (binding.js:77)
    at Unzip.Zlib._processChunk (index.js:511)
    at zlibBufferSync (index.js:236)
    at Object.exports.unzipSync (index.js:149)
    at FileReader.reader.onload (fileChooser.component.ts:44)
```

BridgeAR: I think we could support a range of compressions.

NearW: Unzipping `.gz`-files is working in the web and standalone-version of codecharta.
It seems like webpack is providing some sort of polyfill for zlib. Running the e2e tests after building with `npm run build:web` works. `npm run build` is not working, because the application is build for a NodeEnvironment, which can't be served on the server provided from puppeteer.

I suggest we merge this branch.

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: Right, I see. Thanks!

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.html
```diff
@@ -7,7 +7,7 @@
 	id="input-file-id"
 	multiple
 	type="file"
-	accept=".json"
+	accept=".json,.gz"
```
BridgeAR, line 10: ❤️ 

This should probably be an entry in the changelog under `fixed`.

BridgeAR (approved):

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.html
NearW: Thanks for the reminder!

PR #1111, opened 2020-07-24
Bug/1095/number of edges not visible

### LINKED ISSUE #1095: Number of incoming and outgoing edges not visible
# Bug

## Expected Behavior

When I hover over a building, then I expect the number of incoming and outgoing edges to be visible next to the edge metric, devided by a slash.

When I select a building, then I expect the number of incoming and outgoing edges to be visible next to the edge metric in the side-bar, devided by a slash.

## Actual Behavior

Bot are not visible.

## Screenshots

image

1. Hover over or select a building

## Specifications
- CodeCharta Version: 1.50.0
- OS: Windows
- Browser: Chrome

NearW:
# Fix number of edges not visible when hovering

closes #1095 

## Description

The bug existed, because we already started rendering when the metricData was set and didn't await the edgeMetricData. I decided that instead of adding a simple if-condition to ensure that, we should rework some stuff and finally add the metrics to the redux store. This was a huge task. Some changes are trivial, but especially the nodeMetricService/Reducer and edgeMetricService/Reducer are not.

- Removed metricService and edgeMetricDataService and created a nodeMetricData and edgeMetricData property in the redux store instead
- Moved all logic that involves calculating and setting the state to it's reducer (you can now set the state by dispatching an action)
- Moved all logic that involves retrieving information from the state to the corresponding service
- Created two new types `NodeMetricData` and `EdgeMetricData` and used the previous `MetricData` for the parent-reducer
- Created a new event `metric-data-complete` that is triggered, when both node and edge metric data is calculated
- Fixed bug by listening to that event and rendering when all the metric data is available
- Removed all occurrences of `jest.resetAllMocks()` since they're not doing anything anyway
- **Changed edge-metrics being sorted by name now as we do with the node-metrics instead of by number of incoming and outgoing edges** 

In order to review this, check out the two new services I created and check every file that just required small changes (such as fixing the import and injecting the new service). I can help you with the rest then. :)

### ~ToDo~
- [x] Add regression test

## Definition of Done

- [x] **All tests pass**


visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+	const toPath = filePaths.find(x => x === edge.toNodeName)
+	return fromPath && toPath && isNotBlacklisted(fromPath, blacklist) && isNotBlacklisted(toPath, blacklist)
+}
+
+function isNotBlacklisted(path: string, blacklist: BlacklistItem[]): boolean {
+	return !CodeMapHelper.isPathBlacklisted(path, blacklist, BlacklistType.exclude)
+}
+
+function addEdgeToCalculationMap(edge: Edge) {
+	for (const edgeMetric of Object.keys(edge.attributes)) {
```
JaiJoPa, line 58: Is Object.keys(edge.attributes) containing the metrics of that edge like pairingRate and avgCommits ?

NearW: Yes, So do nodes. 

```ts
export interface CodeMapNode {
	name: string
	id?: number
	type: NodeType
	children?: CodeMapNode[]
	attributes?: KeyValuePair
	edgeAttributes?: {
		[key: string]: EdgeMetricCount
	}
	link?: string
	path?: string
	isExcluded?: boolean
	isFlattened?: boolean
	deltas?: {
		[key: string]: number
	}
}
```

attributes on a node or edge look like this:

```ts
const attributes : KeyValuePair = {
	rloc: 100,
	coverage: 50,
	mcc: 5
}
```

JaiJoPa: Okay perfect !

BridgeAR:
Needs a rebase

CHANGELOG.md
```diff
@@ -51,7 +52,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Changed
 
 - Reduced transition time when opening or collapsing parts of the ribbon bar #1043
-- Search Panel will open now when clicking in the search field and collapse when clicking somewhere else #1071
+- Search Panel will open now when clicking in the search field and collapse when clicking somewhere else #
```
NearW, line 55: Memo: Add issue number again

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+	const pathsPerFileState = getVisibleFileStates(fileStates).map(fileState => CodeMapHelper.getAllPaths(fileState.file.map))
+	const allFilePaths: string[] = [].concat(...pathsPerFileState)
+	getVisibleFileStates(fileStates).forEach(fileState => {
+		fileState.file.settings.fileSettings.edges.forEach(edge => {
+			if (bothNodesAssociatedAreVisible(edge, allFilePaths, blacklist)) {
+				addEdgeToCalculationMap(edge)
+			}
+		})
+	})
+	return nodeEdgeMetricsMap
```
BridgeAR, line 44: Returning `nodeEdgeMetricsMap` here seems rather confusing since the variables scope is the file.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+function createEntryIfNecessary(edgeMetricEntry: Map<string, EdgeMetricCount>, nodeName: string) {
+	if (!edgeMetricEntry.has(nodeName)) {
+		edgeMetricEntry.set(nodeName, { incoming: 0, outgoing: 0 })
+	}
+}
+
+function getMetricDataFromMap(hashMap: Map<string, Map<string, EdgeMetricCount>>): EdgeMetricData[] {
+	const metricData: EdgeMetricData[] = []
+
+	hashMap.forEach((occurences: any, edgeMetric: any) => {
```
BridgeAR, line 87:
```suggestion
	return hashMap.map((occurences: EdgeMetricCount[], edgeMetric: unknown) => {
```

Ideally, never use `any`. Instead, use `unknown` or better, if it's e.g., an object `Record<string, unknown>`.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+		occurences.forEach((value: EdgeMetricCount) => {
+			const combinedValue = value.incoming + value.outgoing
+			if (combinedValue > maximumMetricValue) {
+				maximumMetricValue = combinedValue
+			}
+		})
+		metricData.push({ name: edgeMetric, maxValue: maximumMetricValue })
+	})
+
+	return metricData
```
BridgeAR, line 98:
```suggestion
		return { name: edgeMetric, maxValue: maximumMetricValue }
	})
```

(In combination with the comment above)

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+	const sortedEdgeMetricMap = new Map()
+	if (nodeEdgeMetricsMap) {
+		nodeEdgeMetricsMap.forEach((value, key) => {
+			const sortedMapForMetric = new Map(
+				[...value.entries()].sort((a, b) => b[1].incoming + b[1].outgoing - (a[1].incoming + a[1].outgoing))
+			)
+			sortedEdgeMetricMap.set(key, sortedMapForMetric)
+		})
+	}
+	nodeEdgeMetricsMap = sortedEdgeMetricMap
```
BridgeAR, line 111: Without looking deeper into this: I expect it's possible to sort the metrics up front while creating `nodeEdgeMetricsMap`.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function addNoneMetric(metricData: EdgeMetricData[]) {
+	if (!metricData.some(x => x.name === EdgeMetricDataService.NONE_METRIC)) {
+		metricData.push({ name: EdgeMetricDataService.NONE_METRIC, maxValue: 0 })
```
BridgeAR, line 116: This could be done in `getMetricDataFromMap`. That way there's no need to go over all entries again.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -0,0 +1,111 @@
...
+
+	public onBlacklistChanged(blacklist: BlacklistItem[]) {
+		this.storeService.dispatch(calculateNewEdgeMetricData(this.storeService.getState().files, blacklist))
+	}
+
+	public onFilesSelectionChanged(files: FileState[]) {
+		this.storeService.dispatch(calculateNewEdgeMetricData(files, this.storeService.getState().fileSettings.blacklist))
+	}
+
+	public onAttributeTypesChanged(attributeTypes: AttributeTypes) {
```
BridgeAR, line 49: The `attributeTypes` are never used?

```suggestion
	public onAttributeTypesChanged() {
```

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -0,0 +1,111 @@
...
+
+	public onStoreChanged(actionType: string) {
+		if (isActionOfType(actionType, EdgeMetricDataActions)) {
+			this.notify(this.select())
+		}
+	}
+
+	public onBlacklistChanged(blacklist: BlacklistItem[]) {
+		this.storeService.dispatch(calculateNewEdgeMetricData(this.storeService.getState().files, blacklist))
+	}
```
BridgeAR, line 43: These store dispatches happen very frequently. I personally would expect that it's possible to abstract these somehow to have a simple function that does the dispatching for different kinds of metrics. But I did not check if that is really useful here.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.actions.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface CalculateNewEdgeMetricDataAction extends Action {
+	type: EdgeMetricDataActions.CALCULATE_NEW_EDGE_METRIC_DATA
+	payload: { fileStates: FileState[]; blacklist: BlacklistItem[] }
+}
+
+export type EdgeMetricDataAction = SetEdgeMetricDataAction | CalculateNewEdgeMetricDataAction
+
+export function setEdgeMetricData(edgeMetricData: EdgeMetricData[] = defaultEdgeMetricData): SetEdgeMetricDataAction {
```
BridgeAR, line 22: The name `set...` confuses me. This function is only going to return a new object but it is not going to set anything?

Since the function returns the passed in edgeMetricData 1-to-1, we might also just remove this function and just create the object where it's required

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+import { getVisibleFileStates } from "../../../../model/files/files.helper"
+import { CodeMapHelper } from "../../../../util/codeMapHelper"
+import { FileState } from "../../../../model/files/files"
+import { EdgeMetricDataService } from "./edgeMetricData.service"
+
+const clone = require("rfdc")()
+
+// Required for performance improvements
+export type NodeEdgeMetricsMap = Map<string, Map<string, EdgeMetricCount>>
+export let nodeEdgeMetricsMap: NodeEdgeMetricsMap
```
BridgeAR, line 12:
```suggestion
export let nodeEdgeMetricsMap: EdgeMetricCount = new Map()
```

I would always initialize the map directly.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		return this.storeService.getState().metricData.edgeMetricData.map(x => x.name)
+	}
+
+	public getAmountOfAffectedBuildings(metricName: string): number {
+		if (!nodeEdgeMetricsMap || !nodeEdgeMetricsMap.has(metricName)) {
+			return 0
+		}
+		return nodeEdgeMetricsMap.get(metricName).size
```
BridgeAR, line 67:
```suggestion
		const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
		return nodeEdgeMetrics === undefined ? 0 : nodeEdgeMetrics.size
```

The map should always be initialized when this is called. I would also not use `has` and `get` together. Instead, always use `get` and check for `undefined`. That way there's one less lookup necessary.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -0,0 +1,111 @@
...
+			return []
+		}
+
+		const highestEdgeCountBuildings: string[] = []
+		const edgeMetricMapKeyIterator = nodeEdgeMetricsMap.get(metricName).keys()
+		for (let i = 0; i < numberOfNodes; i++) {
+			highestEdgeCountBuildings.push(edgeMetricMapKeyIterator.next().value)
+		}
+
+		return highestEdgeCountBuildings
```
BridgeAR, line 81:
```suggestion
		const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
		return nodeEdgeMetrics === undefined ? [] : [...nodeEdgeMetrics.keys()]
```

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	public getMetricValuesForNode(node: HierarchyNode<CodeMapNode>): Map<string, EdgeMetricCount> {
+		const metricNames = this.getMetricNames().filter(it => !!nodeEdgeMetricsMap.get(it))
+		const nodeEdgeMetrics = new Map()
+
+		metricNames.forEach(metric => {
+			nodeEdgeMetrics.set(metric, nodeEdgeMetricsMap.get(metric).get(node.data.path))
+		})
```
BridgeAR, line 90:
```suggestion
		const nodeEdgeMetrics = new Map()
		this.getMetricNames().forEach(metric => {
			const edgeMetrics = nodeEdgeMetricsMap.get(metric)
			if (edgeMetrics === undefined)
				nodeEdgeMetrics.set(metric, edgeMetrics.get(node.data.path))
		})
```

Doing a `filter` and a `forEach` after each other is not good performance wise. Instead, just do the latter and do the filtering in the forEach.
```

File: visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.splitter.ts [hunk 15]
```diff
@@ -0,0 +1,6 @@
+import { EdgeMetricDataAction, setEdgeMetricData } from "./edgeMetricData.actions"
+import { EdgeMetricData } from "../../../../codeCharta.model"
+
+export function splitEdgeMetricDataAction(payload: EdgeMetricData[]): EdgeMetricDataAction {
+	return setEdgeMetricData(payload)
+}
```
BridgeAR, line 6: Is there any reason not to call `setEdgeMetricData()` directly? If you definitely want to use an alias, it's possible to do: `export const splitEdgeMetricDataAction = setEdgeMetricData`.

visualization/app/codeCharta/state/store/metricData/metricData.actions.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export interface SetMetricDataAction extends CCAction {
+	type: MetricDataActions.SET_METRIC_DATA
+	payload: RecursivePartial<MetricData>
+}
+
+export type MetricDataAction = SetMetricDataAction
+
+export function setMetricData(metricData: RecursivePartial<MetricData> = defaultMetricData): MetricDataAction {
```
BridgeAR, line 18: This is not really a setter. The name seems rather confusing to me.

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+	let storeService: StoreService
+	let $rootScope: IRootScopeService
+
+	beforeEach(() => {
+		restartSystem()
+		rebuildService()
+		withMockedEventMethods($rootScope)
+	})
+
+	function restartSystem() {
```
BridgeAR, line 24:
```suggestion
	function resetState() {
```

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+		instantiateModule("app.codeCharta.state")
+
+		$rootScope = getService<IRootScopeService>("$rootScope")
+		storeService = getService<StoreService>("storeService")
+	}
+
+	function rebuildService() {
```
BridgeAR, line 31:
```suggestion
	function resetMetricDataService() {
```

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.splitter.ts
```diff
@@ -0,0 +1,6 @@
+import { NodeMetricDataAction, setNodeMetricData } from "./nodeMetricData.actions"
+import { NodeMetricData } from "../../../../codeCharta.model"
+
+export function splitNodeMetricDataAction(payload: NodeMetricData[]): NodeMetricDataAction {
+	return setNodeMetricData(payload)
+}
```
BridgeAR, line 6: The same as above.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
```diff
@@ -0,0 +1,76 @@
...
+	public getAttributeTypeByMetric(metricName: string): AttributeTypeValue {
+		return this.storeService.getState().fileSettings.attributeTypes.nodes[metricName]
+	}
+
+	private select() {
+		return this.storeService.getState().metricData.nodeMetricData
+	}
+
+	private notify(newState: NodeMetricData[]) {
+		this.$rootScope.$broadcast(NodeMetricDataService.NODE_METRIC_DATA_CHANGED_EVENT, { nodeMetricData: newState })
```
BridgeAR, line 68: `notify()` is only used once. Since it's a "one line function", I would rather inline it. As a rule of thumb: one line functions are often more confusing than helpful and they obfuscate the overall code.

visualization/app/codeCharta/state/store/metricData/metricData.splitter.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function splitMetricDataActions(payload: RecursivePartial<MetricData>): CCAction[] {
+	const actions: CCAction[] = []
+
+	// Plop: Append action split here
+	if (payload.edgeMetricData !== undefined) {
+		actions.push(splitEdgeMetricDataAction(payload.edgeMetricData as EdgeMetricData[]))
+	}
+
+	if (payload.nodeMetricData !== undefined) {
```
BridgeAR, line 15: Could the payload contain both, `edgeMetricData` and `nodeMetricData` at the same time?

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.actions.ts
```diff
@@ -0,0 +1,36 @@
...
+export type NodeMetricDataAction = SetNodeMetricDataAction | CalculateNewNodeMetricDataAction
+
+export function setNodeMetricData(nodeMetricData: NodeMetricData[] = defaultNodeMetricData): SetNodeMetricDataAction {
+	return {
+		type: NodeMetricDataActions.SET_NODE_METRIC_DATA,
+		payload: nodeMetricData
+	}
+}
+
+export function calculateNewNodeMetricData(fileStates: FileState[], blacklist: BlacklistItem[]): CalculateNewNodeMetricDataAction {
```
BridgeAR, line 29: This function does not calculate anything. Thus, the name seems rather confusing.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.reducer.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function addUnaryMetric(metricData: NodeMetricData[]) {
+	if (!metricData.some(x => x.name === NodeMetricDataService.UNARY_METRIC)) {
+		metricData.push({
+			name: NodeMetricDataService.UNARY_METRIC,
+			maxValue: 1
+		})
+	}
+}
```
BridgeAR, line 86: This file is almost identical in it's functionality to the other metricData reducer and all comments apply accordingly. Ideally, they would share their logic. I am also still not convinced that there's need to handle them separately but I did not look deep enough into that.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+	function restartSystem() {
+		instantiateModule("app.codeCharta.state")
+
+		$rootScope = getService<IRootScopeService>("$rootScope")
+		storeService = getService<StoreService>("storeService")
+
+		storeService.dispatch(setNodeMetricData(metricData))
+	}
+
+	function rebuildService() {
```
BridgeAR, line 40: Same as in the other test file.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	public static subscribe($rootScope: IRootScopeService, subscriber: NodeMetricDataSubscriber) {
+		$rootScope.$on(NodeMetricDataService.NODE_METRIC_DATA_CHANGED_EVENT, (event, data) => {
+			subscriber.onNodeMetricDataChanged(data.nodeMetricData)
+		})
+	}
+}
```
BridgeAR, line 76: This is almost identical to the other service. I guess this could share most logic if it's not likely to deviate in the future.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export function edgeMetricData(state: EdgeMetricData[] = setEdgeMetricData().payload, action: EdgeMetricDataAction): EdgeMetricData[] {
+	switch (action.type) {
+		case EdgeMetricDataActions.SET_EDGE_METRIC_DATA:
+			return clone(action.payload)
+		case EdgeMetricDataActions.CALCULATE_NEW_EDGE_METRIC_DATA:
+			return clone(calculateMetrics(action.payload.fileStates, action.payload.blacklist))
```
BridgeAR, line 19: I tried to figure out what `calculateMetrics` does but it was not trivial enough to see if the data is already copied or not. I expect the clone not to be necessary here, but this should be looked at again.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+		const edgeMetricEntry = getEntryForMetric(edgeMetric)
+		addEdgeToNodes(edgeMetricEntry, edge.fromNodeName, edge.toNodeName)
+	}
+}
+
+function getEntryForMetric(edgeMetricName: string): Map<string, EdgeMetricCount> {
+	if (!nodeEdgeMetricsMap.has(edgeMetricName)) {
+		nodeEdgeMetricsMap.set(edgeMetricName, new Map())
+	}
+	return nodeEdgeMetricsMap.get(edgeMetricName)
```
BridgeAR, line 68:
```suggestion
	let nodeEdgeMetric = nodeEdgeMetricsMap.get(edgeMetricName)
	if (nodeEdgeMetric === undefined) {
		nodeEdgeMetric = new Map()
		nodeEdgeMetricsMap.set(edgeMetricName, nodeEdgeMetric)
	}
	return nodeEdgeMetric
```

BridgeAR:
Seems like the e2e tests failed: https://travis-ci.org/github/MaibornWolff/codecharta/jobs/715857956

This should be looked at again. I guess some kind of race condition, recently introduced somewhere.

visualization/app/codeCharta/ui/edgeChooser/edgeChooser.component.html
```diff
@@ -38,7 +39,7 @@
 	</md-select>
 
 	<div class="metric-value" ng-show="$ctrl._viewModel.hoveredEdgeValue">
-		<div class="rounded-box">
+		<div class="rounded-box" id="edge-count">
```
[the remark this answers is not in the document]
NearW: No, the component itself can only be rendered once at a time. Opening it again will close the chooser instead.

The `div.metric-value` is also only rendered once when hovering over a folder.

visualization/app/codeCharta/ui/edgeChooser/edgeChooser.e2e.ts
```diff
@@ -22,5 +28,22 @@ describe("MapTreeViewLevel", () => {
...
 			expect(metrics).toHaveLength(2)
 		})
+
+		it("should display the amount of incoming and outgoing edges next to the metric name", async () => {
+			await edgeChooser.selectEdgeMetric("pairingRate")
+			await searchPanel.toggle()
+			await mapTreeViewLevel.openFolder("/root/ParentLeaf")
+			await mapTreeViewLevel.hoverNode("/root/ParentLeaf/smallLeaf.html")
+
+			expect(edgeChooser.isEdgeCountAvailable()).toBeTruthy()
```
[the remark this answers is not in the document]
NearW: Yes. That's a good idea.

visualization/app/codeCharta/state/store/metricData/metricData.splitter.ts
NearW: Yes. We added this functionality to be able to set the state partially. 
You could call e.g.  `storeService.dispatch(setMetricData( { nodeMetricData: [], edgeMetricData: [] } ))`. The payload is propagated to it's subreducer. In this case, nodeMetricData will be forwarded to the `nodeMetricData.reducer`.

But the really cool functionality here is, that you can also call it with a partial state, e.g. `storeService.dispatch(setMetricData( { nodeMetricData: [] } ))` would also be a valid call. It would be the same as calling `storeService.dispatch(setNodeMetricData([]))`
This becomes really handy when you want to update multiple properties of the state at once. We do this, when we apply the scenarios.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
+import { EdgeMetricDataAction, EdgeMetricDataActions, setEdgeMetricData } from "./edgeMetricData.actions"
+import { BlacklistItem, BlacklistType, Edge, EdgeMetricCount, EdgeMetricData } from "../../../../codeCharta.model"
+import { getVisibleFileStates } from "../../../../model/files/files.helper"
+import { CodeMapHelper } from "../../../../util/codeMapHelper"
+import { FileState } from "../../../../model/files/files"
+import { EdgeMetricDataService } from "./edgeMetricData.service"
+
+const clone = require("rfdc")()
```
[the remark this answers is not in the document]
NearW: Yes. but this change applies to a lot of files. I'd rather wait until [rfdc](https://github.com/davidmarkclements/rfdc/releases) releases a new version which enables the default import.

If the release takes longer, we can work on this in a separate PR.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+		nodeEdgeMetricsMap.set(edgeMetricName, new Map())
+	}
+	return nodeEdgeMetricsMap.get(edgeMetricName)
+}
+
+function addEdgeToNodes(edgeMetricEntry: Map<string, EdgeMetricCount>, fromNode: string, toNode: string) {
+	createEntryIfNecessary(edgeMetricEntry, fromNode)
+	createEntryIfNecessary(edgeMetricEntry, toNode)
+	edgeMetricEntry.get(fromNode).outgoing += 1
+	edgeMetricEntry.get(toNode).incoming += 1
```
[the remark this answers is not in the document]
NearW: Not working. We'd need to distinguish between increasing the outgoing or the incoming edgeCount. I'd rather keep it that way.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,118 @@
...
+	const metricData: EdgeMetricData[] = []
+
+	hashMap.forEach((occurences: any, edgeMetric: any) => {
+		let maximumMetricValue = 0
+		occurences.forEach((value: EdgeMetricCount) => {
+			const combinedValue = value.incoming + value.outgoing
+			if (combinedValue > maximumMetricValue) {
+				maximumMetricValue = combinedValue
+			}
+		})
```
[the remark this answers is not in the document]
NearW: I like the idea, but `.reduce()` is not defined in a  forEach loop I guess

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: The function is part of an interface (in this case `AttributeTypesSubscriber`). I didn't know that you could simply removed parameters from the implemented function.
What is a good practice here? Keep the parameters of the function, even though they might not be used or remove them?

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: Of course store dispatches happen a lot. That's the redux way of changing the state and notifying all subscribers to update their viewModel.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.actions.ts
NearW: Those functions are called `ActionCreators`. These function return an action, which is an object with a type and an optional payload. You put these ActionCreators in the dispatch function, e.g. `storeService.dispatch(setEdgeMetricData([]))`
Don't get to confused by the function itself. It makes much more sense when using the function with dispatch.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
NearW: Probably. This PR just moves the the metricData to the store. If this is a bottleneck worth working on, we can do this in another PR in the future.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: This is not correct. Your suggestion does not take the parameter `numberOfNodes` into account. This parameter is actually `amountOfEdgePreviews`. This means: display x buildings with the highest amount of incoming and outgoing edges.


```ts
const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
return nodeEdgeMetrics === undefined ? [] : [...nodeEdgeMetrics.keys()].slice(0, numberOfNodes)
```

We need to splice at the end. I'll rename the parameter to avoid confusion.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: Absolutely. However, the if-condition should be `if(edgeMetric !== undefined)`.
`!!nodeEdgeMetricsMap.get(it)` is the same as `nodeEdgeMetricsMap.has(it))`

visualization/app/codeCharta/state/store/metricData/metricData.actions.ts
NearW: See my comment on ActionCreators

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
NearW: `restartSystem` sets up the angular module and the services that need to be injected.
We sometimes put logic into this function to set the state before running the test suite. However, that's not always the case.
And if we rename it here, we'd have to rename it all our test classes increasing the amount of files changed here.

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
NearW: When creating redux properties or a subreducer, we have to create a lot of files and adapt some files. This is tedious. That's why we use plop to auto-generate those for us. I'd rather do this in another PR and update the templates accordingly if this is wished. It just takes time.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.splitter.ts
NearW: No. The file gets auto generated. We should therefore update the template and all existing implementations. But again: I don't want to do this in this PR. Otherwise we'll hit 200-300 changed files.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
NearW: I'd move this to another PR. We'll need to update our plop-template to prevent this function to be autogenerated like that

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.actions.ts
NearW: action creator, explained in another comment

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
NearW: The identical part only applies to functions with one line in it, like these
```ts
public onBlacklistChanged(blacklist: BlacklistItem[]) {
	this.storeService.dispatch(calculateNewEdgeMetricData(this.storeService.getState().files, blacklist))
}

public onFilesSelectionChanged(files: FileState[]) {
	this.storeService.dispatch(calculateNewEdgeMetricData(files, this.storeService.getState().fileSettings.blacklist))
}

public onAttributeTypesChanged(attributeTypes: AttributeTypes) {
	this.storeService.dispatch(
			calculateNewEdgeMetricData(this.storeService.getState().files, this.storeService.getState().fileSettings.blacklist)
		)
}
```

We could create a class and inherit it, but this still would require us to call `super()` in each of these

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.reducer.ts
NearW: I disagree. The naming of the functions might be similair, but the implementation is completely different.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
NearW: I looked further into this. We need to fill the map in sorted order. This can be done by sorting the attributes of an edge when our map is empty. 

```ts
function addEdgeToCalculationMap(edge: Edge) {
	const keys = Object.keys(edge.attributes)
	if(nodeEdgeMetricsMap.size === 0) {
		keys.sort()
	}
	for (const edgeMetric of keys) {
		const edgeMetricEntry = getEntryForMetric(edgeMetric)
		addEdgeToNodes(edgeMetricEntry, edge.fromNodeName, edge.toNodeName)
	}
}
```

However, we can't be sure if the first edge has EVERY metric. We might get an edge that only has the git metrics, while other edges contain metrics from another importer.
So we have to wait until we looped through all edges.

visualization/app/codeCharta/state/store/state.splitter.ts
```diff
@@ -43,6 +49,10 @@ export function splitStateActions(action: CCAction): CCAction[] {
 		let actions: CCAction[] = []
 
 		// Plop: Split into sub-reducer here
+		if (action.payload.metricData !== undefined) {
+			actions = actions.concat(...splitMetricDataActions(action.payload.metricData))
```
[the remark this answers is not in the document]
BridgeAR: Thanks for also thinking about the plop file <3

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: I have no publishing rights and I Dave is not always easy to reach. I guess it might still take a while :/

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: I have to look into this again but we might have spoken about different things :)

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: I am not certain I can follow. [`.reduce()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) is a default function on Arrays.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
BridgeAR: Removing them is fine.

The example about removing the first argument is not comparable.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
BridgeAR: What I meant was the dispatch declaration happens frequently with very similar patterns and we might be able to consolidate some of those to use a single function for dispatching multiple to reduce the code overhead and simplify the implementation.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.actions.ts
BridgeAR: I see. Is this a common pattern for Redux?

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
BridgeAR: Thanks, I missed that! Slicing is not ideal, since we would end up doing more work than before. Another alternative is the following (not saying that it's definitely better, I personally just prefer this notation over `.next().value`):

```js
const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
const keys = []
if (nodeEdgeMetrics === undefined) {
  return keys
}
for (const key of nodeEdgeMetrics.keys()) {
  keys.push(key)
  if (keys.length === numberOfNodes) {
    break;
  }
}
return keys
```

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
BridgeAR: > the if-condition should be if(edgeMetric !== undefined)

Indeed!

> !!nodeEdgeMetricsMap.get(it) is the same as nodeEdgeMetricsMap.has(it))

That is actually not correct. The first part checks for in-existent values and values that are falsy. The latter only checks for in-existent values.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.splitter.ts
BridgeAR: SGTM

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
BridgeAR: I personally recommend to rename this at some point while working on something else in a separate commit or in a separate PR while working on something. It's fine for now.

visualization/app/codeCharta/state/store/metricData/metricData.service.spec.ts
BridgeAR: SGTM

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
BridgeAR: SGTM

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.actions.ts
BridgeAR: This is really confusing for someone like me who is not that involved with the codebase :)

NearW: That is true. New developers are aware of the framework redux and are [recommended to read](https://maibornwolff.github.io/codecharta/docs/new-to-code/#state-management) the core concepts of it. [Action creators](https://redux.js.org/basics/actions) are explained in there.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
NearW: I'll try to work on #992 again soon. This has been adressed in there by creating our own api for this function. This makes the imports shorter. :)

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
NearW: Yes, but occurences is a map. :) `hashMap: Map<string, Map<string, EdgeMetricCount>>` and `hashMap.forEach((occurences: any, edgeMetric: any)`

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.actions.ts
NearW: Yes, check out actions [here](https://redux.js.org/basics/actions)
There are some patterns to [reduce this boilerplate code](https://redux.js.org/recipes/reducing-boilerplate). Applying that would solve your confusion, I guess. :)

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: Oh, true. Slicing would require us to traverse the array again. Thanks for the suggestion

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: Awesome, let's keep it as it is in that case! I overlooked the `map` in the variable name.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.actions.ts
BridgeAR: TIL. Thanks for the link!

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: This has been adressed in #1156

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
NearW: Issue created #1157

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
NearW: The only case where the dispatch happens frequently is, when we load a new map. In this case, we first trigger `onFilesSelectionChanged`. After that, `onAttributeTypesChanged` and `onBlacklistChanged` are triggered, because we have to reset the blacklist and update the attributeTypes with those from the new map.

In that case, combining them to a single dispatch would be great, but I don't really know how we can do this. We could use a `debounce-function` to wait 200ms for more events before dispatching.

visualization/app/codeCharta/ui/edgeChooser/edgeChooser.po.ts
```diff
@@ -8,4 +10,28 @@ export class EdgeChooserPageObject {
...
+
+		// Check if the text contains a number.
+		return /\d/.test(innerText)
+	}
+
+	public async getAmountOfEdges(): Promise<EdgeMetricCount> {
+		const innerText = await this.getEdgeCountInnerText()
+		const edgeCount = innerText.split("/")
+
+		return { incoming: parseInt(edgeCount[0]), outgoing: parseInt(edgeCount[1]) }
```
BridgeAR, line 30: `parseInt` and `parseFloat` are strong anti-patterns. They parse faulty data in weird ways (e.g., `parseInt('1aeas9jd8')` === 1???). Instead, always use `Number()` for floats. If you want to guarantee it's an integer, use `Math.floor()` or `Math.ceil()` or `Number.isInteger(Number())` or `| 0` for SMIs (small integers = 2 ** 31 - 1) or `>>> 0` for 2 ** 32 .

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function addEdgeToCalculationMap(edge: Edge) {
+	for (const edgeMetric of Object.keys(edge.attributes)) {
+		const edgeMetricEntry = getEntryForMetric(edgeMetric)
+		addEdgeToNodes(edgeMetricEntry, edge.fromNodeName, edge.toNodeName)
+	}
+}
+
+function getEntryForMetric(edgeMetricName: string): EdgeMetricCountMap {
```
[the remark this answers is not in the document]
NearW: Why do we pass the map here as a parameter? We have a globally defined map, since we need it in other files as well.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
```diff
@@ -0,0 +1,76 @@
...
+		return this.storeService.getState().metricData.nodeMetricData.map(x => x.name)
+	}
+
+	public isMetricAvailable(metricName: string): boolean {
+		return this.storeService.getState().metricData.nodeMetricData.some(x => x.name == metricName)
+	}
+
+	public getMaxMetricByMetricName(metricName: string): number {
+		const metric: NodeMetricData = this.storeService.getState().metricData.nodeMetricData.find(x => x.name == metricName)
+		return metric ? metric.maxValue : undefined
```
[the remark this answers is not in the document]
NearW:
```
if (metric !== undefined) {
    return metric.maxValue
}
```

Isn't this faster though? `.?` checks for `null` and `undefined` like `if(!metric)` does.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: It's mostly easier to follow the logic if the used variables are all inside the function. That way it's also clear how each function interacts with each other.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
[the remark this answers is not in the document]
BridgeAR: I do not have a benchmark but `?.` should produce better code than the other version.

`a ? 1 : undefined` actually checks something like: `if (a !== undefined && a !== null && a !== 0 && a !== '' && a !== 0n && a !== false) { return 1 } else { return undefined }` instead of `if (a === undefined || a === null) { return a } else { return 1 }`.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
```diff
@@ -0,0 +1,115 @@
...
+	let nodeEdgeMetric = nodeEdgeMetricsMap.get(edgeMetricName)
+	if (!nodeEdgeMetric) {
+		nodeEdgeMetric = new Map()
+		nodeEdgeMetricsMap.set(edgeMetricName, nodeEdgeMetric)
```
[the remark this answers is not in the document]
NearW: This doesn't add name and undefined. It adds the metric name and an empty map, which is filled later. We need the empty map here.

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	public getMetrics(): string[] {
+		return this.storeService.getState().metricData.nodeMetricData.map(x => x.name)
+	}
+
+	public isMetricAvailable(metricName: string): boolean {
+		return this.storeService.getState().metricData.nodeMetricData.some(x => x.name == metricName)
+	}
+
+	public getMaxMetricByMetricName(metricName: string): number {
```
BridgeAR, line 54:
```suggestion
	public getMaxMetricByMetricName(metricName: string): number | undefined {
```

I could imagine that metric is always truthy due to just accepting known metricNames (I suggest to use an enum to guarantee that metricName is always known).

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
BridgeAR: `nodeEdgeMetric` is falsy and since `nodeEdgeMetricsMap` only contains maps, it's actually `undefined` :-)

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.reducer.ts
```diff
@@ -0,0 +1,69 @@
...
+
+function addMaxMetricValuesToHashMap(node: HierarchyNode<CodeMapNode>, hashMap: Map<string, MaxMetricValuePair>) {
+	const attributes: string[] = Object.keys(node.data.attributes)
+	attributes.forEach((metric: string) => {
+		const maxMetricValuePair = hashMap.get(metric)
+
+		if (maxMetricValuePair === undefined || maxMetricValuePair.maxValue <= node.data.attributes[metric]) {
+			hashMap.set(metric, {
+				maxValue: node.data.attributes[metric]
+			})
```
BridgeAR, line 48:
```suggestion
			hashMap.set(metric, node.data.attributes[metric])
```

Using the value directly is cheaper :-)

BridgeAR (commented):
Thanks for following up on the comments! This is shaping nicely!

visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
NearW: Yes, we only call this method on metrics we know exist.

How can we build an enum dynamically containing only the metrics that are relevant to the visible files though?

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.reducer.ts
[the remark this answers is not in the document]
NearW: `nodeEdgeMetric` is undefined, when you enter the if-block. However, inside the if-block we set it to an empty map. An empty map is not undefined.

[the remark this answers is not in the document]
BridgeAR: Gosh... absolutely...

NearW:
@BridgeAR Great suggestions. I decided that we should sort the edge-metrics the same way, we sort the node-metrics (by name). This way, I was able to resolve the duplicate of the sort-function.

BridgeAR:
@NearW do you have a before and after screenshot for the change? :-)

BridgeAR (approved):

NearW:
Before:
image

After:
image

Edge-Metrics will be sorted the same way, we sort the node-metrics: By name. Before they were sorted be the max-value of incoming and outgoing edges. Sorting them by max-value isn't a bad idea, but we should handle both metrics the same way in my opinion.

I also pushed another performance improvement.

Issue #1101, opened 2020-07-14
Refactor and speed up e2e tests


NearW:
# Feature request

## Description

Somewhat independent: it should speed up the overall e2e tests if you reuse browser sessions. Right now each e2e file opens a new browser. That takes time to launch. Since each test runs in it's own page, it's fine to reuse a single browser instance. I would normally write something like the part below. There is however an even easier way: https://jestjs.io/docs/en/puppeteer

`puppeteerHelper.js`

```js
let browser

async function launch() {
  if (browser)
    return browser
  browser = puppeteer.launch({
        headless: true
      })
  return browser
}

async function newPage() {
  const browser = await launch() // Implicitly launch the browser
  const page = await browser.newPage()
  await page.setViewport({ width: 1920, height: 1080 })
  return page
}

process.on('exit', () => {
  // This is not really enough... the closing won't be fully awaited and it would have to happen in the global jest setup file in a different way
  browser.close()
})

module.exports = {
  newPage,
}
```

_Originally posted by @BridgeAR in https://github.com/MaibornWolff/codecharta/pull/1100_

I want to refactor the e2e tests, improve performance and add some documentation.

## Acceptance criteria
- [x] Move po functions to their actual po
- [x] Speed up e2e tests by using one browser session and just reload the page with .goTo()
- [x] Define global jest.timeout in jest config
- [x] [Refactor delays to some proper event checking](https://github.com/MaibornWolff/codecharta/pull/1100#discussion_r454021823) 
- [x] Add documentation on how to write proper e2e tests and how to run them (and run a single one)

NearW:
[Documentation has been added](https://github.com/MaibornWolff/codecharta/wiki/E2E-Testing-with-puppeteer)

@JaiJoPa @RomanenkoVladimir as always I'd really appreciate your feedback on it. Let me know if something is not clear or if there are any open questions.

PR #1113, opened 2020-07-26
Tech/1101/move po functions

NearW:
# Move page object functions to their designated page object

works on #1101 

## Description

This PR moves all functions to their designated page object files. 
I also refactored some functions to allow better reusability.

**Would be great to get a quick review on that, because this PR is the base for all the upcoming tasks in this issue. Thanks!**

## Definition of Done

- [ ] **All tests pass**

### DIFF
--- a/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.html
+++ b/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.html
@@ -2,7 +2,6 @@
 	ng-if="!$ctrl.node.isExcluded"
 	class="{{::$ctrl.depth?'':'tree-root'}} tree-element-{{::$ctrl.depth?$ctrl.depth:0}} {{::$ctrl.isLeaf()?'tree-leaf':'tree-parent'}}"
 	ng-init="$ctrl.openRootFolderByDefault($ctrl.depth)"
-	id="{{ ::$ctrl.node.path }}"
 >
 	<div
 		class="tree-element-label-{{::$ctrl.depth?$ctrl.depth:0}} tree-element-label"
@@ -11,6 +10,7 @@
 		ng-mouseleave="$ctrl.onMouseLeave()"
 		ng-right-click="$ctrl.openNodeContextMenu($event)"
 		ng-click="$ctrl.onClickNode()"
+		id="{{ ::$ctrl.node.path }}"
 	>
 		<span class="node-type-icon">
 			<span
--- a/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
+++ b/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
@@ -31,7 +31,7 @@ describe("MapTreeViewLevel", () => {
 
 	describe("Blacklist", () => {
 		it("excluding a building should exclude it from the tree-view as well", async () => {
-			await searchPanelModeSelector.openTreeView()
+			await searchPanelModeSelector.toggleTreeView()
 			await mapTreeViewLevel.openContextMenu()
 			await nodeContextMenu.exclude()
 
--- a/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
+++ b/visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
@@ -31,12 +31,14 @@ describe("MapTreeViewLevel", () => {
 
 	describe("Blacklist", () => {
 		it("excluding a building should exclude it from the tree-view as well", async () => {
+			const nodePath = "/root/ParentLeaf/smallLeaf.html"
+
 			await searchPanelModeSelector.toggleTreeView()
 			await mapTreeViewLevel.openFolder("/root/ParentLeaf")
-			await mapTreeViewLevel.openContextMenu("/root/ParentLeaf/smallLeaf.html")
+			await mapTreeViewLevel.openContextMenu(nodePath)
 			await nodeContextMenu.exclude()
 
-			expect(await mapTreeViewLevel.nodeExists("/root/ParentLeaf/smallLeaf.html")).toBeFalsy()
+			expect(await mapTreeViewLevel.nodeExists(nodePath)).toBeFalsy()
 		})
 	})
 })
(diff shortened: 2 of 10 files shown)


visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.po.ts
```diff
@@ -3,12 +3,19 @@ import { Page } from "puppeteer"
...
+	public async openFolder(path: string) {
+		await this.page.click(`[id='${path}'] > div`)
+	}
+
+	public async hoverNode(path: string) {
+		await this.page.hover(`[id='${path}'] > div`)
+	}
+
+	public async nodeExists(path: string) {
+		return !!(await this.page.$(`[id='${path}'] > div`))
```
BridgeAR, line 19: Is the `> div` part actually necessary? I would expect an ID to be unique?

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
```diff
@@ -31,11 +31,12 @@ describe("MapTreeViewLevel", () => {
...
 		it("excluding a building should exclude it from the tree-view as well", async () => {
-			await searchPanelModeSelector.openTreeView()
-			await mapTreeViewLevel.openContextMenu()
+			await searchPanelModeSelector.toggleTreeView()
+			await mapTreeViewLevel.openFolder("/root/ParentLeaf")
+			await mapTreeViewLevel.openContextMenu("/root/ParentLeaf/smallLeaf.html")
 			await nodeContextMenu.exclude()
 
-			expect(await mapTreeViewLevel.nodeExists()).toBeFalsy()
+			expect(await mapTreeViewLevel.nodeExists("/root/ParentLeaf/smallLeaf.html")).toBeFalsy()
```
BridgeAR, line 39: Nit: it is nice to use a variable for strings that are reused in the same function.

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
```diff
@@ -4,8 +4,8 @@ import { delay } from "../../../puppeteer.helper"
 export class SearchPanelModeSelectorPageObject {
 	constructor(private page: Page) {}
 
-	public async openTreeView() {
-		await this.page.click("search-panel-mode-selector-component #tree-view")
+	public async toggleTreeView() {
+		await this.page.click("#tree-view")
 		await delay(500)
```
BridgeAR, line 9: Suggestion: it might be useful to return the state (if the tree view is now enabled or disabled) as boolean or string. That way it's possible to check that in the test.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.po.ts
NearW: It was necessary, when clicking on folders. But I moved the id to the inner div, so that we don't need that anymore.
The inner div is where the click listeners are located anyway. So it all makes sense now 👍

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
NearW: Instead of returning a boolean here, I added a new function `isTreeViewOpen`.
It was not tested before if `toggleTreeView` actually works (I just saw it does when running the tests headful), so I added a test for this function now.

NearW:
@BridgeAR Thanks for the quick review. Everything has been adressed. PTAL

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.html
```diff
@@ -11,6 +10,7 @@
 		ng-mouseleave="$ctrl.onMouseLeave()"
 		ng-right-click="$ctrl.openNodeContextMenu($event)"
 		ng-click="$ctrl.onClickNode()"
+		id="{{ ::$ctrl.node.path }}"
```
BridgeAR, line 13: I hope this does not change the event propagation in a way that other parts are highlighted / active / similar due to the id change?

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.e2e.ts
```diff
@@ -0,0 +1,35 @@
...
+	})
+
+	it("should open the tree-view when clicking on the tree-view icon", async () => {
+		await searchPanelModeSelector.toggleTreeView()
+
+		expect(await searchPanelModeSelector.isTreeViewOpen()).toBeTruthy()
+	})
+
+	it("should open and close tree-view when clicking on the tree-view icon twice", async () => {
+		await searchPanelModeSelector.toggleTreeView()
```
BridgeAR, line 30:
```suggestion
```

It is better to just do both of these things in a single test. The reason is that you would otherwise also have to verify that the first call actually opened it properly. Doing it in the same test also reduces the runtime. These are both favorable things.

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
BridgeAR: Thanks, that's definitely an improvement. The reason why I would still implement it with the return value though is that it's not guaranteed that the test conditions could in theory differ in a way that the new `toggleTreeView` would work as expected but that it works differently under other circumstances.

A reason for the current implementation might be that it's not that obvious how to track what page clicked how many times. That is possible by using a `WeakMap`. The key is the page and the value is the number of clicks. That way it's possible to know if it should be open or closed. And it would of course also be possible to wait for the addition or removal of the `current` class from `#tree-view` to remove the delay.

That's just a suggestion though and I am fine with it is as well.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.html
NearW: No, not at all. The id here is not used in our production code. It's only there to access the tree-structure easily.

Introduced here:
https://github.com/MaibornWolff/codecharta/pull/1100/files#diff-58b5703126f177e3766594a6099259ac

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
NearW: I think think that checking if the tree-view is open or not by simply checking if the classNames include `current` is way more convinient than keeping track of the current clicks in a map.

```
public async isTreeViewOpen() {
	const classNames = await this.page.$eval("#tree-view", el => el.className)

	return classNames.includes("current")
}
```

This has been implemented already.

I'll get rid of the delays in another PR. The next PR will introduce jest-puppeteer and speed things up with a different setup and teardown. After that I'll replace all delays with `waitForSelector()` calls.

PR #1115, opened 2020-07-28
Tech/1101/use jest puppeteer

NearW:
# Add jest-puppeteer and speed up e2e tests

_**Please read the [CONTRIBUTING.md](https://github.com/MaibornWolff/codecharta/blob/master/CONTRIBUTING.md) before opening a PR.**_

Issue: #1101 

## Description

This PR is not ready for a review yet and I encountered some issues that need further investigation.

**What has been done?**
- Added jest-puppeteer and used it's global variables for a much cleaner and faster setup and teardown
- Merged jest presets
- Refactored npm scripts for better reusability
- Reload page before every test instead of creating a new browser session
- Use expect-puppeteer-click-event, because it checks if the selector is available. This increases stability.
- Use `waitForSelector()` every time before accessing a selector to increase stability
- Removed all delays and instead use `waitForSelector` 
- **Decreased testing time from 130s to 68s**

**Future**
It would be great, if we could write our own puppeteer-api, so that when we call `page.$eval(selector)` we always call `page.waitForSelector(selector)` before. We could get rid of like 30% of the code in our PageObjects.
However, it seems like there is a problem with jest-puppeteer and babel. When passing a selector as a parameter, the selector will be undefined.

**Example**
_Before:_
```
public async isTreeViewOpen() {
	await page.waitForSelector("#tree-view")
	const treeViewClassNames = await page.$eval("#tree-view", el => el.className)

	await page.waitForSelector("#search-panel-card")
	const searchPanelClassNames = await page.$eval("#search-panel-card", el => el.className)

	return treeViewClassNames.includes("current") && searchPanelClassNames.includes("expanded")
}
```

_After:_
```
public async isTreeViewOpen() {
	return await classNameIncludes("#tree-view", "current") &&
        await classNameIncludes("#search-panel-card", "expanded")
}

async function classNameIncludes(selector:string, className: string) {
	const classNames = await getClassNames(selector)
	return classNames.includes(className)
}

async function getClassNames(selector: string) {
	await page.waitForSelector(selector)
	return await page.$eval(selector, el => el.className)
}
```


visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
```diff
@@ -1,17 +1,23 @@
...
-		await delay(500)
+		const wasOpen = await this.isTreeViewOpen()
+
+		await expect(page).toClick("#tree-view", { timeout: 3000 })
+
+		if (wasOpen) {
+			await page.waitForSelector("#search-panel-card", { visible: false })
+		} else {
+			await page.waitForSelector("#search-panel-card.expanded")
+		}
```
BridgeAR, line 11: I still suggest to return `!wasOpen`. That way the test has to do less work.

Right now there's a test that looks like:

```js
expect(await searchPanelModeSelector.isTreeViewOpen()).toBeFalsy()
await searchPanelModeSelector.toggleTreeView()
expect(await searchPanelModeSelector.isTreeViewOpen()).toBeTruthy()
```

If you return `!wasOpen` it could be:

```js
const isOpen = await searchPanelModeSelector.toggleTreeView()
expect(isOpen).toBeTruthy()
```
Thus, you safe two async calls and it's less code. The same applies to the other toggle functions as well.

BridgeAR:
The branch has some conflicts.

visualization/app/codeCharta/ui/searchPanelModeSelector/searchPanelModeSelector.po.ts
NearW: If we return `!wasOpen` we don't actually know if the tree-view is actually open. We just assume that the button toggled it correctly.

```
public async toggleTreeView() {
	const wasOpen = await this.isTreeViewOpen()

	await expect(page).toClick("#tree-view", { timeout: 3000 })

	if (wasOpen) {
		await page.waitForSelector("#search-panel-card", { visible: false })
	} else {
		await page.waitForSelector("#search-panel-card.expanded")
	}
        return !wasOpen
}
```

```
it("should open and close the tree-view when clicking on the tree-view icon", async () => {
	let isOpen = await searchPanelModeSelector.toggleTreeView()
	expect(isOpen).toBeTruthy()

	isOpen = await searchPanelModeSelector.toggleTreeView()
	expect(isOpen).toBeFalsy()
})
```

We can't verify that the last `await searchPanelModeSelector.toggleTreeView()`-call actually closed the tree-view, because we just return a negation of the previous state. We can't be sure about the actual state.

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.e2e.ts
```diff
@@ -61,69 +51,59 @@ describe("RibbonBar", () => {
...
-
-			expect(await ribbonBar.isPanelOpen(panel)).toBeFalsy()
+			isAreaSettingsPanelOpen = await ribbonBar.togglePanel(panel)
+			expect(isAreaSettingsPanelOpen).toBeFalsy()
 		})
 
 		it("color-metric card", async () => {
 			const panel = "color-metric"
 
 			expect(await ribbonBar.isPanelOpen(panel)).toBeFalsy()
```
BridgeAR, line 88:
```suggestion
```

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.e2e.ts
```diff
@@ -61,69 +51,59 @@ describe("RibbonBar", () => {
...
-
-			expect(await ribbonBar.isPanelOpen(panel)).toBeFalsy()
+			isColorSettingsPanelOpen = await ribbonBar.togglePanel(panel)
+			expect(isColorSettingsPanelOpen).toBeFalsy()
 		})
 
 		it("edge-metric", async () => {
 			const panel = "edge-metric"
 
 			expect(await ribbonBar.isPanelOpen(panel)).toBeFalsy()
```
BridgeAR, line 100:
```suggestion
```

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.e2e.ts
```diff
@@ -1,39 +1,28 @@
...
+		await goto()
 	})
 
 	describe("Blacklist", () => {
 		it("excluding a building should exclude it from the tree-view as well", async () => {
 			const filePath = "/root/ParentLeaf/smallLeaf.html"
 
-			await searchPanelModeSelector.toggleTreeView()
+			const isTreeViewOpen = await searchPanelModeSelector.toggleTreeView()
+			expect(isTreeViewOpen).toBeTruthy()
```
NearW, line 24: Do we really have to assert this line? togging the treeview is tested in `searchPanelModeSelector` already.

BridgeAR: I am good with not testing it again.

visualization/plop/templates/ui/po.ts.hbs
```diff
@@ -1,10 +1,7 @@
-import { Page } from "puppeteer"
-
 export class {{properCase name}}PageObject {
-	constructor(private page: Page) {
-	}
 
     public async doSomething() {
-    	return await this.page.$eval("A > CSS_SELECTOR", el => el["someAttribute"])
+        await page.waitForSelector("CSS_SELECTOR")
+    	return await page.$eval("A > CSS_SELECTOR", el => el["someAttribute"])
```
BridgeAR, line 5:
```suggestion
    	  return await page.$eval("A > CSS_SELECTOR", el => el["someAttribute"])
```

Issue #783, opened 2019-11-13
Update releases in github-pages automatically


NearW:
We have a release category in our [gh-pages](https://maibornwolff.github.io/codecharta/categories/#release) now and we need to update it automatically, when releasing. This can be done by creating the required files and move them to the gh-pages folder BEFORE making the release commit. When we release, we will do two commits. One updates the gh-pages, the other one tags the release.

Please note that you probably want to extract the latest unreleased patch notes. We already wrote a script for this earlier [here](https://github.com/MaibornWolff/codecharta/issues/533#issuecomment-505508299)

BridgeAR:
Is this still something that needs to be addressed?

NearW:
Yes, please. This would get rid of the release preparation PRs before evey release.

#voteforhigherpriority

BridgeAR:
Go for it

PR #1119, opened 2020-07-30
Tech/783/update release post automatically

NearW:
# Add gh-pages release post automatically

_**Please read the [CONTRIBUTING.md](https://github.com/MaibornWolff/codecharta/blob/master/CONTRIBUTING.md) before opening a PR.**_

Issue: #783 

## Description
Took some time, but it was actually quite fun to write a script with python (never done before).

- Converted the `make_release.sh` script into a python script
- Added release-posts to gh-pages automatically on release now
- Added pipfile to manage dependencies and script calls

This script has been manually tested on windows so far, but I couldn't test the commit/push part. I'd like to test this on a release later then.

## How to use
- Install Python >3.8
- Install pipenv by running `pip install pipenv`
- Install all dependencies by running `pipenv install` from the root directory
- Run the make_release script by calling `pipenv run make_release` from the root directory


script/make_release.py
```diff
@@ -0,0 +1,225 @@
...
+
+release_type = PyInquirer.prompt(questions)["version"]
+
+if release_type == None:
+    quit()
+if release_type == "Major":
+    new_version = new_major_version
+if release_type == "Minor":
+    new_version = new_minor_version
+if release_type == "Patch":
```
BridgeAR, line 119: I suggest to compare `release_type` in lower case.

script/make_release.py
```diff
@@ -0,0 +1,225 @@
...
+    "choices": ["Major", "Minor", "Patch"],
+}]
+
+release_type = PyInquirer.prompt(questions)["version"]
+
+if release_type == None:
+    quit()
+if release_type == "Major":
+    new_version = new_major_version
+if release_type == "Minor":
```
BridgeAR, line 117: Ideally, these are all `if ... else if ... else ...`. Otherwise the CPU still has to do the computations.

In this specific case it would also be good to have the else case be a warning and `quit()` as we likely do not want to accept values other than `major`, `minor` and `patch`?

script/make_release.py
```diff
@@ -0,0 +1,225 @@
...
+            fp.write(line)
+
+print(f"v{new_version}")
+print("incremented version in ./analysis/gradle.properties")
+
+
+# bump version in package.jsons
+analysis_package = f"{root}/analysis/node-wrapper"
+analysis_package_json = f"{analysis_package}/package.json"
+analysis_package_lock_json = f"{analysis_package}/package-lock.json"
```
BridgeAR, line 147: I could not figure out what's done with the package-lock. Why would that change?

script/make_release.py
NearW: That is true, but there are not other value selectable in the prompt.

script/make_release.py
NearW: Because we bump the version in the package.json and package-lock.json using npm version

```
# bump version in package.jsons
npm --prefix ./analysis/node-wrapper --no-git-tag-version version $NEW_VERSION
echo "incremented version in ./analysis/node-wrapper/package.json + locks"

npm --prefix ./visualization --no-git-tag-version version $NEW_VERSION
echo "incremented version in ./visualization/package.json + locks"
```

script/make_release.py
NearW: The release_type can only be a value from the choices-array `"choices": ["Major", "Minor", "Patch"]`. Lowering both strings is overengineering imo.

BridgeAR: I agree. I expected it to be a text field :-)

script/make_release.py
BridgeAR: Awesome. In that case the only change I would still do is to use `elif`. Mainly to keep it in mind, since this is absolutely no hot code path ;-)

NearW:
@BridgeAR PTAL

PR #1139, opened 2020-08-03
Use closes prefix to close issues after merging a pr

NearW:
# Use closes prefix to close issues after merging a pr

Issue: #

## Description

I thought linking issues with zenhub is enough to do so as well, but I was wrong.
Let's use the old [github provided solution](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) again

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,7 @@
 
 _**Please read the [CONTRIBUTING.md](https://github.com/MaibornWolff/codecharta/blob/main/CONTRIBUTING.md) before opening a PR.**_
 
-Issue: #
+closes: #
 
 ## Description
 


BridgeAR:
Note: I would expect the person who opens the PR to change the reference to `Closes` / `Fixes` if it actually does that. Otherwise we'll close issues that should not be closed.

NearW:
> Note: I would expect the person who opens the PR to change the reference to `Closes` / `Fixes` if it actually does that. Otherwise we'll close issues that should not be closed.

But that's mostly the case. That's why I'd set the default to `Closes`. The person who opens the PR can change it to `Fixes` or `Issue` if it's not closing anything.

BridgeAR:
I am certain that it's the most frequent case. But I would rather not close issues than not close them. But it's your call. I would just expect people to forget to change it in case it should not close something.

Issue #1036, opened 2020-07-04
Option not hidden when clicking on a node or other option


BridgeAR:
# Bug

## Expected Behavior

GIVEN e.g., a metric option is open

WHEN clicking on a node on the map or on another option

THEN the metric option is collapsed

## Actual Behavior

The metric option is still visible

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/86512237-abcf7a00-be00-11ea-9039-e51b19a0e605.png)
![image](https://user-images.githubusercontent.com/8822573/86512344-9e66bf80-be01-11ea-9556-40354a3aefc3.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.49.1
- OS: Ubuntu 20.04
- Browser: Chromium

NearW:
![metric visible](https://user-images.githubusercontent.com/12533626/86521018-874cbf80-be4b-11ea-9efa-563ff19d8056.gif)

Works for me.

BridgeAR:
@NearW what I meant is also visible in your example: the expanded mcc part should be hidden when clicking somewhere else.

NearW:
@BridgeAR Oh, you mean the color-range-slider and the two checkboxes?
For now this was intentional design, but you're right. I don't see a reason for those to be open.

PR #1141, opened 2020-08-04
Fix/1036/close panels when clicking somewhere else

NearW:
# Automatically close ribbon bar sections when clicking somewhere else in the application

fixes: #1036 

## Description

- Add new redux property `panelSelection` keeping track of which panel is open
- Refactor how we set the `expanded` className to toggle the sections
- Add `CodeChartaMouseEventService` to handle closing components in a expandable way
- Add some new tests incl. regression e2e test

## Screenshots or gifs

toggling-panels


visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.html
```diff
@@ -9,12 +9,16 @@
 	</div>
 </md-card>
 
-<md-card ng-attr-id="area-metric-card">
+<md-card ng-attr-id="area-metric-card" ng-class="{ expanded: $ctrl._viewModel.panelSelection === 'AREA_PANEL_OPEN' }">
 	<div class="section">
```
alschmut, line 13: Do you want to set an enum variable `panelSelectionValues: PanelSelection` into the viewModel, so you can replace the string values (also below)? :)

NearW: Good suggestions. Done.

Another way would be to provide functions from the controller, each checking if a certain panel is open. This would result in n functions, where n is the number of entries in the enum.
All functions implement something like
```ts
public isAreaPanelOpen() {
	return this._viewModel.panelSelection === PanelSelection.AREA_PANEL_OPEN
}
```

Hard to tell which approach is better though

BridgeAR:
Seems like all open PRs need a rebase and lots of unrelated changes got in somehow?

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
```diff
@@ -1,18 +1,41 @@
...
+		private $rootScope: IRootScopeService,
+		private codeChartaMouseEventService: CodeChartaMouseEventService
+	) {
+		PanelSelectionService.subscribe(this.$rootScope, this)
+	}
 
-	public toggle(event) {
-		const boxElement = $(event.srcElement).closest("md-card")
-		$(document.activeElement).blur()
+	private _viewModel: {
```
BridgeAR, line 19: Using an underscore to indicate private methods is not ideal. Instead, use an actual private property `#viewModel`.

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
```diff
@@ -1,18 +1,41 @@
...
+	}
+
+	public toggle(panelSelection: PanelSelection) {
+		$(document.activeElement).blur()
+		if (this._viewModel.panelSelection !== panelSelection) {
+			this.storeService.dispatch(setPanelSelection(panelSelection))
 		} else {
-			boxElement.addClass(this.EXPANDED_CLASS)
+			this.storeService.dispatch(setPanelSelection(PanelSelection.NONE))
 		}
```
BridgeAR, line 37: This could also be written as:

```js
const selected = this.#viewModel.panelSelection !== panelSelection ? PanelSelection.NONE : panelSelection
this.storeService.dispatch(setPanelSelection(selected))
```

If you want to prevent unnecessary dispatch operations you could also check if both are already `NONE`. It's probably fine as it is though.

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
```diff
@@ -0,0 +1,31 @@
+import { StoreService } from "./state/store.service"
+import { PanelSelection, SearchPanelMode } from "./codeCharta.model"
+import { setPanelSelection } from "./state/store/appSettings/panelSelection/panelSelection.actions"
+import { setSearchPanelMode } from "./state/store/appSettings/searchPanelMode/searchPanelMode.actions"
+
+export class CodeChartaMouseEventService {
+	constructor(private storeService: StoreService) {}
+
+	public closeComponentExcept(fun?: Function) {
```
BridgeAR, line 9: Should this not always be called with an actual function?

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
```diff
@@ -0,0 +1,31 @@
...
+		const functions: Function[] = [this.closeRibbonBarSections, this.closeSearchPanel]
+
+		functions
+			.filter(x => x !== fun)
+			.forEach(fn => {
+				fn = fn.bind(this)
+				fn()
+			})
```
BridgeAR, line 17:
```suggestion
			.forEach(fn => fn.call(this))
```

visualization/app/codeCharta/state/store/appSettings/panelSelection/panelSelection.actions.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export interface SetPanelSelectionAction extends Action {
+	type: PanelSelectionActions.SET_PANEL_SELECTION
+	payload: PanelSelection
+}
+
+export type PanelSelectionAction = SetPanelSelectionAction
+
+export function setPanelSelection(panelSelection: PanelSelection = defaultPanelSelection): SetPanelSelectionAction {
```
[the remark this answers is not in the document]
NearW: action creators. discussed in another pr.

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
NearW: I never really thought about why we'd put an underscore before that. Since I'm working with codeCharta, viewModels were always prefixed with an underscore. I think they shouldn't prefixed at all.

Can private properties from es6-classes be used with typescript and set to a type, e.g. `#viewModel: MY_TYPE = {}`?
I don't think our configs support private properties right now and using them would require a huge refactoring in all classes.

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -67,9 +68,7 @@ export class CodeChartaController {
 	}
 
 	public onClick() {
-		if (this.storeService.getState().appSettings.searchPanelMode !== SearchPanelMode.minimized) {
-			this.storeService.dispatch(setSearchPanelMode(SearchPanelMode.minimized))
-		}
+		this.codeChartaMouseEventService.closeComponentExcept()
```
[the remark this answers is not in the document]
NearW: This was a complicated issue. We have a global event listener that registers a click to close the search-panel for example, when clicking somewhere else. Now, if we click in the search-bar, the search-bar will open. The event-propagation is stopped to prevent the listener from closing the search-bar again.

Now, we want to do the same with the other panels in this PR. We open one of the metric-panels and then click on the search panel. The click-event won't be propagated and the metric-panel will not close. Instead, both the search-panel and the metric-panel would be open.

To solve this issue, we can just close all panels by dispatching an action (triggering a redux event), but this would close the search-panel we just opened. That's why this function has a list of actions, that would close panels and you can provide an optional function as a parameter that would be triggered from the list. I think we should list those functions in an enum. So for the searchBar, we'd call `this.codeChartaMouseEventService.closeComponentExcept(this.codeChartaMouseEventService.closeSearchPanel)`

gh-pages/_posts/release/2020-08-07-v1_52_0.md
```diff
@@ -1,28 +1,27 @@
...
+- Rename master branch to main for a more inclusive naming #1117
+
+### Fixed 🐞
+
+- After loading an invalid file the filechooser pops up again, so that the user can choose a valid file #1021
+- Quality gates on sonarcloud.io are available again #879
+
+### Docs 🔎
+
+- Moved developer guides to our [gh-pages](https://maibornwolff.github.io/codecharta/) #986
```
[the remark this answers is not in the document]
NearW: No. Merging the main branch resolved the unrelated changes.

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
NearW: Your suggested ternary expression is wrong. It should be

```ts
const newSelection = this._viewModel.panelSelection !== panelSelection ? panelSelection : PanelSelection.NONE
this.storeService.dispatch(setPanelSelection(newSelection))
```

There are no unnecessary dispatches. The toggle is triggered by the user and a toggle always triggers an event, either close or open.

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
NearW: No. This function is triggered with no function, when clicking on the codeMap now. In this case, you want both functions in the array to be called in order to close all panels.

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
```diff
@@ -0,0 +1,30 @@
...
+	constructor(private storeService: StoreService) {}
+
+	public closeComponentExcept(fun?: Function) {
+		const functions: Function[] = [this.closeRibbonBarSections, this.closeSearchPanel]
+
+		functions.forEach(fn => {
+			if (fn !== fun) {
+				fn.call(this)
+			}
+		})
```
BridgeAR, line 16:
```suggestion
		if (fun !== this.closeRibbonBarSections) {
			this.closeRibbonBarSections()
		}
		if (fun !== this.closeSearchPanel) {
			this.closeSearchPanel()
		}
```

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
BridgeAR: I agree that they should not be prefixed.

https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields

If it's too much work to refactor, let's not do that now :-)

visualization/app/codeCharta/codeCharta.component.ts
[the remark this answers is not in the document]
BridgeAR: Ah, in that case maybe name it `closeComponentExceptCurrent` or `closeAllButLastComponent`?

BridgeAR (approved):

visualization/app/codeCharta/codeCharta.component.ts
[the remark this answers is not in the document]
NearW: `closeComponentExceptCurrent` sounds good. I also renamed the parameter to `currentCloseFunction`. Hopefully, this makes things clearer.

Issue #1027, opened 2020-07-04
Right click to move the position causes node options to show


BridgeAR:
# Bug

## Expected Behavior

GIVEN an open map

WHEN clicking with the right mouse button on parts of the map without releasing the cursor and moving the mouse

THEN the map moves without showing the node options of the node that was clicked upon

## Actual Behavior

GIVEN an open map

WHEN clicking with the right mouse button on parts of the map without releasing the cursor and moving the mouse

THEN the map moves but the node options of the node that was clicked upon are visible

## Steps to Reproduce the Problem

1. Open a map
1. Click with the right mouse button on a node without releasing the button
1. The node option is visible immediately

## Suggested fix

It should be enough to only trigger the node options in case the mouse button is _released_, not when clicked.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.49.1
- OS: All
- Browser: All

NearW:
Regarding the suggested fix:
We should also check if the mouse moved since last right-clicked. If that's the case, we shouldn't show the context menu at all. To avoid bad user experience, we should also define a threshold in which the mouse has to move in order to not show the context menu. Something like 10 should be enough I guess.

BridgeAR:
I very much agree. I missed to outline that in the suggestion! I would actually not accept any movement from the actual click to the release (but that might be to strict. In that case your suggestion would be the ideal way).

PR #1143, opened 2020-08-05
Fix/1027/dont show node context menu when mouse moved

NearW:
# NodeContextMenu showing up after moving the mouse while holding right-mouse-button

fixes: #1027

## Description

- NodeContextMenu will show up on mouseup instead of mousedown
- Keep track of mouse position, when we clicked somewhere the last time
- Only show up the NodeContextMenu if mouse didn't move after clicking
- Removed all `.mockReturnValue(null)` calls, since `jest.fn()` returns undefined. This is enough for the tests
- Reorderd functions. There were private functions inbetween and the organization in the file was a mess
- Removed `clickType` property. Instead of saving the clickType onmousedown, we just check for the button in the event on mouseup
- Made lots of functions private and adapted tests to test on a higher level (this applies to onmouseup now)
- `this.mouse` will now contain HTML-Coordinates. We transform these to Scene-Coordinates before raycasting into the scene

## Note
Unfortunately the diff is a mess, due to all the reorders and restructuring of methods and tests. This needs a review in an IDE I guess. :(

## Screenshots or gifs
nodecontextmenu


BridgeAR:
Seems like this needs a rebase to master? Some changes are in here that do not seem related? :)

Issue #901, opened 2020-03-27
Excluding the last building on the map results in an error, reincluding the buildings makes them all the same size


RomanenkoVladimir:
# Bug
Console Error after last building is excluded, Error in Building Size and Color afterwards if buildings are included again
## Expected Behavior
When the user excludes the last building the map is empty, reincluded buildings look the same as before exclusion.
## Actual Behavior
When the user excludes the last building on the map a console error is thrown, reincluding buidlings afterwards results in all of them being color, due to the metric switching to unary
## Steps to Reproduce the Problem
  1. Exclude all buildings
  2.  The Web Application throws a TypeError: "reduce of empty array with no initial value"
  3. Reincluding buildings makes them all red, metric switches to unary

## Specifications
  - OS: Windows
  - Browser: Firefox, Chrome

alschmut:
I don't want to defend this behaviour, its not ideal, but I wanna mention that this is strictly not a bug but by design. 
- The root node includes the metric (attributes) of its children. In this case the root node does only contain the fake metric `unary`, as it has no children. 
- And whenever the available metrics change, the selected metrics reset themselves to the first available metric, which in the case is `unary`

RomanenkoVladimir:
Ah ok, I understood what led to the behaviour but still thought a user would expect a similar color scheme to the one before the exclusion. If this is a decision by design, I was wrong here.

However, the console error still remains.

Should I change the issue or close it?

alschmut:
Yes you are right, we could think of another solution here. Not sure what you mean with color scheme though, as there is not other available metric nor more than one node ;) 

Nah, lets keep the issue as a bug (as it looks like one from a user perspective). But it requires more like a workaround than a fix.

BridgeAR:
I had another look at this. There are actually different ways how a user might end up with no nodes at all.

If a user clicks on a building or folder that would end up in having no nodes left by excluding them, I would just open a popup warning that it's not possible to hide all nodes.

Another way to exclude all nodes is by e.g., searching for `*` and then to use that as an exclude pattern. I would open the same popup that notifies the user that this is not possible.

To do so, we could just check if the exclude operation would result in having no nodes left. If that's the case, open the mentioned notification and prevent the actual operation from succeeding (i.e., the exclude pattern from the search should not be added to the pattern list).

I changed the priority of this issue, since it's pretty easy to run into this and it ends up with very weird behavior afterwards.

PR #1152, opened 2020-08-12
Fix/901/excluding last building error

NearW:
# Show error dialog when all buildings would be excluded

fixes: #901 

## Description

This fix was a bit tricky. Before decorating the blacklist-attributes onto the renderMap, we have to check, if the number of excluded buildings would equal the number of buildings in general. If that's the case, show a dialog and revert the latest blacklist entry.
This has been achieved by refactoring the dataflow in the `codeMapPreRenderService` and splitting the decoration process.
**Before**: When we blacklist something, we recalculate the `metricData` and then trigger a `metricDataAddedEvent`.
**Now**: When we blacklist something, we recalculate the `metricData` and then trigger a `metricDataAddedEvent`. However, we only decorate the map, if the map has not been decorated yet. We call this `decorateNewMap`. This only needs to be done once on a completly new map. For an already decorated map, we use `decorateExistingMap`. This reduces unnecessary calculations and should improve the performance slightly.

- Also fixes a problem with all excluded building displayed with the same path

## Screenshots or gifs

exclude-last


visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -47,6 +46,29 @@ export class NodeDecorator {
...
 	}
 
+	public static doesExclusionResultInEmptyMap(map: CodeMapNode, blacklist: BlacklistItem[]): boolean {
+		let excludedNodes = 0
+		const flattened = ignore()
+		const excluded = ignore()
+
+		for (const item of blacklist) {
+			const path = CodeMapHelper.transformPath(item.path)
+			item.type === BlacklistType.flatten ? flattened.add(path) : excluded.add(path)
```
BridgeAR, line 56: Please always use `if (foo) { func(bar) } else { otherFunc(bar) }` instead of the ternary operator when calling functions. Maybe use a lint rule for it :)

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.service.ts
```diff
@@ -15,14 +19,30 @@ export interface BlacklistSubscriber {
...
+					this.codeMapPreRenderService.getRenderMap(),
+					this.storeService.getState().fileSettings.blacklist
+				)
+			) {
+				this.storeService.dispatch(setIsLoadingMap(true))
+				this.notify(this.select())
+			} else {
+				this.storeService.dispatch(removeLastBlacklistItem(), true)
+				this.dialogService.showErrorDialog("Excluding all buildings is not possible.", "Blacklist Error")
+			}
```
BridgeAR, line 45: Is it possible to combine the logic in a way that it actually does remove all elements first, then you check if there are still elements and if there are, it's done, but if there are not, then you undo the last operation?

That would be more efficient :)

visualization/app/codeCharta/util/nodeDecorator.ts
NearW: Done. I couldn't find a rule that checks for non-return calls in ternary though. :(

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.service.ts
NearW: If you do it that way, you'll first trigger that the blacklist has changed. Based on that event a lot of different services and components change their state, e.g. we calculate new metric data, since the maxValue might have changed. This would also trigger the re-decoration of the map.

We'd need to clone the previous map first in order to reset it, if all buildings will be excluded after decoration. We'd also need to notify all components and service, that they need to update their state again to the previous state.

BridgeAR: I personally would still do this, if it's reversible. It would be an optimization for the common case and a slow case for the case all folders are removed. This comment should not be considered blocking though.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: Seems like you missed this one? :-) let's check for such rule later on. Seems like `flattened` is not used here at all though?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -48,6 +51,29 @@ export class NodeDecorator {
...
+		}
+
+		const tree = hierarchy(map)
+		const desc = tree.descendants()
+
+		desc.map(node => {
```
BridgeAR, line 67: If `desc` is an array, it's better to use `some` or to use a plain for loop. If it's an iterator, using a for loop would also be best, since it's possible to stop the loop as soon as the excludedNodes exceed the tree leaves length.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -48,6 +51,29 @@ export class NodeDecorator {
...
+			const path = CodeMapHelper.transformPath(node.data.path)
+			if (excluded.ignores(path) && CodeMapHelper.isLeaf(node.data)) {
+				excludedNodes++
+			}
```
BridgeAR, line 71: Instead of counting the excluded nodes, let's check for any node that is not excluded and stop right there. That way it's significantly less work :-) That way there's also no need to check the tree leaves length.

visualization/app/codeCharta/util/codeMapHelper.ts
```diff
@@ -78,6 +78,10 @@ export class CodeMapHelper {
 		return markingColor
 	}
 
+	public static isLeaf(node: CodeMapNode): boolean {
+		return !node.children || (node.children && node.children.length === 0)
```
BridgeAR, line 82:
```suggestion
		return !(node.children?.length)
```

visualization/app/codeCharta/util/nodeDecorator.ts
NearW:
```ts
	public static decorateMapWithBlacklist(map: CodeMapNode, blacklist: BlacklistItem[]) {
		const flattened = ignore()
		const excluded = ignore()

		for (const item of blacklist) {
			const path = CodeMapHelper.transformPath(item.path)
			if (item.type === BlacklistType.flatten) {
				flattened.add(path)
			} else {
				excluded.add(path)
			}
		}

		hierarchy(map)
			.descendants()
			.map(node => {
				const path = CodeMapHelper.transformPath(node.data.path)
				node.data.isFlattened = flattened.ignores(path)
				node.data.isExcluded = excluded.ignores(path)
			})
	}
```

It has been adressed and flattened is required to set the flattened attribute as well. Otherwise flattening a building would not work. :)

NearW: OH. I just realized there are actually two methods who used that ternary. You marked another one. my bad

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.flatten.component.html
```diff
@@ -17,7 +17,7 @@ <h3 class="header-text">Flattened</h3>
 			</div>
 		</md-list-item>
 
-		<md-list-item ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] track by $index" title="{{::item.path}}">
+		<md-list-item ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] track by $index+item.path" title="{{::item.path}}">
```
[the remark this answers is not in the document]
NearW: When using `track-by`, we need to track by something that is actually unique. AngularJS usually applies a hash to each object used in `ng-repeat`, but with `track-by` we have to deal with this by ourselves. `track by $index` is used, when you have a list with duplicates. Usually, angularjs would throw an error since duplicates result in the same hash, but with `$index` the hash is now the index. 

For some reason, if only the index is used here, we will have a blacklist in our view where each entry has the same path, even though they're different during runtime. They're just not displayed correctly. I don't know, why this is the case in this branch and not on master.

visualization/app/codeCharta/util/nodeDecorator.ts
NearW: I agree, that a for-loop is benificial since you can exit it earlier, but in this case, the number of excluded nodes can never exceed the number of nodes in the map. It's either all nodes are excluded or less.
I think that a for-loop is easier to read though and changed it. I also refactored, that we count the leaves as we traverse the tree. This should be more efficient than calling `hierarchy(map).leaves()`, since `leaves()` traverses the whole tree again.

```js
export default function() {
  var leaves = [];
  this.eachBefore(function(node) {
    if (!node.children) {
      leaves.push(node);
    }
  });
  return leaves;
}

```

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.service.spec.ts
```diff
@@ -50,20 +65,34 @@ describe("BlacklistService", () => {
...
 
 	describe("onStoreChanged", () => {
-		it("should notify all subscribers with the new blacklist", () => {
-			const item: BlacklistItem = { type: BlacklistType.exclude, path: "foo/bar" }
+		it("should notify all subscribers with the new blacklist and show the loading gif", () => {
+			NodeDecorator.doesExclusionResultInEmptyMap = jest.fn().mockReturnValue(false)
+
 			const action: BlacklistAction = { type: BlacklistActions.ADD_BLACKLIST_ITEM, payload: item }
-			storeService["store"].dispatch(action)
+			storeService.dispatch(action, true)
```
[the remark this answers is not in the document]
NearW: Adressed in a separate PR: https://github.com/MaibornWolff/codecharta/pull/1218

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.service.ts
NearW: But this would require a clone. Cloning the map has the same complexity as checking if each building is excluded.

BridgeAR: Just for me to understand: how come we'd have to clone the map in that case? I would expect to actually change the map and then to reverse that action if we realize it runs into an issue (no files left).

NearW: Well, cloning the map would be the easiest implementation. But now that I thought about this again:
While we update the map, we could add entries to a HashMap. Each entry represents the path of a node and it's previous state before we modified it.
If we run into a problem, we just traverse the tree again and check, if the path is in the HashMap and restore the previous state. We also remove the entry from the HashMap. Once the HashMap is empty, we're done.

I'll give this a shot

NearW: I tried my best and pushed a solution, but I encounterd some issues that are quite hard to explain without code as a context.

Listening to blacklist-changes and as a reaction trigger a blacklist-change was tricky, since we block all other subscribers. e.g:

Blacklist all nodes -> store changes -> blacklist-changes -> trigger decoration -> dispatch blacklist-action -> store changes -> blacklist-changes -> Now the second blacklist-change is applied to all subscribers first and then we can execute the blacklist-change we triggered the first time.

This resulted in wrong data being displayed and wrong metrics being selected. I fixed this, by listening to the store directly and check for blacklist-events:

Blacklist all nodes -> store changes -> check if it was because of blacklist -> trigger decoration -> dispatch blacklist-action -> store changes -> blacklist-changes -> blacklist-changes from the first store-change.

NearW:
**Current state of this PR:**
We changed a lot regarding the decoration process in the last weeks. This branch needs a complete rework and I think, I'll just start from scratch again. This PR still contains some valuable lines of code and should not be closed.
As in #1294, I'll implement an undo functionality for the blacklist-reducer.

BridgeAR:
@NearW is it possible to revive this PR without #1294?

PR #1160, opened 2020-08-13
Tech/move scripts to root

NearW:
# Move hooks to root package.json

## Description

The hooks provided by husky are currently located in the visualization. Therefore, the prettier-hook only works on the visualization folder. However, we have a lot of files that could benefit from formatting as well, e.g. .md files and .json files.

This PR moves the pre-commit and pre-push hook to the root-folder. This will require everyone to once run `npm i` on the root folder now. I also formatted the files.


package.json
```diff
@@ -1,12 +1,28 @@
...
+	"description": "MaibornWolff CodeCharta placeholder",
+	"license": "BSD-3-Clause",
+	"scripts": {
+		"format": "prettier --write \"./**/*\"",
+		"format:quick": "pretty-quick --staged",
+		"lint:visualization": "cd visualization && npm run lint:fix"
+	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "npm run lint:visualization && npm run format:quick",
```
BridgeAR, line 19: Running the formatter during the commit stage changes files locally without adding the changes to the commit. That might be confusing? I personally use [lint-staged](https://github.com/okonet/lint-staged) due to that reason. Also check https://prettier.io/docs/en/precommit.html#:~:text=You%20can%20use%20Prettier%20with,git%20add%20before%20you%20commit. for further options.

NearW: That's why we use `pretty-quick`. It allows us to add the formatted files again before commiting. 
As long as we run lint before commit, everything should be staged.

https://www.npmjs.com/package/pretty-quick#--staged-only-git

BridgeAR: Perfect!

Issue #1161, opened 2020-08-16
Right click on node does not hide open node options


BridgeAR:
# Bug

## Expected Behavior

GIVEN a right click on a node in the map and releasing the cursor (the NODE options are visible)

WHEN clicking with the right mouse button on parts of the map without releasing the cursor

THEN the originally open NODE options are hidden

## Actual Behavior

THEN the open NODE options are kept open

## Steps to Reproduce the Problem

1. Open a map
1. Click with the right mouse button on a node and release the button
1. Click on another node with the right mouse button
1. The node option is not hidden

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.53.0
- OS: All
- Browser: All

BridgeAR:
This is related to #1027

NearW:
> This is related to #1027

Absolutely. Seems like a small bug sneaked in. I guess, I just tried to hide the context menu by pressing the left mouse button.

PR #1169, opened 2020-08-24
fix/1161/solve-unclosing-node-options

ce-bo:
# Close open node options with a right click

Issue: #1161 

## Description
GIVEN opened node options menu.
WHEN clicking with the right mouse button on parts of the map without releasing the cursor
THEN the originally open NODE options SHOULD be hidden

This PR fixes the issue, that opened node options are kept open when right clicking.


ce-bo:
I am not sure on how to fire an appropriate event in a good way.
This PR adds a Subscriber for a generic "MAP RIGHT CLICK" event.
Any suggestions?

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
```diff
@@ -54,6 +55,7 @@ export class NodeContextMenuController
 	) {
 		MapColorsService.subscribe(this.$rootScope, this)
 		CodeMapMouseEventService.subscribeToBuildingRightClickedEvents(this.$rootScope, this)
+		CodeMapMouseEventService.subscribeToMapRightClickEvents(this.$rootScope, this)
```
NearW, line 58: at test, that CodeMapMouseEventService subscribes to your desired event

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
```diff
@@ -88,6 +90,10 @@ export class NodeContextMenuController
 		this.onShowNodeContextMenu(building.node.path, nodeType, x, y)
 	}
 
+	public onMapRightClick(x: number, y: number) {
+		this.hideNodeContextMenu()
+	}
```
NearW, line 95: add a tests, that checks if `this._viewModel.codeMapNode` is null afterwards

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
```diff
@@ -378,7 +378,7 @@ describe("codeMapMouseEventService", () => {
 
 				codeMapMouseEventService.onDocumentMouseUp(event)
 
-				expect($rootScope.$broadcast).not.toHaveBeenCalled()
+				expect($rootScope.$broadcast).not.toHaveBeenCalledWith("building-right-clicked")
```
NearW, line 381: You're inside the `onDocumentMouseUP` test. The event is triggered on mosue down.

NearW (changes requested):
fix-wise it LGTM.

Please add some more tests and especially, add an e2e test. We already have an e2e test checking if a node context menu is open, so it should be quite easy to add one, that right-clicks a node, right-clicks the map and checks, if the node context menu is closed.

Good work

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
ce-bo: Thanks @NearW. I just fixed failing tests for now. :)

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
ce-bo: The event passes click coordinates (x and y) that are not used in this particular Event Handler. The Linter will break the build.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -163,6 +168,13 @@ export class CodeMapMouseEventService
 	public onDocumentMouseDown(event) {
 		this.mouseOnLastClick = { x: event.clientX, y: event.clientY }
 		$(document.activeElement).blur()
+
+		if (event.button === ClickType.RightClick) {
+			this.$rootScope.$broadcast(CodeMapMouseEventService.MAP_RIGHT_CLICK_EVENT, {
```
JaiJoPa, line 173: I took a look too, and fix wise I like it too!
But your branch seems out-of-date. On our current main branch, there is a subscriber for the onHideNodeContextMenu. You could fire that one right here by calling `NodeContextMenuController.broadcastHideEvent($rootScope)`
That way you would not need the right-click-subscriber

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
NearW: Yes. Simply remove them.

NearW (approved):
LGTM

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.po.ts
```diff
@@ -10,7 +10,7 @@ export class NodeContextMenuPageObject {
 		await page.waitForSelector("#loading-gif-map", { visible: false })
 	}
 
-	public async isOpen(){
-		return !!await page.waitForSelector("node-context-menu-component", {visible: true})
+	public async isClosed() {
+		return await page.waitForSelector("node-context-menu-component", { visible: false })
```
BridgeAR, line 14: In an async function, it's best not to await the return value if it's not inside a `try` clause. Otherwise we'll delay the return value by another tick without any further benefit (every return value is automatically wrapped in a promise when returned from an async function).

NearW: What would the best way here then?

We could do something like this 

```ts
public async isClosed() {
	await page.waitForSelector("node-context-menu-component", { visible: false })
        return true
}
```

If the nodeContextMenu would still be open, we'd get a timeout.

NearW: or as shown [here](https://github.com/puppeteer/puppeteer/issues/1149#issuecomment-434302298)

```ts
public async isClosed() {
	try {
		await page.waitForSelector("node-context-menu-component", { visible: false })
		return true
	} catch (e) {
		return false
	}
}
```

BridgeAR: It's much simpler:

```suggestion
		return page.waitForSelector("node-context-menu-component", { visible: false })
```

The promise is resolved on the receiving end.

PR #1174, opened 2020-08-26
Update analysis dependencies

BridgeAR:
#### deps: update analysis dependencies

This updates to Kotlin 1.4.0 and also updates a couple plugins.
Everything is fixed in a separate commit.

All versions are from now on in the `gradle.properties` file instead
of being distributed between the different gradle files.

#### refactor: remove unused code and fix types

Some code paths could never be reached due to always evaluating to
e.g., a falsy value. Other code never used some of the arguments
passed through. This is fixed in this commit as well as formatting
the files along the new ktlint rules.

#### refactor: apply ktlint auto formatting

This formats our analysis files according to the current ktlint rules.

--------------------

Supersedes https://github.com/MaibornWolff/codecharta/pull/1107


analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/NodeJsonDeserializer.kt
```diff
@@ -45,16 +45,16 @@ internal class NodeJsonDeserializer : JsonDeserializer<Node> {
...
         val gson = GsonBuilder().create()
         return gson.fromJson<Map<String, Any>>(attributes, Map::class.java)
     }
 
-    private fun deserializeChildren(context: JsonDeserializationContext, jsonNode: JsonObject): List<Node> {
+    private fun deserializeChildren(jsonNode: JsonObject): List<Node> {
         val children = jsonNode.get("children") ?: return listOf()
 
-        return children.asJsonArray.mapTo(ArrayList()) { deserialize(it, Node::class.java, context) }
+        return children.asJsonArray.mapTo(ArrayList()) { deserialize(it, Node::class.java, null) }
```
NearW, line 58: What was the purpose of the context? Why is it never required?

NearW (approved):
Wow, those are some welcoming changes. LGTM codewise and Travis is working smoothly as well. Approved.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/NodeJsonDeserializer.kt
BridgeAR: It was never used before. I am not certain what it was meant to be used for. It is just required by the interface, that's why I kept the argument.

PR #1186, opened 2020-08-26
refactor: a few more minor dependency updates

BridgeAR:
A few more updates. These should be the last. There is an issue with dependabot right now that it's not able to detect the version anymore. Seems like it does not read `gradle.properties`.

### DIFF
--- a/analysis/build.gradle
+++ b/analysis/build.gradle
@@ -1,4 +1,40 @@
 buildscript {
+    ext {
+        // Dependency versions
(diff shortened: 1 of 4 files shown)


BridgeAR:
Moving the versions should fix the issue with dependabot

NearW (approved):
Ohh. I once wondered why there were so few dependency updates for kt-libraries. That explains a lot.

PR #1199, opened 2020-08-27
Feature/fixed folders

NearW:
# Fixate folders visualization 

## Description

This part implements the visualization of the fixated folders. I will work on the validation in another PR.

- Test: One snapshot test for this feature, I don't know how to test this in another way. There is just way too much data and coordinates
- Refactor: Make `getCCFile` static, so I can re-use it in a test
- Improve performance, by not traversing twice through the nodes

The really complicated part was to transform the coordinates correctly and apply the margin. The algorithm works like this:
- Build a squarified-tree-map for each fixated folder (each has it's own coordinate space, starting at 0 then) with the length and width of the fixated folder. Also apply the margin to it, so that it is rescalable.
- Now, we want to transform the coordinates from [0, INFINITY] to a space of [0, 100]. This space result in treemaps with the size of the actual folders defined in the cc.json. We also translate the folders according to the x and y positions.
- Build the root-folder of size 100x100
- Now, we have a 1:1 replica of the cc.json. Lastly, scale everything again, so that we match our usual map-size of 500 + some margin.

## Screenshots or gifs

image

image


visualization/app/codeCharta/util/treeMapGenerator.ts
```diff
@@ -1,44 +1,108 @@
...
+
+		const treeMap: TreemapLayout<CodeMapNode> = treemap<CodeMapNode>().size([width, height]).paddingOuter(padding).paddingInner(padding)
+
+		return { treeMap: treeMap(hierarchyNode.sum(node => this.calculateAreaValue(node, s))), height, width }
+	}
+
+	private static getEstimatedNodesPerSide(hierarchyNode: HierarchyNode<CodeMapNode>): number {
+		const descendants = hierarchyNode.descendants()
+		const blacklisted = descendants.filter(node => node.data.isExcluded || node.data.isFlattened).length
+		return 2 * Math.sqrt(descendants.length - blacklisted)
```
BridgeAR, line 103: How come it's multiplied by 2? :-) Maybe add a comment?

visualization/app/codeCharta/util/treeMapGenerator.ts
```diff
@@ -1,44 +1,108 @@
...
-		const treeMap: TreemapLayout<CodeMapNode> = treemap<CodeMapNode>()
-			.size([mapLength, mapLength])
-			.paddingOuter(padding)
-			.paddingInner(padding)
 
-		return treeMap(hierarchyNode.sum(node => this.calculateAreaValue(node, s)))
-	}
+		const nodes: Node[] = [TreeMapHelper.buildRootFolderForFixedFolders(map, heightScale, state, isDeltaState)]
+		const scale =
+			(state.treeMap.mapSize * 2 + this.getEstimatedNodesPerSide(hierarchyNode) * state.dynamicSettings.margin) / nodes[0].length
```
BridgeAR, line 39: I suggest to add comments to all of these calculations. Otherwise it's difficult to follow their logic.

BridgeAR:
This is looking pretty good!

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -28,6 +28,39 @@ export class TreeMapHelper {
...
+			height: 2,
+			length: 100,
+			depth: 0,
+			x0: 0,
+			z0: 0,
+			y0: 0,
+			isLeaf: false,
+			attributes: map.attributes,
+			edgeAttributes: map.edgeAttributes,
+			deltas: map.deltas,
```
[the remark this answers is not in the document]
NearW: They're not copied, they're just passed by reference.

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -28,6 +28,39 @@ export class TreeMapHelper {
...
+	public static buildRootFolderForFixedFolders(map: CodeMapNode, heightScale: number, state: State, isDeltaState: boolean): Node {
+		const flattened: boolean = this.isNodeToBeFlat(map, state)
+		const height = 2
+
+		return {
+			name: map.name,
+			id: 0,
+			width: 100,
+			height: 2,
+			length: 100,
```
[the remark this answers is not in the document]
NearW: While we use `width` and `height` in the `.cc.json`, we convert the term `height` to `length` since height is defined as the height of a building in a 3-dimensional space.
That's why the height and length are 100 here, which are the boundaries defined in the cc.json

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -28,6 +28,39 @@ export class TreeMapHelper {
...
+					? heightScale * map.deltas[state.dynamicSettings.heightMetric]
+					: 0,
+			visible: this.isVisible(map, false, state, flattened),
+			path: map.path,
+			link: map.link,
+			markingColor: CodeMapHelper.getMarkingColor(map, state.fileSettings.markedPackages),
+			flat: false,
+			color: this.getBuildingColor(map, state, isDeltaState, flattened),
+			incomingEdgePoint: this.getIncomingEdgePoint(length, height, length, new Vector3(0, 0, 0), state.treeMap.mapSize),
+			outgoingEdgePoint: this.getOutgoingEdgePoint(length, height, length, new Vector3(0, 0, 0), state.treeMap.mapSize)
```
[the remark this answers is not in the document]
NearW: length was never defined and I have no idea how this worked.
I changed it to `incomingEdgePoint: this.getIncomingEdgePoint(width, height, length, new Vector3(0, 0, 0), state.treeMap.mapSize)` and defined `height` and `length` as 100.

visualization/app/codeCharta/util/treeMapGenerator.ts
```diff
@@ -1,44 +1,108 @@
...
-		})
+
+		if (this.hasFixedFolders(map)) {
+			return this.buildSquarifiedTreeMapsForFixedFolders(map, s, heightScale, maxHeight, isDeltaState)
+		} else {
+			const squarifiedTreeMap = this.getSquarifiedTreeMap(map, s)
+			return squarifiedTreeMap.treeMap
+				.descendants()
+				.map(squarifiedNode => TreeMapHelper.buildNodeFrom(squarifiedNode, heightScale, maxHeight, s, isDeltaState))
+		}
```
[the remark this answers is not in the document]
NearW: Love this. Added the rule and fixed all issues. Increased files changed a lot, but changes are trivial and autofixed.

visualization/app/codeCharta/util/treeMapGenerator.ts
NearW: I wish I could explain this. This piece of code has been there since I started. All I know is, that it somehow estimates the number of nodes that are on each side, so I moved it to it's own function to explain at least this.

visualization/app/codeCharta/util/treeMapGenerator.ts
NearW:
```ts
private static scaleAndTranslateSquarifiedNode(
		squarifiedNode: HierarchyRectangularNode<CodeMapNode>,
		fixedFolder: CodeMapNode,
		squarified: SquarifiedTreeMap,
		scale: number
	) {
		// Transform coordinates from local folder space to world space (between 0 and 100).
		const scaleX = fixedFolder.fixedPosition.width / squarified.width
		const scaleY = fixedFolder.fixedPosition.height / squarified.height

		// Scales to usual map-size of 500 matching the three-scene-size
		squarifiedNode.x0 = (squarifiedNode.x0 * scaleX + fixedFolder.fixedPosition.left) * scale
		squarifiedNode.x1 = (squarifiedNode.x1 * scaleX + fixedFolder.fixedPosition.left) * scale
		squarifiedNode.y0 = (squarifiedNode.y0 * scaleY + fixedFolder.fixedPosition.top) * scale
		squarifiedNode.y1 = (squarifiedNode.y1 * scaleY + fixedFolder.fixedPosition.top) * scale
	}
```

I added the comments on how the scaling process works (which was the most difficult part) here.
Basically, we translate and scale each node of a treeMap from a [0: INF] range, to a [0, 100] range. After this, each folder and buildings matches the `.cc.json` exactly. However, we have to scale the map again to match our mapSize, which we use in the threeScene. This was previously defined (not by me) as `(state.treeMap.mapSize * 2 + this.getEstimatedNodesPerSide(hierarchyNode) * state.dynamicSettings.margin)`.

visualization/app/codeCharta/util/treeMapHelper.ts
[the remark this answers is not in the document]
BridgeAR: Could we rename the building height to e.g., `buildingHeigth`? That seems more intuitive than naming it `length`?

visualization/app/codeCharta/util/treeMapHelper.ts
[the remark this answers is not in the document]
BridgeAR: Sorry, I guess I could have worded this better :)

What I meant was if all properties of the `map` are currently shallowly copied to the new object such as `{ attributes: map.attributes }`. If that's the case, it would be possible to do: `{ foo: 5, bar: 'value', ...map }`. If not _all_ properties from `map` get copied, that's not the way to go but I thought I'll ask.

visualization/app/codeCharta/util/treeMapGenerator.ts
BridgeAR: I see, thanks!

visualization/app/codeCharta/ressources/fixed-folders/example.json
```diff
@@ -0,0 +1,386 @@
+{
+	"projectName": "example-project",
+	"apiVersion": "1.1",
```
BridgeAR, line 3:
```suggestion
	"apiVersion": "1.2",
```

visualization/app/codeCharta/util/codeMapHelper.ts
```diff
@@ -1,84 +1,94 @@
...
-		const matchingNode = hierarchy<CodeMapNode>(root)
-			.descendants()
-			.find(node => node.data.path === path && node.data.type === nodeType)
-		return matchingNode ? matchingNode.data : null
-	}
+function getAnyCodeMapNodeFromPath(path: string, root: CodeMapNode): CodeMapNode {
+	const matchingNode = hierarchy<CodeMapNode>(root)
+		.descendants()
+		.find(node => node.data.path === path)
+	return matchingNode ? matchingNode.data : null
```
BridgeAR, line 9: ❤️

visualization/app/codeCharta/util/fileHelper.ts
```diff
@@ -0,0 +1,45 @@
...
+	if (_.isEmpty(attributeTypes) || !attributeTypes || Array.isArray(attributeTypes.nodes) || Array.isArray(attributeTypes.edges)) {
+		return {
+			nodes: {},
+			edges: {}
+		}
+	}
+
+	return {
+		nodes: !attributeTypes.nodes ? {} : attributeTypes.nodes,
+		edges: !attributeTypes.edges ? {} : attributeTypes.edges
```
BridgeAR, line 34:
```suggestion
		nodes: attributeTypes.nodes ?? {},
		edges: attributeTypes.edges ?? {}
```

visualization/app/codeCharta/util/fileHelper.ts
```diff
@@ -0,0 +1,45 @@
...
+				blacklist: potentiallyUpdateBlacklistTypes(fileContent.blacklist || []),
+				markedPackages: fileContent.markedPackages || []
+			}
+		},
+		map: fileContent.nodes[0]
+	}
+}
+
+function getAttributeTypes(attributeTypes: AttributeTypes | OldAttributeTypes): AttributeTypes {
+	if (_.isEmpty(attributeTypes) || !attributeTypes || Array.isArray(attributeTypes.nodes) || Array.isArray(attributeTypes.edges)) {
```
BridgeAR, line 25:
```suggestion
	if (!attributeTypes || Array.isArray(attributeTypes.nodes) || Array.isArray(attributeTypes.edges)) {
```

visualization/app/codeCharta/util/treeMapGenerator.spec.ts
```diff
@@ -46,6 +50,12 @@ describe("treeMapGenerator", () => {
 
 			expect(nodes).toMatchSnapshot()
 		})
+
+		it("should build the a tree map with valid coordinates using the fixed folder structure", () => {
```
BridgeAR, line 54:
```suggestion
		it("should build the tree map with valid coordinates using the fixed folder structure", () => {
```

visualization/app/codeCharta/util/treeMapGenerator.ts
BridgeAR: Thanks. I can't yet follow the comments but I hope they'll already help others :)

Please feel free to resolve this comment as it's not new code.

visualization/app/codeCharta/util/treeMapHelper.spec.ts
```diff
@@ -118,210 +118,194 @@ describe("treeMapHelper", () => {
...
+				} as HierarchyRectangularNode<CodeMapNode>
+
+				state = STATE
+				state.treeMap.mapSize = 1
+				state.dynamicSettings.margin = 15
+			})
+
+			it("should not be a flat node when no visibleEdges", () => {
+				state.fileSettings.edges = []
+				expect(buildNode().flat).toBeFalsy()
```
BridgeAR, line 150: Nice!

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -4,171 +4,190 @@ import { Vector3 } from "three"
...
+	return new Vector3(vector.x - mapSize + width / 2, vector.y + height, vector.z - mapSize + 0.75 * length)
+}
 
-	private static nodeHasNoVisibleEdges(squaredNode: HierarchyRectangularNode<CodeMapNode>, s: State): boolean {
-		return (
-			squaredNode.data.edgeAttributes[s.dynamicSettings.edgeMetric] === undefined ||
-			s.fileSettings.edges.filter(edge => squaredNode.data.path === edge.fromNodeName || squaredNode.data.path === edge.toNodeName)
-				.length == 0
-		)
+function isNodeToBeFlat(codeMapNode: CodeMapNode, s: State): boolean {
```
BridgeAR, line 144: Could we rename this to e.g, `isNodeFlat` or `shouldNodeBeFlat`?

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -4,171 +4,190 @@ import { Vector3 } from "three"
...
+		length,
+		depth: 0,
+		x0: 0,
+		z0: 0,
+		y0: 0,
+		isLeaf: false,
+		attributes: map.attributes,
+		edgeAttributes: map.edgeAttributes,
+		deltas: map.deltas,
+		heightDelta: map.deltas?.[state.dynamicSettings.heightMetric] ? heightScale * map.deltas[state.dynamicSettings.heightMetric] : 0,
```
BridgeAR, line 50:
```suggestion
		heightDelta: (map.deltas?.[state.dynamicSettings.heightMetric] ?? 0) * heightScale,
```

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -4,171 +4,190 @@ import { Vector3 } from "three"
...
+		x0,
+		z0,
+		y0,
+		isLeaf: isNodeLeaf,
+		attributes: squaredNode.data.attributes,
+		edgeAttributes: squaredNode.data.edgeAttributes,
+		deltas: squaredNode.data.deltas,
+		heightDelta: squaredNode.data.deltas?.[s.dynamicSettings.heightMetric]
+			? heightScale * squaredNode.data.deltas[s.dynamicSettings.heightMetric]
+			: 0,
```
BridgeAR, line 96: Suggestion: this could be simplified to the comment above :-)

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -4,171 +4,190 @@ import { Vector3 } from "three"
...
-		)
+function isNodeToBeFlat(codeMapNode: CodeMapNode, s: State): boolean {
+	let flattened = false
+	if (s.appSettings.showOnlyBuildingsWithEdges && s.fileSettings.edges?.filter(edge => edge.visible).length > 0) {
+		flattened = nodeHasNoVisibleEdges(codeMapNode, s)
 	}
 
-	private static isNodeNonSearched(squaredNode: HierarchyRectangularNode<CodeMapNode>, s: State): boolean {
-		return !s.dynamicSettings.searchedNodePaths.has(squaredNode.data.path)
+	if (s.dynamicSettings.searchedNodePaths && s.dynamicSettings.searchPattern && s.dynamicSettings.searchPattern.length > 0) {
```
BridgeAR, line 150: Should this be an `else if` case? `flattened` is always defined inside of this if and as such, if the above if case would also apply, it would be redundant work?

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -4,171 +4,190 @@ import { Vector3 } from "three"
...
-				return s.appSettings.mapColors.flat
-			} else if (metricValue < s.dynamicSettings.colorRange.from) {
-				return s.appSettings.invertColorRange ? s.appSettings.mapColors.negative : mapColorPositive
-			} else if (metricValue > s.dynamicSettings.colorRange.to) {
-				return s.appSettings.invertColorRange ? mapColorPositive : s.appSettings.mapColors.negative
-			} else {
-				return s.appSettings.mapColors.neutral
-			}
-		}
+	flattened = codeMapNode.isFlattened || flattened
```
BridgeAR, line 154: Please move the `codeMapNode.isFlattened` condition to the top and return right away if it's truthy. Otherwise all other code is redundant.

```js
if (codeMapNode.isFlattened) {
  return true
}
...
```

visualization/app/codeCharta/util/fileHelper.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+			const result = getCCFile("fileName", fileContent)
+
+			expect(result.settings.fileSettings.blacklist).toEqual([{ path: "foo", type: BlacklistType.flatten }])
+		})
+
+		it("should ignore old attribute types", () => {
+			fileContent.attributeTypes = {
+				nodes: [{ mcc: AttributeTypeValue.absolute }],
+				edges: [{ pairingRate: AttributeTypeValue.relative }]
+			}
```
BridgeAR, line 33: What do you think about deprecating this and providing a warning that this is not supported in a future major version?

BridgeAR (commented):
This is shaping up nicely! I very much like the readability improvements. It feels like it's easier to follow the code with these changes.

visualization/app/codeCharta/util/treeMapHelper.ts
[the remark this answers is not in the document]
NearW: image

building height is already called `height`. I hope this illustration is clear

visualization/app/codeCharta/util/fileHelper.spec.ts
NearW: Sounds like a plan. Unfortunately we're lacking documentation of our api-version changes and I tried to sum it up as best as I could. I thought that we'd convert the old attribute types into our current schema and I was suprised that we just drop them.

I'll improve the validation and implement a warning.

BridgeAR: Yes, transforming them would likely even be better. I just don't know how long the old attributeTypes style was used.

visualization/app/codeCharta/ressources/fixed-folders/example.json
NearW: I'd like to keep the versioning and the validation in the other branch. Otherwise the current validation would throw an error, because the `.cc.json` is too new for the application.

visualization/app/codeCharta/util/treeMapHelper.ts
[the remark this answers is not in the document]
BridgeAR: Ah! That does indeed clarify the naming! Thanks!

visualization/app/codeCharta/util/fileHelper.spec.ts
NearW: The main problem was, that the old attribute types were never really set in the analysis. Also the interfaces in the analysis were different than the one in the visualization. It was a mess and we fixed it alltogether.

I know it's not a good practice, but I think that we should actually start fresh, since our userbase is quite small and I don't think that there are many `.cc.json` files out there with old attribute types. Even when, those can easily be fixed by hand.

It's up to you though. :)

BridgeAR: I'll give it another thought.

visualization/app/codeCharta/util/treeMapHelper.ts
NearW: Yes, we can return earlier. We just have to make sure, that we first check, if the nodePath doesn't match the current searchTerm. In this case, it's always flattened, no matter what settings we chose.

visualization/app/codeCharta/util/treeMapHelper.ts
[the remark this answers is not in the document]
NearW: I still can't follow yet. Let's talk about this tomorrow. :)

BridgeAR (approved):

Issue #1189, opened 2020-08-26
Performance regression due to lots of small events triggered


BridgeAR:
The current unreleased changes on `main` regressed the drawing performance by roughly 50%. The main reason is that now there are even more events than before. We fire way to many events and should reduce that a lot.

BridgeAR:
Closed as fixed by now.

PR #1200, opened 2020-08-28
Fix/improve merge performance

NearW:
# Fix bottleneck when merging edges on file selection change

## Description

- Use map to find already encountered edges / entries in constant time instead of O(n)
- Switching to multiple on maps with edges is 3 times faster now than before

## Screenshots or gifs

Before:

before

After:

after

## Open Questions

@BridgeAR Might this be related to #1189 ? Can you please specify the steps on how to reproduce your bottleneck in the issue?

### DIFF
--- a/visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.merger.ts
+++ b/visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.merger.ts
@@ -21,5 +21,5 @@ export function getMergedBlacklist(inputFiles: CCFile[], withUpdatedPath: boolea
 			}
 		}
 	}
-	return Array.from(blacklist.values())
+	return [...blacklist.values()]
 }
--- a/visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.merger.ts
+++ b/visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.merger.ts
@@ -2,7 +2,7 @@ import { getUpdatedBlacklistItemPath } from "../../../../util/nodePathHelper"
 import { BlacklistItem, CCFile } from "../../../../codeCharta.model"
 
 export function getMergedBlacklist(inputFiles: CCFile[], withUpdatedPath: boolean): BlacklistItem[] {
-	const blacklist: BlacklistItem[] = []
+	const blacklist: Map<string, BlacklistItem> = new Map()
 
 	if (inputFiles.length == 1) {
 		return inputFiles[0].settings.fileSettings.blacklist
@@ -17,13 +17,9 @@ export function getMergedBlacklist(inputFiles: CCFile[], withUpdatedPath: boolea
 						: oldBlacklistItem.path,
 					type: oldBlacklistItem.type
 				}
-				const equalBlacklistItems = blacklist.find(b => b.path == blacklistItem.path && b.type == blacklistItem.type)
-
-				if (!equalBlacklistItems) {
-					blacklist.push(blacklistItem)
-				}
+				blacklist.set(`${blacklistItem.path}|${blacklistItem.type}`, blacklistItem)
 			}
 		}
 	}
-	return blacklist
+	return Array.from(blacklist.values())
 }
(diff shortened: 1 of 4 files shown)


visualization/app/codeCharta/state/store/fileSettings/edges/edges.merger.ts
```diff
@@ -19,18 +19,18 @@ export function getMergedEdges(inputFiles: CCFile[], withUpdatedPath: boolean):
...
 					visible: oldEdge.visible
 				}
-				const equalEdgeItem = edges.find(e => e.fromNodeName == edge.fromNodeName && e.toNodeName == edge.toNodeName)
+				const equalEdgeItem = edges.get(`${edge.fromNodeName}|${edge.toNodeName}`)
 
-				if (equalEdgeItem) {
+				if (equalEdgeItem !== undefined) {
 					for (const key in edge.attributes) {
 						equalEdgeItem.attributes[key] = edge.attributes[key]
+						edges.set(`${edge.fromNodeName}|${edge.toNodeName}`, equalEdgeItem)
```
BridgeAR, line 27:
```suggestion
```

This call is just going to replace itself. `equalEdgeItem` is already a reference and as such, you'll change the actual entry inside the map implicitly.

And even if that would not be the case, it would actually be better to place the set after the for loop so that it's just called once.

As a change independent comment: using `for ... in` should ideally not be used since it'll also iterate over the prototype. Instead, it's better to use e.g., `for (const key of Object.keys(edge.attributes))` or `for (const [key, value] of Object.entries(edge.attributes))`.
And is it actually intentional to override existing keys? I just wonder if that's always the best behavior.

visualization/app/codeCharta/state/store/fileSettings/blacklist/blacklist.merger.ts
```diff
@@ -17,13 +17,9 @@ export function getMergedBlacklist(inputFiles: CCFile[], withUpdatedPath: boolea
...
-
-				if (!equalBlacklistItems) {
-					blacklist.push(blacklistItem)
-				}
+				blacklist.set(`${blacklistItem.path}|${blacklistItem.type}`, blacklistItem)
 			}
 		}
 	}
-	return blacklist
+	return Array.from(blacklist.values())
```
BridgeAR, line 24: Nit: I personally prefer the shorter spread notation:

```suggestion
	return [...blacklist.values()]
```

BridgeAR (approved):
LGTM, this is a great improvement! I just wonder if it would not make sense to always return a map directly for e.g., `BlacklistItem`? That way there's no need for any transformations besides maybe one while loading a cc.json.

BridgeAR:
> Might this be related to #1189

No, it's not related. To reproduce #1189, benchmark the loading period of a big map.

visualization/app/codeCharta/state/store/fileSettings/edges/edges.merger.ts
NearW: I couldn't tell whether or not overriding the values of the keys here was intentional or not. There exists an test to ensure this is the case, so I thought it was important.
The question is: When can two edges have the same source and destination?
- In a single file, this shouldn't be the case.
- In multiple mode, it's impossible, because the paths would be different. We use the fileName as the name of the folder we put as a root for each project and fileNames are unique in codecharta. 
- In delta mode, there might be a lot of these edges, but edges don't make sense in delta mode actually. We can't display a delta of those. Edges should be disabled there in my opinion.

NearW:
> I just wonder if it would not make sense to always return a map directly for e.g., `BlacklistItem`? That way there's no need for any transformations besides maybe one while loading a cc.json.

I don't think it doesn't really matter that much. In order to display the blacklist for angular, we need to transform it to an array to render it using `ng-repeat`. So we'll have a transformation to an in our components for sure, when the blacklist changes. We also can't use those beautiful array functions like `filter` or `map` anymore. And the redux store is easier to maintain using arrays, see [immutable update patterns](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns). I don't actually know how to properly handle Maps or Sets in the state.

Issue #1201, opened 2020-08-31
EdgeMetric always available in Scenarios


JaiJoPa:
# Bug

## Expected Behavior

<GIVEN> 

WHEN EdgeMetric is saved inside a Scenario

THEN the Scenario indicates, that the EdgeMetric of the Sceanrio could be applied, eventhough it can not.

## Actual Behavior

WHEN EdgeMetric is saved inside a Scenario

THEN the EdgeMetric icon should be gray when it cannot be applied to the current map.

## Screenshots

![NoEdge](https://user-images.githubusercontent.com/50145538/91711484-41c81c80-eb86-11ea-8eff-e3bfbd545ccc.gif)

## Steps to Reproduce the Problem
1. Open a map with edgeMetrics
2. Select an EdgeMetric
3. Save it inside a Scenario
4. Open a map where the saved EdgeMetric is not available
5. Open ScnearioDropDown Menu, the Scenario should indicate the EdgeMetric is appliable

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version:
- OS: All
- Browser: All

PR #1217, opened 2020-09-03
Bug/1201/edge metric always available on scenarioDropDown

JaiJoPa:
#EdgeMetric always available on scenarioDropDown

fixes: #1201 

## Description

- Scenario with saved edgeMetric will only be appliable if the ScenarioEdgeMetric is existing in edgeMetricData
- in isScenarioAppliable, there is condition added, which checks, if the ScenarioEdgeMetric is existing in the current edgeMetricData
- getScenarioItems now gets metricData instead of just nodeMetricData, because we need edgeMetricData too for the isAppliable calculation.

## Screenshots or gifs
NoEdge


visualization/app/codeCharta/util/scenarioHelper.ts
```diff
@@ -49,14 +49,17 @@ export class ScenarioHelper {
 		return scenarioItems
 	}
 
-	private static isScenarioAppliable(scenario: RecursivePartial<Scenario>, metricData: NodeMetricData[]) {
```
NearW, line 52:
```suggestion
    private static isScenarioAppliable(scenario: RecursivePartial<Scenario>, metricData: MetricData) {
        const expected = Object.keys(scenario).filter(x => x === "area" || x === "height" || x === "color" || x === "edge").length
        let existingNodeMetricData = 0
        let existingEdgeMetricData = 0

        for (const nodeMetricData of metricData.nodeMetricData) {
            if(nodeMetricData.name === scenario.area?.areaMetric || nodeMetricData.name === scenario.color?.colorMetric
                || nodeMetricData.name === scenario.height?.heightMetric) {
                existingNodeMetricData++
            }
        }

        for (const edgeMetricData of metricData.edgeMetricData) {
            if(edgeMetricData.name === scenario.edge?.edgeMetric) {
                existingEdgeMetricData++
            }
        }

        return existingNodeMetricData + existingEdgeMetricData === expected
    }
```

This is slightly more performant. This only traverse each metricData once instead of traversing the nodeMetricData three times.
Better apply this in the IDE to ensure formatting the code.

NearW (approved):
LGTM.

Please prefix your future bug-fix branches with `fix` instead of `bug`.
We updated our CONTRIBUTING-file. :)

visualization/app/codeCharta/util/scenarioHelper.ts
BridgeAR: I would actually use a different approach:

```js
const { area, color, height } = scenario

if (area || color || height) {
  const nodeMetricSet = new Set(metricData.nodeMetricData.map((data) => data.name)
  
  if (area && !nodeMetricSet.has(area.areaMetric)) {
    return false
  }
  ...
}
```

I'd keep the suggested edge metric detection as it's the fasted implementation (we do not always have to go through all of them).

JaiJoPa: ``  if(nodeMetricData.name === scenario.area?.areaMetric || nodeMetricData.name === scenario.color?.colorMetric
                || nodeMetricData.name === scenario.height?.heightMetric) {
                existingNodeMetricData++
            }``

I like the idea of iterating just once through the nodeMetricData, but we have to find another solution for the if statement. The problem is, that when areaMetric and colorMetricfor exampe have the same metric saved, it will just increment +1 instead of +2.
That way it will return false as the expected value is 3.

BridgeAR (approved):

visualization/app/codeCharta/util/scenarioHelper.ts
```diff
@@ -49,14 +49,24 @@ export class ScenarioHelper {
...
+			}
+			if (height && !nodeMetricSet.has(height.heightMetric)) {
+				return false
+			}
 		}
-		if (scenario.height && !metricData.find(x => x.name === scenario.height.heightMetric)) {
+
+		if (edge && !metricData.edgeMetricData.find(x => x.name === edge.edgeMetric)) {
 			return false
 		}
```
JaiJoPa, line 72: Just out of curiosity:

We could remove the last return true and include it to the edge query. It would look like this

`return !(edge && !metricData.edgeMetricData.find(x => x.name === edge.edgeMetric))`

but I don't know if the readability got worse. How would you guys handle that?

NearW: I like the shorter notation. As long as we don't have any other cases, I'd try to keep the code as short and readable as possible.

visualization/app/codeCharta/util/scenarioHelper.ts
```diff
@@ -49,14 +49,24 @@ export class ScenarioHelper {
...
+
+			if (area && !nodeMetricSet.has(area.areaMetric)) {
+				return false
+			}
+			if (color && !nodeMetricSet.has(color.colorMetric)) {
+				return false
+			}
+			if (height && !nodeMetricSet.has(height.heightMetric)) {
+				return false
+			}
```
BridgeAR, line 66:
```suggestion
			if (
				area && !nodeMetricSet.has(area.areaMetric) ||
				color && !nodeMetricSet.has(color.colorMetric) ||
				height && !nodeMetricSet.has(height.heightMetric)
			) {
				return false
			}
```

If we prefer short notations.

JaiJoPa: Oh I like that short notation of the if statements

visualization/app/codeCharta/util/scenarioHelper.spec.ts
```diff
@@ -142,7 +113,26 @@ describe("scenarioHelper", () => {
...
-			const result = ScenarioHelper.getScenarioItems([{ name: "mcc", maxValue: 56 }])
+			const result = ScenarioHelper.getScenarioItems({
+				nodeMetricData: [{ name: "mcc", maxValue: 56 }],
+				edgeMetricData: [{ name: "None", maxValue: 0 }]
+			})
+
+			expect(result).toEqual(expected)
+		})
+
+		it("should not be appliable when EdgeMetric is not existing", () => {
```
BridgeAR, line 124:
```suggestion
		it("should not be applicable when EdgeMetric is not existing", () => {
```

visualization/app/codeCharta/util/scenarioHelper.ts
```diff
@@ -49,18 +49,22 @@ export class ScenarioHelper {
...
-	private static isScenarioAppliable(scenario: RecursivePartial<Scenario>, metricData: NodeMetricData[]) {
-		if (scenario.area && !metricData.find(x => x.name === scenario.area.areaMetric)) {
-			return false
-		}
-		if (scenario.color && !metricData.find(x => x.name === scenario.color.colorMetric)) {
-			return false
-		}
-		if (scenario.height && !metricData.find(x => x.name === scenario.height.heightMetric)) {
-			return false
+	private static isScenarioAppliable(scenario: RecursivePartial<Scenario>, metricData: MetricData) {
```
BridgeAR, line 52:
```suggestion
	private static isScenarioApplicable(scenario: RecursivePartial<Scenario>, metricData: MetricData) {
```

Seems like this needs some other updates as well :)

Issue #1024, opened 2020-07-04
Cut loading time into half


BridgeAR:
# Bug / Feature

## Expected Behavior

__GIVEN__: A big cc.json

__WHEN__: Open the cc.json in the visualizer

__THEN__: It loads the file quickly

## Actual Behavior

It takes a significant amount of time to initialize everything. Most time is spend in few functions, e.g.,
`decorateParentNodesWithAggregatedAttributes`, `decorateMapWithBlacklist` and `buildMeshFromIntermediateVertexData`. A lot of time is spend on cloning as well and it's questionable if the cloning is indeed always necessary. It is likely, that the data structure could be cloned once and that it's all that is needed instead of multiple times. Some functions only have smaller optimization potential (~1%) while others provide significant potential e.g., ~10% of the total current time.

Reducing the total loading time by more than 1/3 is likely going to become more challenging, so try to focus on obvious gains. Often, the same structure is iterated upon multiple times doing one step at a time instead of iterating once and doing multiple steps on each iteration.

To have an actual starting point: use the current version and profile the performance with a big map (e.g., [Apache OpenOffice](https://maibornwolff.github.io/codecharta/showcase/aoo/)). That way you have a better comparison. It takes almost seven seconds to load the file on my device. The goal is to reduce that to 3.5 seconds or below. Split the work into small PRs that improve the performance of small areas per PR that all reference this issue and that outline how much gain the change provides.

This task has some overlap with #1025 and some gains will be noticeable for both, the loading and filtering time.

## Steps to Reproduce the Problem

1. Open the web visualizer
2. Start the browsers profiler
3. Load a big cc file.

## Specifications

- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.49.1
- OS: All
- Browser: Chromium

BridgeAR:
Note: there is a very big potential by refactoring the internal code to have an initial cc.json and one that is "worked" upon. Meaning the latter is changed depending on the current settings. That way there's only need to copy the cc.json once up front and never afterwards. That's currently not the case and it is likely that we copy the cc.json frequently.

BridgeAR:
Current main bottlenecks:

1. Map decoration is triggered six times during loading. This is likely the biggest bottleneck.
1. Decoration is done for all metrics up front instead of only decorating the visible metrics. A lot of work is not necessary.
[...]

PR #1246, opened 2020-09-12
Tech/improve performance

BridgeAR:
# Improve loading performance

Issue: #1024

## Description

This reduces the decoration overhead significantly. It regressed recently and this brings it back to a better level. There is still a lot of room for improvement but it's already a lot faster than before.


visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
-		isDeltaState: boolean,
-		attributeTypes: AttributeTypes
-	) {
-		metricData.forEach(metric => {
-			if (node.data.children && node.data.children.length > 0) {
-				node.data.attributes[metric.name] = this.aggregateLeafMetric(
-					leaves.map(x => x.data.attributes),
-					metric.name,
+			if (!hasChildren(data)) {
+				return
```
NearW, line 120: Why do we return here? Shouldn't we `continue`?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+			}
+
+			for (const { name } of edgeMetricData) {
+				data.edgeAttributes[name] = { incoming: 0, outgoing: 0 }
+			}
```
NearW, line 125: This seems out of place. I'd expect these to be set on the `decorateMap` method already, since we add empty objects there.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+				if (data.isExcluded) {
+					continue;
+				}
+				leafAttributes.push(data.attributes)
+				if (data.deltas !== undefined) {
+					leafDeltas.push(data.deltas)
+				}
+				if (!data.edgeAttributes) {
+					continue;
+				}
```
NearW, line 154: Aren't these set previously? Do we even encounter a case where these are not set?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+				// Optimize for the common case
+				for (const [name, value] of Object.entries(data.edgeAttributes)) {
+					if (!value) {
+						continue;
+					}
```
NearW, line 159: Same here: We set the `edgeAttributes` for each node to `data.edgeAttributes[name] = { incoming: 0, outgoing: 0 }`.
Vale should always be defined.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+					if (attributeTypes.edges[name] === AttributeTypeValue.relative) {
+						// Use the median after collecting all entries.
+						if (edgeMetrics === undefined) {
+							edgeMetrics = {}
+						}
```
NearW, line 164: Why don't we initiliaze `edgeMetrics = {}` right from the start?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
-	private static decorateNodeWithChildrenSumEdgeMetrics(
-		leaves: HierarchyNode<CodeMapNode>[],
-		node: HierarchyNode<CodeMapNode>,
-		edgeMetricData: NodeMetricData[],
-		attributeTypes: AttributeTypes
-	) {
-		edgeMetricData.forEach(edgeMetric => {
-			if (node.data.children && node.data.children.length > 0) {
-				node.data.edgeAttributes[edgeMetric.name] = this.aggregateLeafEdgeMetric(leaves, edgeMetric.name, attributeTypes)
+			if (edgeMetrics) {
```
NearW, line 194: At this point edgeMetrics is always defined, isn't it?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+			// finding the median is linear instead of O(n log n). It is likely
+			// possible to optimize the latter further but that is probably too much
+			// work.
+			for (const { data } of node.leaves()) {
+				if (data.isExcluded) {
+					continue;
+				}
+				leafAttributes.push(data.attributes)
+				if (data.deltas !== undefined) {
+					leafDeltas.push(data.deltas)
```
NearW, line 150: When is the delta-property set to the node?

NearW (commented):
Found some pieces of code that I couldn't quite follow yet.
I'd find it helpful if we split this code up into smaller functions again.

E.g.
```ts
			for (const { name } of edgeMetricData) {
				data.edgeAttributes[name] = { incoming: 0, outgoing: 0 }
			}
```
could me moved to a function called `setDefaultEdgeAttributes()`.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
-			return (
-				current &&
-				current.children &&
-				current.children.length === 1 &&
-				current.children[0].children &&
-				current.children[0].children.length > 0
-			)
+		const defaultAttributes = { [NodeMetricDataService.UNARY_METRIC]: 1 }
+		const defaultLeafAttributes = { ...defaultAttributes }
+		if (metricData?.length) {
```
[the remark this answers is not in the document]
BridgeAR: They should definitely be defined but I kept multiple safeguards in place as they were before. I am fine to try to remove a couple of these. I did in fact already remove a few.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: No, it's not a for loop but a `each` function that calls the callback for each node (similar to e.g., `forEach()`).

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: It's to differentiate between relative edges and "regular" edges. The relative ones are defined lazily as they require extra work. The "regular" ones are directly counted together.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: It's the lazy relative edge metric detection. If the node does not have those, they won't be defined.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
-				current.deltas = child.deltas
-				rec(current)
-			} else if (current && current.children && current.children.length > 1) {
-				for (let i = 0; i < current.children.length; i++) {
-					rec(current.children[i])
+		for (const { data } of root.descendants()) {
+			data.id = id
+			id++
+			if (data.attributes) {
+				data.attributes[NodeMetricDataService.UNARY_METRIC] = 1
```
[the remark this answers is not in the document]
BridgeAR: How come? The logic is the same as before and tests will fail by removing this.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
+				if (metricData?.length && !hasChildren(data)) {
+					for (const metric of metricData) {
```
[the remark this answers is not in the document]
BridgeAR: I am not certain I am able to follow. The default attributes are defined a single time on one object to place that object multiple times where it's missing. This case is about already existing attributes where just a few default metrics are missing. It is currently written in the way that is the least amount of work for the CPU as far as I see it?

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: Good point. These are indeed always defined.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: That would require to pass through the `edgeMetricData` to `decorateMap` and it would for now also increase the amount of work to determine the number correct metrics. There is potential to optimize this further as soon as `decoarateMap` and `decorateParentNodesWithAggregatedAttributes` is combined into a single function. I would not want to do that right now though.

visualization/app/codeCharta/util/nodeDecorator.ts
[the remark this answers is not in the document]
NearW: That's probably because those are unit test and the provided metric data for the test doesn't include the `unary` metric. When we set new metrics, we add the unary metric though. I guess we might have to update the test data, but I gotta check again.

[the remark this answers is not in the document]
BridgeAR: Maybe, I would still rather not change anything that requires changes in until tests in this PR.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: Until tests are going to fail if I remove that. I am going to keep this code for now.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: Our unit tests contain entries that have undefined values. Thus, this safeguard should be kept for now.

visualization/app/codeCharta/util/nodeDecorator.ts
BridgeAR: This is only to gather the the delta values. They are set in the `deltaGenerator.ts` file.

BridgeAR:
> Found some pieces of code that I couldn't quite follow yet.
> I'd find it helpful if we split this code up into smaller functions again.

I can do that if you wish, while I personally would just add a few comments instead until the function is fully optimized. Until then, it's difficult to see the interactions of the code if all parts are in different functions.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -16,220 +15,210 @@ import { CodeMapHelper } from "./codeMapHelper"
...
-	private static decorateMapWithCompactMiddlePackages(map: CodeMapNode) {
-		const isEmptyMiddlePackage = current => {
-			return (
-				current &&
-				current.children &&
-				current.children.length === 1 &&
-				current.children[0].children &&
-				current.children[0].children.length > 0
-			)
+		const defaultAttributes = { [NodeMetricDataService.UNARY_METRIC]: 1 }
```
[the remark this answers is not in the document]
BridgeAR: Maybe, but this is required for the tests. I would rather keep this for now. I do remember that there was a recent change where this was removed and we had a regression as well.

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -42,20 +42,19 @@ export class CodeChartaController {
...
+			const data = await this.urlUtils.getFileDataFromQueryParam()
+			if (data.length > 0) {
+				this.tryLoadingFiles(data)
+				this.setRenderStateFromUrl()
+			} else {
 				this.tryLoadingSampleFiles()
-			})
+			}
+		} catch (error) {
+			console.error(error)
```
BridgeAR, line 55: We should probably improve the error handling here?

visualization/app/codeCharta/codeCharta.model.ts
```diff
@@ -253,7 +253,7 @@ export interface Node {
 	z0: number
 	y0: number
 	isLeaf: boolean
-	deltas: KeyValuePair
+	deltas?: KeyValuePair
```
BridgeAR, line 256: This is an important change as it's not guaranteed to have deltas.

visualization/app/codeCharta/e2e/url.e2e.ts
```diff
@@ -1,9 +1,30 @@
 import { CC_URL, goto } from "../../puppeteer.helper"
 import { DialogErrorPageObject } from "../ui/dialog/dialog.error.po"
 import { FilePanelPageObject } from "../ui/filePanel/filePanel.po"
-import sample1 from "../assets/sample3.cc.json"
+import sample1 from "../assets/sample1.cc.json"
```
BridgeAR, line 4: Test bug fix

visualization/app/codeCharta/state/store/dynamicSettings/margin/margin.reset.ts
```diff
@@ -1,24 +1,25 @@
...
+import { hierarchy } from "d3-hierarchy"
 
 const MIN_MARGIN = 15
 const MAX_MARGIN = 100
 const MARGIN_FACTOR = 4
 
-export function getResetMargin(dynamicMargin: boolean, map: CodeMapNode, areaMetric: string): number {
-	if (dynamicMargin && map) {
+export function getResetMargin(dynamicMargin: boolean, map: CodeMapNode, areaMetric: string) {
+	if (dynamicMargin) {
```
BridgeAR: Check this one again. It might need some more work.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
```diff
@@ -505,7 +505,7 @@ describe("codeMapMouseEventService", () => {
 		})
 
 		it("should add property node", () => {
-			codeMapBuilding.setNode(undefined)
+			codeMapBuilding.setNode(CODE_MAP_BUILDING.node)
```
BridgeAR, line 508: Test bug fix

visualization/app/codeCharta/ui/codeMap/codeMap.preRender.service.ts
```diff
@@ -130,28 +130,34 @@ export class CodeMapPreRenderService implements StoreSubscriber, MetricDataSubsc
...
-			const comparisonFile = visibleFileStates.find(x => x.selectedAs == FileSelectionState.Comparison).file
-			return DeltaGenerator.getDeltaFile(referenceFile, comparisonFile)
+	private getDeltaFile(visibleFileStates: FileState[]) {
+		if (visibleFileStates.length === 2) {
+			let [reference, comparison] = visibleFileStates
+			if (reference.selectedAs !== FileSelectionState.Reference) {
+				const temporary = reference
+				comparison = reference
+				reference = temporary
+			}
```
BridgeAR, line 155: Simplified lookup

visualization/app/codeCharta/ui/codeMap/rendering/boxGeometryGenerationHelper.ts
```diff
@@ -47,46 +47,44 @@ export class BoxGeometryGenerationHelper {
...
-		uvs[sides.back * verticesPerSide + vertexLocation.bottomRight] = new Vector2(0.0, 1.0)
+		const back = sides.back * verticesPerSide
+		positions[back] = positions[bottom + vertexLocation.topRight]
+		positions[back + vertexLocation.topLeft] = positions[left + vertexLocation.bottomRight]
+		positions[back + vertexLocation.topRight] = positions[left + vertexLocation.topRight]
+		positions[back + vertexLocation.bottomRight] = new Vector3(maxPos.x, maxPos.y, maxPos.z)
+		uvs[back] = uvs[bottom + vertexLocation.bottomRight]
+		uvs[back + vertexLocation.topLeft] = uvs[bottom + vertexLocation.topRight]
+		uvs[back + vertexLocation.topRight] = uvs[bottom + vertexLocation.topLeft]
+		uvs[back + vertexLocation.bottomRight] = uvs[bottom]
```
BridgeAR, line 82: Reuse identical vectors

visualization/app/codeCharta/ui/codeMap/rendering/codeMapMesh.ts
```diff
@@ -44,11 +44,11 @@ export class CodeMapMesh {
 		}
 	}
 
-	public getThreeMesh(): Mesh {
+	public getThreeMesh() {
 		return this.threeMesh
 	}
 
-	public selectBuilding(building: CodeMapBuilding, selected: CodeMapBuilding, color: string) {
+	public selectBuilding(building: CodeMapBuilding, color: string) {
```
BridgeAR, line 51: Removed unused variable

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -1,28 +1,17 @@
...
 	private files: NameDataPair[] = []
 
 	/* @ngInject */
-	constructor(
-		private $scope,
-		private dialogService: DialogService,
-		private codeChartaService: CodeChartaService,
-		private storeService: StoreService
-	) {}
+	constructor(private $scope, private codeChartaService: CodeChartaService, private storeService: StoreService) {}
```
BridgeAR, line 14: Removed unused argument

visualization/app/codeCharta/ui/fileExtensionBar/fileExtensionBar.component.ts
```diff
@@ -34,16 +34,13 @@ export class FileExtensionBarController implements CodeMapPreRenderServiceSubscr
...
+			this._viewModel.distribution.filter(metric => metric.fileExtension !== "other").map(metric => metric.fileExtension)
+		)
 
 		buildings.forEach(building => {
 			const buildingExtension = FileExtensionCalculator.estimateFileExtension(building.node.name)
-			if (
-				buildingExtension === hoveredExtension ||
-				(hoveredExtension === "other" && !visibleFileExtensions.includes(buildingExtension))
-			) {
+			if (buildingExtension === hoveredExtension || (hoveredExtension === "other" && !visibleFileExtensions.has(buildingExtension))) {
```
BridgeAR, line 43: Faster lookup

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.ts
```diff
@@ -149,11 +149,11 @@ export class LegendPanelController
 	}
 
 	private insertMPIntoPackageList(mp: MarkedPackage, colorPixel: string) {
-		this._viewModel.packageLists.filter(packageList => packageList.colorPixel == colorPixel)[0].markedPackages.push(mp)
+		this._viewModel.packageLists.find(packageList => packageList.colorPixel === colorPixel).markedPackages.push(mp)
```
BridgeAR, line 152: Faster lookup

visualization/app/codeCharta/util/codeMapHelper.ts
```diff
@@ -2,38 +2,39 @@ import { hierarchy } from "d3-hierarchy"
...
+function transformPath(toTransform: string) {
+	let removeNumberOfCharactersFromStart = 2
 
-	if (toTransform.startsWith("./")) {
-		removeNumberOfCharactersFromStart = 2
-	} else if (toTransform.startsWith("/")) {
+	if (toTransform.startsWith("/")) {
 		removeNumberOfCharactersFromStart = 1
+	} else if (!toTransform.startsWith("./")) {
+		return toTransform
```
BridgeAR, line 31: Prevents slicing

visualization/app/codeCharta/util/deltaGenerator.ts
```diff
@@ -16,23 +15,19 @@ export class DeltaGenerator {
 		return this.getNewCCFileWithDeltas(map, fileMeta)
 	}
 
-	private static getCodeMapNodesAsHashMap(rootNode: CodeMapNode): Map<string, CodeMapNode> {
+	private static getCodeMapNodesAsHashMap(rootNode: CodeMapNode) {
 		const hashMap = new Map<string, CodeMapNode>()
 
 		hierarchy(rootNode)
 			.descendants()
-			.map(x => x.data)
```
BridgeAR, line 24: Removed extra mapping

visualization/app/codeCharta/util/nodeDecorator.spec.ts
```diff
@@ -44,15 +36,16 @@ describe("nodeDecorator", () => {
...
+		let isUnique = true
 		map.each(node => {
 			if (ids.has(node.id)) {
-				return false
+				isUnique = false
 			}
 			ids.add(node.id)
 		})
-		return true
+		return isUnique
```
BridgeAR: Fixed uniqueness check

visualization/app/codeCharta/util/urlExtractor.spec.ts
```diff
@@ -96,25 +96,25 @@ describe("urlExtractor", () => {
...
 
-		it("should reject if statuscode is not 200", async () => {
+		it("should reject if statuscode is not 2xx", async () => {
 			$http.get = jest.fn().mockImplementation(async () => {
-				return { data: "some data", status: 201 }
+				return { data: "some data", status: 301 }
 			})
 
-			return expect(urlExtractor.getFileDataFromFile("test.json")).rejects.toEqual(undefined)
+			return expect(urlExtractor.getFileDataFromFile("test.json")).rejects.toEqual(new Error(`Could not load file "test.json"`))
```
BridgeAR, line 117: Improved errors

visualization/app/codeCharta/util/urlExtractor.spec.ts
```diff
@@ -96,25 +96,25 @@ describe("urlExtractor", () => {
...
+		it("should resolve data and return an object with content and fileName", async () => {
 			const expected = { content: "some data", fileName: "test.json" }
 			return expect(urlExtractor.getFileDataFromFile("test.json")).resolves.toEqual(expected)
 		})
 
-		it("should reject if statuscode is not 200", async () => {
+		it("should reject if statuscode is not 2xx", async () => {
 			$http.get = jest.fn().mockImplementation(async () => {
-				return { data: "some data", status: 201 }
+				return { data: "some data", status: 301 }
```
BridgeAR, line 114: Improved status check

visualization/app/codeCharta/codeCharta.component.ts
NearW: I assume, that his piece of code is quite old and from a time, where we had no proper validation.
I'm not sure if this catch-block is ever reached. We catch the errros already during validation.

visualization/app/codeCharta/codeCharta.model.ts
NearW: Exactly. As far as I understood the code, the delta property is only calculated and decorated, when we're in the delta mode. In this case, we check the two maps `reference` and `comparison`, determine the difference between nodes and add that as the property `delta`. Am I correct?

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
```diff
@@ -6,6 +6,7 @@ import { setSearchPanelMode } from "./state/store/appSettings/searchPanelMode/se
 export class CodeChartaMouseEventService {
 	constructor(private storeService: StoreService) {}
 
+	// eslint-disable-next-line @typescript-eslint/ban-types
```
NearW, line 9: Would be great to have a TODO here as well. I'm afraid we miss this in the future otherwise.

visualization/app/codeCharta/model/files/files.helper.ts
```diff
@@ -1,35 +1,34 @@
-import { CCFile } from "../../codeCharta.model"
 import { FileSelectionState, FileState } from "./files"
 
-export function getCCFiles(fileStates: FileState[]): CCFile[] {
+export function getCCFiles(fileStates: FileState[]) {
```
NearW, line 3: Why do we remove the return-types here?

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -32,19 +21,23 @@ export class FileChooserController {
...
+				}
 
-				reader.onloadstart = () => {
+				reader.addEventListener("loadstart", () => {
 					this.storeService.dispatch(setIsLoadingFile(true))
-				}
+				})
 
-				reader.onload = event => {
+				reader.addEventListener("load", event => {
```
NearW, line 34: Why don't we use the existing event listeners? Creating new ones and NOT removing them is a bad idea.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.ts
```diff
@@ -47,12 +47,7 @@ export class MapTreeViewLevelController implements BuildingHoveredSubscriber, Bu
 	}
 
 	public onBuildingHovered(hoveredBuilding: CodeMapBuilding) {
-		this._viewModel.isHoveredInCodeMap = !!(
-			this.node &&
-			this.node.path &&
-			hoveredBuilding.node &&
-			hoveredBuilding.node.path === this.node.path
-		)
+		this._viewModel.isHoveredInCodeMap = Boolean(this.node?.path && hoveredBuilding.node?.path === this.node.path)
```
NearW, line 50: Is the cast even necessary? This evaluates to a boolean anyway

visualization/app/codeCharta/util/codeMapHelper.ts
NearW: Before:
`transformPath("./root/path")` => `root/path`

after:
`transformPath("./root/path")` => `./root/path`

visualization/app/codeCharta/util/nodeDecorator.spec.ts
```diff
@@ -23,7 +15,7 @@ describe("nodeDecorator", () => {
 
 	beforeEach(() => {
 		file = clone(TEST_DELTA_MAP_A)
-		map = file.map
+		;({ map } = file)
```
NearW, line 18: ew, that's ugly. I'd rather read the old style. :b

visualization/app/codeCharta/util/nodeDecorator.spec.ts
NearW: Why is this fixed? We check if ALL nodes have unique ids here. We can return as soon as we encounter an id, that we've seen already. Your suggestion is working the same, but it's traversing through all nodes instead of returning earlier.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
NearW: `codeMapBuidling` is `CODE_MAP_BUILDING` already.

```suggestion
```

NearW:
@BridgeAR The branch exploded with the addition of unicorn. I really like the changes, but this should've been done in a separate PR. Also it would've been great to have a dedicated commit for the auto-fixed linting rules.
I hope you commented all the changes you did manually, otherwise I'll have to go through 226 files and take a closer look.

I'll review the remaining performance part again soon.

visualization/app/codeCharta/util/nodeDecorator.spec.ts
NearW: Can't we do it like this?
```ts
		file = clone(TEST_DELTA_MAP_A);
	        ({ map } = file)
```

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
BridgeAR: Good catch. This was actually auto fixed. I'll deactivate the rule for now.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.ts
BridgeAR: It's not strictly necessary but the return type is `boolean`. That is only guaranteed in case of the cast. Otherwise it might be `null`, `undefined` or an empty string.

visualization/app/codeCharta/util/nodeDecorator.spec.ts
BridgeAR: I agree and I prefer the old style as well. I'll fix that in the config.

visualization/app/codeCharta/util/codeMapHelper.ts
BridgeAR: This works as before, it just does less work :-)

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
BridgeAR: The test is somewhat redundant. I could either remove the test or make it a tad stricter.

visualization/app/codeCharta/codeCharta.mouseEvent.service.ts
BridgeAR: Done

visualization/app/codeCharta/util/nodeDecorator.spec.ts
BridgeAR: As a hint: it's not a for loop and there's a function scope :-)

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: We do trigger it in our tests. We likely won't reach this point in "regular" code though.

BridgeAR: It is actually possible to reach this in regular code. This part is there to load the files and there could e.g., be a network outage.

Our error handling here indicates that there was an error but it does not provide any information about what actually has happened. We might want to change that.

NearW: How do we determine the error though? Should we throw something generic like `Something went wrong while loading files. Please try again or contact us on Github`?

visualization/app/codeCharta/codeCharta.component.ts
```diff
@@ -42,20 +42,19 @@ export class CodeChartaController {
...
-			.then((data: NameDataPair[]) => {
-				if (data.length > 0) {
-					this.tryLoadingFiles(data)
-					this.setRenderStateFromUrl()
-				} else {
-					this.tryLoadingSampleFiles()
-				}
-			})
-			.catch(() => {
+	async loadFileOrSample() {
```
NearW, line 45: Either set it to public here or let's remove the public type overall. Functions without a keyword there are implicit public. [This rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md) should do the job with the `no-public` property.

visualization/app/codeCharta/state/store/dynamicSettings/colorRange/colorRange.service.ts
```diff
@@ -41,17 +41,17 @@ export class ColorRangeService implements StoreSubscriber, ColorMetricSubscriber
 	}
 
 	private tryToResetIfNull() {
-		const colorRange = this.storeService.getState().dynamicSettings.colorRange
-		const colorMetric = this.storeService.getState().dynamicSettings.colorMetric
-		const maxMetricValue: number = this.nodeMetricDataService.getMaxMetricByMetricName(colorMetric)
+		const { colorRange } = this.storeService.getState().dynamicSettings
+		const { colorMetric } = this.storeService.getState().dynamicSettings
```
NearW, line 45:
```suggestion
		const { colorMetric, colorRange } = this.storeService.getState().dynamicSettings
```

visualization/app/codeCharta/state/store/dynamicSettings/margin/margin.service.ts
```diff
@@ -40,12 +39,12 @@ export class MarginService implements StoreSubscriber, DynamicMarginSubscriber,
 	}
 
 	public reset() {
-		const map: CodeMapNode = this.codeMapPreRenderService.getRenderMap()
-		const areaMetric = this.storeService.getState().dynamicSettings.areaMetric
-		const margin = this.storeService.getState().dynamicSettings.margin
-		const dynamicMargin = this.storeService.getState().appSettings.dynamicMargin
+		const map = this.codeMapPreRenderService.getRenderMap()
+		const { areaMetric } = this.storeService.getState().dynamicSettings
+		const { margin } = this.storeService.getState().dynamicSettings
```
NearW, line 44:
```suggestion
		const { areaMetric, margin } = this.storeService.getState().dynamicSettings
```

visualization/app/codeCharta/ui/attributeTypeSelector/attributeTypeSelector.component.html
```diff
@@ -5,8 +5,8 @@
 		</md-button>
 	</div>
 	<md-menu-content width="2">
-		<md-menu-item style="min-height: 20px; height: 20px">
-			<md-button style="min-height: 20px; height: 20px; line-height: 20px; font-size: 9pt; text-align: center" ng-disabled="true"
+		<md-menu-item style="min-height: 20px; height: 20px;">
+			<md-button style="min-height: 20px; height: 20px; line-height: 20px; font-size: 9pt; text-align: center;" ng-disabled="true"
```
NearW, line 9: How did this sneak in? Is this part of a new linting rule? Prettier didn't add a semicolon before.

visualization/app/codeCharta/ui/codeMap/codeMap.render.service.ts
```diff
@@ -8,31 +8,25 @@ import { CodeMapArrowService } from "./codeMap.arrow.service"
...
 		this.scaleMap()
 	}
 
 	private setNewMapMesh(sortedNodes) {
-		const mapMesh: CodeMapMesh = new CodeMapMesh(
-			sortedNodes,
-			this.storeService.getState(),
-			isDeltaState(this.storeService.getState().files)
-		)
+		const mapMesh = new CodeMapMesh(sortedNodes, this.storeService.getState(), isDeltaState(this.storeService.getState().files))
```
NearW, line 29: did we change anything in the prettier width?

visualization/app/codeCharta/ui/codeMap/rendering/geometryGenerator.ts
```diff
@@ -134,52 +136,48 @@ export class GeometryGenerator {
...
-		const deltas: Float32Array = new Float32Array(numVertices)
-
-		for (let i = 0; i < numVertices; ++i) {
-			positions[i * dimension + 0] = data.positions[i].x
-			positions[i * dimension + 1] = data.positions[i].y
-			positions[i * dimension + 2] = data.positions[i].z
+		const positions: Float32Array = new Float32Array(size)
+		const normals: Float32Array = new Float32Array(size)
+		const uvs: Float32Array = new Float32Array(numberVertices * uvDimension)
+		const colors: Float32Array = new Float32Array(size)
```
NearW, line 148: can't these types be removed as well?

visualization/app/codeCharta/ui/dialog/dialog.download.component.ts
```diff
@@ -47,7 +47,7 @@ export class DialogDownloadController {
 			this.codeMapPreRenderService.getRenderMap(),
 			this.codeMapPreRenderService.getRenderFileMeta(),
 			this.storeService.getState().fileSettings,
-			this._viewModel.fileContent.filter(x => x.isSelected == true).map(x => x.name),
+			this._viewModel.fileContent.filter(x => x.isSelected === true).map(x => x.name),
 			this._viewModel.fileName
```
NearW, line 51: Unfortunately I can't mark all lines of code, but we can combine the filtering and mapping:

```ts
	public download() {
		const downloadSettingsNames: string[]  = []
		for (const fileContent of this._viewModel.fileContent) {
			if (fileContent.isSelected) {
				downloadSettingsNames.push(fileContent.name)
			}
		}

		FileDownloader.downloadCurrentMap(
			this.codeMapPreRenderService.getRenderMap(),
			this.codeMapPreRenderService.getRenderFileMeta(),
			this.storeService.getState().fileSettings,
			downloadSettingsNames,
			this._viewModel.fileName
		)
		this.hide()
	}
```

I'd recommend to move the for-loop to a private function `getSettingsNames` though to keep the one-liner in `download()`

visualization/app/codeCharta/ui/fileExtensionBar/fileExtensionBar.component.ts
```diff
@@ -34,16 +34,13 @@ export class FileExtensionBarController implements CodeMapPreRenderServiceSubscr
 			.getMeshDescription()
 			.buildings.filter(building => building.node.isLeaf)
 
-		const visibleFileExtensions: string[] = this._viewModel.distribution
-			.filter(metric => metric.fileExtension !== "other")
-			.map(metric => metric.fileExtension)
+		const visibleFileExtensions = new Set(
+			this._viewModel.distribution.filter(metric => metric.fileExtension !== "other").map(metric => metric.fileExtension)
```
NearW, line 38: this can be combined again

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.ts
```diff
@@ -90,15 +85,15 @@ export class MapTreeViewLevelController implements BuildingHoveredSubscriber, Bu
 		return markingColor ? markingColor : defaultColor
 	}
 
-	public isSearched(): boolean {
-		if (this.node != null && this.storeService.getState().dynamicSettings.searchedNodePaths) {
+	public isSearched() {
+		if (this.node && this.storeService.getState().dynamicSettings.searchedNodePaths) {
```
NearW, line 89:
```suggestion
		if (this.storeService.getState().dynamicSettings.searchedNodePaths) {
```

Just tested this and I couldn't encounter a case where node is not set. Let's remove this safeguard

visualization/app/codeCharta/ui/metricDeltaSelected/metricDeltaSelected.component.ts
```diff
@@ -35,7 +35,7 @@ export class MetricDeltaSelectedController implements BuildingSelectedEventSubsc
 		}, MetricDeltaSelectedController.TIME_TO_INIT_BINDING)
 	}
 
-	public onBuildingSelected(selectedBuilding: CodeMapBuilding) {
+	public onBuildingSelected(selectedBuilding?: CodeMapBuilding) {
```
NearW, line 38: there exists an interface of this. Let's update the parameter to optional there as well?

visualization/app/codeCharta/util/fileExtensionCalculator.ts
```diff
@@ -23,14 +23,14 @@ export class FileExtensionCalculator {
 		return distribution
 	}
 
-	private static getAbsoluteDistribution(map: CodeMapNode, metric: string): MetricDistribution[] {
+	private static getAbsoluteDistribution(map: CodeMapNode, metric: string) {
 		const distribution: MetricDistribution[] = []
 		hierarchy(map)
 			.leaves()
-			.forEach((node: HierarchyNode<CodeMapNode>) => {
+			.forEach(node => {
```
NearW, line 30: We can destructure this to `{data}` again, since only data is used.

visualization/app/codeCharta/util/fileValidator.spec.ts
```diff
@@ -179,15 +179,14 @@ describe("FileValidator", () => {
 
 		beforeEach(() => {
 			file = clone(fileWithFixedFolders)
-			folder1 = file.nodes[0].children[0]
-			folder2 = file.nodes[0].children[1]
+			;[folder1, folder2] = file.nodes[0].children
```
NearW, line 182: move semicolon behind `file = clone(fileWithFixedFolders)` maybe?

visualization/app/codeCharta/util/scenarioHelper.spec.ts
```diff
@@ -17,7 +17,7 @@ describe("scenarioHelper", () => {
 
 	describe("importScenarios", () => {
 		it("should convert vectors when importing scenarios", () => {
-			scenarios[0].settings.appSettings.camera = { x: 0, y: 1, z: 2 } as any
+			scenarios[0].settings.appSettings.camera = { x: 0, y: 1, z: 2 } as Vector3
```
NearW, line 20:
```suggestion
			scenarios[0].settings.appSettings.camera = new Vector3(0,1,2)
```

visualization/app/codeCharta/util/scenarioHelper.ts
```diff
@@ -79,8 +79,8 @@ export class ScenarioHelper {
 
 	private static transformScenarioAsSettingsToScenario(scenarioAsSettings: ExportScenario) {
 		const scenario: RecursivePartial<Scenario> = { name: scenarioAsSettings.name }
-		const dynamicSettings: RecursivePartial<DynamicSettings> = scenarioAsSettings.settings.dynamicSettings
-		const appSettings: RecursivePartial<AppSettings> = scenarioAsSettings.settings.appSettings
+		const { dynamicSettings } = scenarioAsSettings.settings
+		const { appSettings } = scenarioAsSettings.settings
```
NearW, line 83:
```suggestion
		const { appSettings, dynamicSettings } = scenarioAsSettings.settings
```

visualization/app/codeCharta/util/treeMapHelper.ts
```diff
@@ -100,10 +100,10 @@ function buildNodeFrom(
 		color: getBuildingColor(squaredNode.data, s, isDeltaState, flattened),
 		incomingEdgePoint: getIncomingEdgePoint(width, height, length, new Vector3(x0, z0, y0), s.treeMap.mapSize),
 		outgoingEdgePoint: getOutgoingEdgePoint(width, height, length, new Vector3(x0, z0, y0), s.treeMap.mapSize)
-	}
+	} as Node
```
NearW, line 103: I encountered another line of code where you defined a return type instead of declaring what the type is during return.
We should choose one. I'd rather keep the return type.

visualization/app/codeCharta/util/urlExtractor.spec.ts
```diff
@@ -72,7 +72,7 @@ describe("urlExtractor", () => {
 			expect(urlExtractor.getFileDataFromFile).toHaveBeenCalledWith({ data: "some data" })
 		})
 
-		it("should return an array of resolved file data", () => {
+		it("should return an array of resolved file data", async () => {
```
NearW, line 75: is `async` even required when we don't use await?
I see the benefit of telling the developer that some asynchronous code execution is happening here, but the keyword shouldn't be required.

visualization/plop/templates/redux/reducer.ts.hbs
```diff
@@ -4,7 +4,7 @@ import { clone } from "../../../util/clone"
 export function {{camelCase name}}(state: {{type}} = set{{properCase name}}().payload, action: {{properCase name}}Action): {{type}} {
```
NearW, line 4:
```suggestion
export function {{camelCase name}}(state = set{{properCase name}}().payload, action: {{properCase name}}Action) {
```

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: That is already the case. The error is: `One or more files from the given file URL parameter could not be loaded. Loading sample files instead.`. I thought we might extend it to: `One or more files from the given file URL parameter could not be loaded. Loading sample files instead.\n\nError: $error.message`

visualization/app/codeCharta/ui/attributeTypeSelector/attributeTypeSelector.component.html
BridgeAR: I just ran prettier against all files *shrug*

visualization/app/codeCharta/ui/dialog/dialog.download.component.ts
BridgeAR: I kept it as is, since the number of files to download is low. Thus, the readability wins here IMO.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.ts
BridgeAR: I would rather keep this for now and maybe change it in another PR.

visualization/app/codeCharta/util/fileValidator.spec.ts
BridgeAR: That's done by prettier.

visualization/app/codeCharta/util/scenarioHelper.spec.ts
BridgeAR: This test is intended to check that it works with regular objects as well.

visualization/app/codeCharta/ui/attributeTypeSelector/attributeTypeSelector.component.html
NearW: As long as prettier keeps this format I'm fine. :)

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.level.component.ts
NearW: I'll leave the conversations regarding removing safeguards, which need a test adaptation open. It's easier for me to create a follow up PR based on the open conversations then.

visualization/app/codeCharta/util/urlExtractor.spec.ts
BridgeAR: The behavior is a tad different when using `async` as keyword. This is a lint rule that I personally would keep. It's otherwise difficult to determine that it's indeed an async function.

visualization/app/codeCharta/ui/attributeTypeSelector/attributeTypeSelector.component.html
NearW: I just run `npm run format` and it removed those semicolons. I'll commit my formatation.

visualization/app/codeCharta/ui/fileExtensionBar/fileExtensionBar.component.ts
BridgeAR: Done. I also combined the buildings for loop with the loop below.

visualization/app/codeCharta/codeCharta.component.ts
BridgeAR: Done

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.spec.ts
NearW: I just read the test for the highlighting. Those are horrible.
The test I marked here should be removed and the tests should be adapted in the future. They're not really describing what is actually expected and are not testing the actual behavior.
Do you want to remove this test and add a TODO?

visualization/app/codeCharta/codeCharta.component.ts
NearW: e2e test is not happy atm

visualization/app/codeCharta/e2e/url.e2e.ts
```diff
@@ -38,7 +38,7 @@ describe("codecharta", () => {
 
 	async function handleErrorDialog() {
 		const message = await dialogError.getMessage()
-		expect(message).toEqual("One or more files from the given file URL parameter could not be loaded. Loading sample files instead.")
+		expect(message).toEqual("One or more files from the given file URL parameter could not be loaded. Loading sample files instead. undefined")
```
NearW, line 41: Is this actually a reasonable test? Printing `undefined` at the end is not providing any useful information.
Can this happen in the application during runtime?

BridgeAR: Not really. The error should always have a message.

NearW: We should find out why we can't simulate this in our tests. This test is not useful like this and will confuse people in the future.

BridgeAR: In theory, yes. It should already be an improvement in the average case though. Should I add a TODO comment?

NearW: Yes, please. I really want to merge this branch.

Issue #787, opened 2019-11-16
CodeCharta Analysis Coverage is 0%


Richargh:
# Bug

## Expected Behavior

Coverage fulfills the quality profile we set ourselves.

## Actual Behavior

* Version 1.22.0 March 2019 had 48% coverage
* Version 1.23.0 April 2019 dropped to 12%, then to "up" to 14.1%
* Version 1.24.0 May 2019 went from 14.1% to 0%

BridgeAR:
@RomanenkoVladimir had a look at this and the outcome is that it's indeed as low as reported. Writing tests should always be a high priority and to make sure the coverage is increasing we'll take the following steps:

1. We include tools such as codecov to verify that pull requests always increase the coverage. Lowering the coverage would prevent merging the PR and as such enforce writing more tests.
2. We start with an initial sprint focusing on writing tests for files with a coverage < 10%.
3. Anyone who is just remotely working on parts that are not yet covered should write a test for such code.

It'll then be an ongoing process to increase the coverage while not blocking development of other new features.

RomanenkoVladimir:
We discussed the frontend sonar issues to also be a part of this, 
the frontend part should be fixed now

NearW:
I am pretty sure that our coverage is not 0%. There are actually a lot of tests in the analysis.
The problem here must be an issue with our sonarcloud configuration.

Sonarcloud is not running tests. Sonarcloud is actually importing coverage results. We build these with travis and publish them on sonar afterwards. You have to specify where sonar finds these.

For Javascript/Typescript this must be specified in a sonar.properties file with `sonar.typescript.lcov.reportPaths=./dist/coverage/lcov.info`. 

For Kotlin this must be done here `sonar.coverage.jacoco.xmlReportPaths`. See https://sonarcloud.io/documentation/analysis/coverage/ for more information on that.

We need to generate coverage reports and specify the path to that in our build.gradle file.

PR #1287, opened 2020-09-23
Tech/test not logged and not run

RomanenkoVladimir:

Issue: #787 (connected to?)

## Description
Some tests in the backend are skipped due to us using a Junit5 test engine and attempting to run Junit4 tests on it.
This PR adresses the issue, when merged it will:

- Change Testing to Junit 5
  - enable legacy support, with Vintage Engine for SCMLogParser
- Add Logging for test output
- Update Mokk version to the current 1.10 to fix compatibility issues between Mokk and Corroutines
- remove deprecated gradle key words from scripts (still lack compatibility to gradle 7 due to used libraries)


analysis/import/UnderstandImporter/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
 
     runtime group: 'org.slf4j', name: 'slf4j-simple', version: slf4j_version
 
+    testImplementation('org.junit.jupiter:junit-jupiter:5.6.0')
```
NearW, line 17: Why do we need junit5 if we run the tests using spek2?
Isn't it enough to just include spek2?

analysis/filter/EdgeFilter/build.gradle
```diff
@@ -27,7 +27,10 @@ dependencies {
 }
 
 test {
-    useJUnitPlatform {
+    useJUnitPlatform{
```
NearW, line 30: Do we have formatting for `build.gradle` files?

NearW (commented):
LGTM, just had a few questions.

analysis/filter/StructureModifier/build.gradle
```diff
@@ -7,10 +7,13 @@ dependencies {
     testImplementation group: 'org.jetbrains.kotlin', name: 'kotlin-test', version: kotlin_version
     testImplementation project(path: ':model', configuration: 'testOutput')
 
-    testImplementation group: 'junit', name: 'junit', version: junit4_version
+    testImplementation('org.junit.jupiter:junit-jupiter:5.6.0')
```
BridgeAR, line 10: Please keep the former notation (it is also possible to use a variable for the version) and move the version out in a variable (if there's none already that could be used instead)

analysis/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Sep 23 13:22:07 CEST 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6.1-all.zip
```
BridgeAR, line 2: I suggest to use a variable for the version here to keep the version up to date.

analysis/import/CSVImporter/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
 
     runtime group: 'org.slf4j', name: 'slf4j-simple', version: slf4j_version
 
+    testImplementation('org.junit.jupiter:junit-jupiter:5.6.0')
```
BridgeAR, line 17: The same as above

analysis/import/CrococosmoImporter/build.gradle
```diff
@@ -6,9 +6,13 @@ dependencies {
     implementation group: 'io.fastjson', name: 'boon', version: boon_version
     implementation group: 'info.picocli', name: 'picocli', version: picocli_version
 
+    testImplementation('org.junit.jupiter:junit-jupiter:5.6.0')
```
BridgeAR, line 9: The same as above and this applies to multiple other files as well.

RomanenkoVladimir:
Ok, so I found the/a? issue:
Currently in SCMLogParser on main (as well as SCMLogParserV2 in development), we run junit4 tests on a junit5 engine, that seems to be the cause for some tests being skipped. 
I switched to junit5 where it was easy to do and enabled legacy support elsewhere (SCMLogParser).

The good: The change would run all tests previously skipped due to not being runnable on the Junit5 test engine, sadly I am not sure how many those are but it affects SCMLogParser especially (previously most tests here seem to have been skipped, from what I could tell with manual experiments and logging)

The bad: Some tests fail, 3 tests in SCMLogParser, however they also fail on the main branch if you call them manually (without gradle), meaning they never actually worked in the first place, but travis will be red with this "fix"

Additional thoughts:
- While legacy support is enabled and supported by JUnit we have already a huge bloat of different testing libraries and a testing refactor PR would be very usefull in my opinion, this could also help us remove JUnit4 by changing the few tests accordingly

- Our gradle script is really broken in regards to testing, with too many libraries doing more or less the same to the point where even the IDEA fails with autoimports from time to time, I am not sure if  more tests are skipped due to conflicts or outdated engines or similar problems

NearW:
> The bad: Some tests fail, 3 tests in SCMLogParser, however they also fail on the main branch if you call them manually (without gradle), meaning they never actually worked in the first place, but travis will be red with this "fix"

Well, at least we found out now. Let's fix them though.

> While legacy support is enabled and supported by JUnit we have already a huge bloat of different testing libraries and a testing refactor PR would be very usefull in my opinion, this could also help us remove JUnit4 by changing the few tests accordingly.

I absolutely agree. I'd use JUnit5.

RomanenkoVladimir:
@NearW the SourceCodeParser was also using junit4 Tests before, so tests would have been skipped.
It is possible that this caused the behaviour you mentioned.

RomanenkoVladimir:
I would propose merging this at its stage and remove Junit 4 entirely during a follow-up PR.
This would enable to test the backend properly again, also the new SCMLogParserV2 has a few Junit4 tests as well, therefore I would propose waiting with the change until the feature is merged.

NearW (approved):

PR #1294, opened 2020-09-24
Feature/remove unary metric

NearW:
# Remove automatically added unary metric

## Description

This PR removes the automatically added unary metric and replaces the calculation for the number of files with something more intuitive.

- Removed some test data only used to verify. Instead a snapshot is created now.
- Refactored some tests to use common test data (TEST_DELTA_MAP_A)
- Added validation to verify that we have metric data available now (otherwise the application would crash)


visualization/app/codeCharta/assets/scenarios.json
```diff
@@ -20,10 +20,10 @@
         "invertColorRange": false
       },
       "dynamicSettings": {
-        "areaMetric": "unary",
+        "areaMetric": "Average Complexity*",
         "heightMetric": "Average Complexity*",
         "colorMetric": "Average Complexity*",
-        "distributionMetric": "unary"
+        "distributionMetric": "Average Complexity*"
```
NearW, line 26: We should get some feedback on how this scenario should be set now.

visualization/MANUAL_TESTING.md
```diff
@@ -354,21 +354,6 @@
 
 ## When
 
--
```
NearW, line 357: This document was hard work but we never used it in the last 2 years. Should we remove this?
Tests should be automated.

BridgeAR: I had a brief look at it and it does seem rather outdated. I agree that it's probably best to remove it.

visualization/app/codeCharta/assets/scenarios.json
BridgeAR: This might be a blocker as far as I see it?

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -46,16 +45,13 @@ export class NodeDecorator {
 		let id = 0
 		for (const { data } of hierarchy(map).descendants()) {
 			data.id = id
+			data.descendants = 0
```
BridgeAR, line 48: I suggest to only set this property on nodes that are not leaves.

BridgeAR (commented):
LGTM besides the fact that changes of our dynamic settings. Is it possible to check the node number in that case and to keep `unary` for this specific case? We'd not add the metric itself but know that we have to check for the node number instead?

visualization/app/codeCharta/assets/scenarios.json
NearW: I think that the areaMetric should be set to `rloc` all the time. In this case I'd replace the unary-metric with rloc here.

NearW:
@BridgeAR I found a really good point to catch the error, when no metrics were found. Please check [this commit](38a3419fd5b4597609ce5ca53497c064edd72d36).
However, it's not possible to revert back to a previous state with our current architecture. We need some sort of history of all actions, that have been dispatched to the state including their previous state. We then just need to traverse the history backwards and reapply the previous state.

NearW:
Current state of this PR:
In order to improve performance, we'll add an undo functionality to the filestate-reducer. If we load a file without metrics, we can revert back to the previous file states. A guide on how to implement such functionality can be found [here](https://redux.js.org/recipes/implementing-undo-history).

Issue #1266, opened 2020-09-18
First start of local codecharta-visualization OFFLINE fails


ce-bo:
# Bug
After the installation of codecharta-visualization, I want to start the visualization offline.
But it doesn't work.

## Expected Behavior
codecharta-visualization is able to be launched in offline mode.

## Actual Behavior
After the installation of codecharta-visualization, I want to start the visualization offline.
But an error occurs:

## Screenshots
![cc-visualization-install-offline-error3](https://user-images.githubusercontent.com/26900540/93565589-e5d60400-f98b-11ea-9077-aeef42826f8b.jpg)

## Steps to Reproduce the Problem
1. > npm install codecharta-visualization -g
2.  Turn on flight mode / turn off internet
3. > codecharta-visualization

## Specifications
- CodeCharta Version: latest 1.57
- OS: Windows
- Browser: cmd.exe :)

NearW:
We don't ship NWJS with the installation. NWJS is downloaded when you open the application for the first time.

Should we change that? And can we change that?

NearW:
We currently use the [nwjs-builder-phoenix](https://github.com/evshiron/nwjs-builder-phoenix) to run the application as a standalone. Unfortunately, this library can't pre-fetch the nwjs binary (which are around 130mb in size). Instead, it downloads them when starting a standalone for the first time.

We could install [nw](https://www.npmjs.com/package/nw) directly and run `nw` to start our application. This is pre-fetching the binaries correctly and starts a standalone. However: It prints some really weird messages into the console, which do not affect the application. They just look really suspicious.

What do we want to do?
Keep `nwjs-builder-pheonix` and instantly start the application with a post-install-script after installing it to downlaod the binaries **OR** use `nw` directly and ignore the console spam.

![Bildschirmfoto 2020-09-22 um 18.57.32.png](https://images.zenhubusercontent.com/5cd16d629795a865159e710f/cee1c1e7-729c-4299-a06c-9cb0bc091a6f)

PR #1320, opened 2020-10-03
Fix/1266/first startup fails

NearW:
# Fix first startup of standalone without internet connection

fixes: #1266 

## Description

- Install nw directly and remove nwjs-builder-phoenix, since we just run the application in nwjs and don't build it with nw. We use webpack instead.
- This will install the nw binaries on install now.


RomanenkoVladimir:
It works for me now
I will fix the conflicts and then check again just in case.

What would you say @ce-bo?

RomanenkoVladimir (approved):

ce-bo (approved):
It works from my point of view!

Issue #1324, opened 2020-10-04
Zooming should close the nodes context menu


BridgeAR:
# Bug

## Expected Behavior

GIVEN a context menu being opened by a right click on a maps node

WHEN zooming in or out

THEN the context menu vanishes

## Actual Behavior

The context menu stays around at the current position

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/95025184-40f44000-0688-11eb-8e7e-6d11b8f4f17b.png)
![image](https://user-images.githubusercontent.com/8822573/95025188-45b8f400-0688-11eb-831d-445dad61b808.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.58.1

PR #1366, opened 2020-10-13
Feature/1324/close context menu on zoom

ce-bo:
# Close node context menu when zoom in and out the map

fixes: #1324 

## Description

- The node context menu will not be closed at the moment, if the user "scrolls" or zoom in and out the map.
- This PR will dispatch the appropriate event to close the menu properly.
- Furthermore, an E2E test was added, which can not be finalized due to missing TS type definitions of the latest puppeteer API. A TODO was added to finalize the test later.
- Some false-positive E2E tests have been improved within this PR.


visualization/app/codeCharta/ui/codeMap/codeMap.po.ts
```diff
@@ -1,6 +1,13 @@
 export class CodeMapPageObject {
 	async rightClickMap() {
 		await expect(page).toClick("#codeMap", { button: "right", timeout: 3000 })
-		await page.waitForSelector("node-context-menu-component", { visible: false })
+	}
+
+	// TODO: activate this function call
+	//  if the new puppeteer mouse wheel API is supported by it's type definition:
+	// https://github.com/DefinitelyTyped/DefinitelyTyped/issues/48710
```
BridgeAR, line 8: Please already use the API and ignore the type error with a comment as in `// @ts-ignore`.

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
```diff
@@ -72,6 +72,11 @@ export class NodeContextMenuController
 			event => NodeContextMenuController.broadcastHideEvent(this.$rootScope, new Vector2(event.x, event.y)),
 			true
 		)
+		document.body.addEventListener(
+			"wheel",
+			event => NodeContextMenuController.broadcastHideEvent(this.$rootScope, new Vector2(event.x, event.y)),
```
BridgeAR, line 77: Would it not be better to attach the event listener to the concrete element that we want to check the wheel upon? That way we would not need the vector. The vector seems like a bit of overhead in general as it's only an object with x and y and I would just pass through the current event, since it's already allocated instead of allocating a new object.

`onHideNodeContextMenu()` could also be improved (which is triggered due to the event) to accept `undefined` as input. That way there's no need to check for the bounds with a default vector of -1, -1.

BridgeAR (approved):
Functionality wise LGTM. The name could be improved a bit and while being on it, we might also improve the event handling for cases like these in general :)

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.po.ts
```diff
@@ -10,7 +10,10 @@ export class NodeContextMenuPageObject {
 		await page.waitForSelector("#loading-gif-map", { visible: false })
 	}
 
-	async isClosed() {
-		return page.waitForSelector("node-context-menu-component", { visible: false })
+	async toBeClosed() {
```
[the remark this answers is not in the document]
ce-bo: I thought this name is a good idea. You would read  `await contextMenu.toBeClosed()`.
But it is not a typical name for an `isser` function. I apply your suggestion.

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
ce-bo: I refactored the NodeContextMenu component and applied your suggestion regarding the EventListeners.
Appropriate EventListeners will be added when the menu is shown and removed immediately after the listeners have been called. Therefore, the listeners will be called only, if the menu is actually opened.

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.spec.ts
```diff
@@ -10,11 +10,9 @@ import { StoreService } from "../../state/store.service"
 import { setMarkedPackages } from "../../state/store/fileSettings/markedPackages/markedPackages.actions"
 import { BlacklistType, MarkedPackage, NodeType } from "../../codeCharta.model"
 import { addBlacklistItem } from "../../state/store/fileSettings/blacklist/blacklist.actions"
-import { FocusedNodePathService } from "../../state/store/dynamicSettings/focusedNodePath/focusedNodePath.service"
-import { BlacklistService } from "../../state/store/fileSettings/blacklist/blacklist.service"
-import { MarkedPackagesService } from "../../state/store/fileSettings/markedPackages/markedPackages.service"
 import { focusNode, unfocusNode } from "../../state/store/dynamicSettings/focusedNodePath/focusedNodePath.actions"
 import { NodeDecorator } from "../../util/nodeDecorator"
+import {CodeMapMouseEventService} from "../codeMap/codeMap.mouseEvent.service";
```
BridgeAR, line 15: Nit: maybe run the formatter (prettier) once? :-) We should likely run the formatter automatically. It's placed in the root scope of the monorepo right now.

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
```diff
@@ -62,82 +59,96 @@ export class NodeContextMenuController
...
 		this.setPosition(x, y)
+
 		this.synchronizeAngularTwoWayBinding()
+
+		// Add event listeners, so that opened node context menu can be closed again later
+		// when clicking (left or right button)
+		// or using the mouse wheel on the body element.
+		document.body.addEventListener("click", this.onBodyLeftClickHideNodeContextMenu, true)
+		document.body.addEventListener("mousedown", this.onBodyRightClickHideNodeContextMenu, true)
+		document.body.addEventListener("wheel", this.onBodyWheelHideNodeContextMenu, true)
```
BridgeAR, line 87: Would it not be sufficient to add the event listener not to the body but to the dom element that we want to track the wheel upon?

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.po.ts
```diff
@@ -10,7 +10,17 @@ export class NodeContextMenuPageObject {
...
+		const menu = await page.waitForSelector("#codemap-context-menu", { visible: true })
+		const isHidden = await menu.evaluate(element => { return element.getAttribute("aria-hidden") })
+
+		return isHidden !== 'true'
+	}
+
 	async isClosed() {
-		return page.waitForSelector("node-context-menu-component", { visible: false })
+		const menu = await page.waitForSelector("#codemap-context-menu", { hidden: true })
+		const isHidden = await menu.evaluate(element => { return element.getAttribute("aria-hidden") })
```
BridgeAR, line 22: Aria attributes are used for e.g., screen readers. Do we actually always use these at all?

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
ce-bo: I checked this. For the wheel listener, I agree.
The other listeners should be added to the body element, so that clicks on the ribbon bar and other interfaces will close the node context menu.

PR #1400, opened 2020-10-19
Fix size issues for expanded ribbon bar

RomanenkoVladimir:
# Fix expanded ribbon bar size

## Description

The last PR introduced an ugly ribbon-bar size for non height-setting components, this fixes the issue as well as making it a bit more prettier by changing a few other expanded sizes.


BridgeAR:
Please only merge reviewed PRs :-)

RomanenkoVladimir:
Ok, that was my fault. 

It was only a small CSS change and we never released the issue to the user so I thought I can release it with the new fixed 1.60.1 release right now, before any user would be bothered by it.

I will refrain from it in the future.

Issue #1318, opened 2020-10-02
Manage custom configuration sets (add/delete/load)


ce-bo:
# Feature request (raw)
Feature: Save current configuration. Add a button to the main menu (next to "scenario") that allows to save the current setting. It requires a setting name. As soon as saved the setting (redux state without the cc.json part - the latter is ) (simple + or "add" button next to an input field for the name) it's possible to jump back to that configuration / setting by clicking on the configurations name. Next to the configuration name there's a delete button (similar to the custom scenario ones). This is only kept in memory! Do not use the local storage.

## Description

As an auditor, I want to be able to switch between preconfigured map configurations so that I can return to earlier states easily and switch back and force.

## Acceptance criteria
- There is a button to open a menu which lists all currently saved configurations with its names next to "custom scenarios".
- There is another button inside the opened menu to add a new custom view ("+" label).
- The buttons show appropriate titles on hover.
- Inside the menu, I can delete individual configurations
- If a specific configuration is clicked, the configuration will be loaded and the map will be re-rendered.

## Development notes (optional Task Breakdown)
- An API has to be defined as similar to the cc.json schema
- The API has to include the following parts initially:
    - Overall everything that is necessary to provide the exact view of the user later on
    - Excluded/Filtered/Blacklisted parts
    - Presentation mode and view settings
    - Current render mode (single/multi/delta)
    - Focused building
    - View angle
    - Zoom factor
    - Hash value of the map (a single configuration is always only valid for a specific map.)
    - used metrics and thresholds (colors) as well
    - distribution of a metric (e.g. rloc) per programming language
     - ....
- The API has to be versioned
- Configurations for delta and multiple mode are only valid in case that both maps are loaded and match the hash value.
- It may not be identical to the redux state as the redux state may change at any point of time
- Configuration name (optional: default is Map Name + increasing number)
- It is possible to add infinite configurations per map.
- We need a logic to transform the redux state and all it's necessary (API) information to a configuration and vice versa.
- The configuration should never include the metric data itself, just the metric types
- The data is saved in the Local Storage
- Mark configurations which are not mapped to the current map as disabled or rather hide them in a collapsed list.

ce-bo:
We have encountered a problem when setting the state and setting the camera position from the custom view afterwards, the camera position will be overridden by the default camera position.
This is due to the fact that events are processed without guaranteeing the order.

PR #1405, opened 2020-10-20
Feature/1318/manage and apply custom views

ce-bo:
# Manage and apply CustomViews

closes: #1318 

## Description

  - The new CustomViews Feature has been added.
  - You can add/delete/load CustomViews from LocalStorage
  - A stored CustomView represents a CodeCharta configuration at a certain time. By loading it, you can replay that configuration (Blacklist entries, Camera View Angle, selected Metrics, etc.).
  - The Feature is experimental and hidden by default
  - A new Global Settings Option has been added called "Enable Experimental Features"
  - The CustomViews Feature can be enabled by this option. And afterwards the appropriate buttons are shown within the ribbon bar.

## Screenshots or gifs
CustomViews Buttons are hidden
image

As soon as you enable experimental features, you will be able to manage your CustomViews.
image


CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Added 🚀
 
+-   A new option in the Global Settings allows to enable/disable experimental features #1318
+    -   Click on the settings button in the upper right corner to open the Global Settings dialog.
+    -   Activate/Deactivate the new option "Enable Experimental Features"
+    -   The features will be shown/hidden accordingly
```
RomanenkoVladimir, line 15: A general question if old things don't work / are not existent do we want to keep flagging them as experimental?
aka. for me where do we draw the line? 
In my opinion if something is tested and an improvment over the orginal shouldn't we deploy it as a new feature?

visualization/README.md
```diff
@@ -46,8 +46,9 @@ Once you have installed the project, you can use all grunt tasks described in th
 
 #### Test
 
-`npm run test` runs all unit tests on the source files in app/ and generates a coverage report in dist/coverage/.
-`npm run e2e` runs all e2e tests on the built app in headless mode.
+- `npm run test` runs all unit tests on the source files in app/ and generates a coverage report in dist/coverage/.
+- `npm run e2e` runs all e2e tests on the built app in headless mode. You have to stop your running dev-webserver.
+To follow/watch the steps the e2e test is performing, deactivate headless mode in jest-puppeteer.config.js (and maybe set the slowMo parameter).
```
RomanenkoVladimir, line 51: Great addition, we should have done that way earlier

visualization/app/codeCharta/state/store/appSettings/appSettings.splitter.ts
```diff
@@ -1,136 +1,141 @@
...
+    if (payload.showMetricLabelNodeName !== undefined) {
+        actions.push(splitShowMetricLabelNodeNameAction(payload.showMetricLabelNodeName))
+    }
 
-	if (payload.showMetricLabelNameValue !== undefined) {
-		actions.push(splitShowMetricLabelNameValueAction(payload.showMetricLabelNameValue))
-	}
+    if (payload.showMetricLabelNameValue !== undefined) {
+        actions.push(splitShowMetricLabelNameValueAction(payload.showMetricLabelNameValue))
+    }
```
RomanenkoVladimir, line 42: The spacing seem to have changed? 
Are you using the commit pre-hooks, which should set those as far as I know, or the IDEA code-format option?

RomanenkoVladimir: Those are an examplary line choice, the spacing is changed in the whole file.

visualization/app/codeCharta/state/store/state.splitter.ts
```diff
@@ -1,5 +1,5 @@
 import { CCAction } from "../../codeCharta.model"
-import { StateActions } from "./state.actions"
+import {StateActions} from "./state.actions"
```
RomanenkoVladimir, line 2:
```suggestion
import { StateActions } from "./state.actions"
```

visualization/app/codeCharta/ui/customViews/customViewFileStateConnector.ts
```diff
@@ -0,0 +1,69 @@
...
+        if (typeof this.files !== 'undefined') {
+            this.processFiles()
+        }
+    }
+
+    private processFiles() {
+        for (const file of this.files) {
+            if (file.selectedAs !== FileSelectionState.None) {
+                this.setMapSelectionMode(file.selectedAs)
+                this.mapChecksums.push(md5(JSON.stringify(file.file.map)))
```
RomanenkoVladimir, line 25: How expensive is it to stringify and then to calculate the md5 hash on the result?
I'm just thinking maybe we can call this operation somewhere else, where we already work with the JSON, to reduce the potential overhead? 
Please correct me if I'm wrong here

visualization/app/codeCharta/ui/customViews/customViewFileStateConnector.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+            fileState.push(fileStateDelta)
+
+            const customViewFileStateConnector = new CustomViewFileStateConnector(fileState)
+
+            expect(customViewFileStateConnector.getJointMapName()).toBe('SINGLE MULTIPLE DELTA')
+        })
+    })
+
+    describe("set/is/getMapSelectionMode", () => {
+        it("should set SINGLE mode ad default", () => {
```
RomanenkoVladimir, line 45:
```suggestion
        it("should set SINGLE mode as default", () => {
```

visualization/app/codeCharta/ui/customViews/customViews.component.html
```diff
@@ -0,0 +1,47 @@
...
+                    title="Remove custom view"
+            >
+                <i class="fa fa-minus-square"></i>
+            </md-button>
+        </md-menu-item>
+    </md-menu-content>
+</md-menu>
+
+<md-menu class="custom-views-component">
+    <md-button class="custom-views-button plus" ng-click="$ctrl.showAddCustomViewSettings();" title="Fast save current view">
```
RomanenkoVladimir, line 43:
```suggestion
    <md-button class="custom-views-button plus" ng-click="$ctrl.showAddCustomViewSettings();" title="Quick save current view">
```
Just a personal preference, thats what it is called in my experience and would be more intuitive, in my opinion.

visualization/app/codeCharta/ui/dialog/dialog.globalSettings.component.spec.ts
```diff
@@ -1,160 +1,196 @@
...
 
 describe("DialogGlobalSettingsController", () => {
-	let dialogGlobalSettingsController: DialogGlobalSettingsController
-	let $mdDialog
-	let $rootScope: IRootScopeService
-	let storeService: StoreService
+    let dialogGlobalSettingsController: DialogGlobalSettingsController
+    let $mdDialog
+    let $rootScope: IRootScopeService
+    let storeService: StoreService
```
RomanenkoVladimir, line 20: The spacings are changed in this file as well?

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,140 @@
...
+
+		this.deepMapOneToOther(state, customView.stateSettings)
+
+		return customView
+	}
+
+	private static initializeAppSettings(target: CustomView) {
+		target.stateSettings.appSettings = {
+			showMetricLabelNameValue: false,
+			showMetricLabelNodeName: false,
```
RomanenkoVladimir, line 40:
```suggestion
			showMetricLabelNameValue: true,
			showMetricLabelNodeName: true,
```

from what I understood label should be shown by default, with the option to toggle them off?
Correct me if I'm wrong?

RomanenkoVladimir: Ok I see now that all is set to 0/ false?
Wouldn't it make more sense to set it all to the default settings from the codeCharta.model?

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,140 @@
...
+		this.initializeDynamicSettings(customView)
+		this.initializeFileSettings(customView)
+		this.initializeTreeMapSettings(customView)
+
+		this.deepMapOneToOther(state, customView.stateSettings)
+
+		return customView
+	}
+
+	private static initializeAppSettings(target: CustomView) {
```
RomanenkoVladimir, line 37: Shouldn't we set it all to the default settings in codeCharta.model, just for consistency?

visualization/app/codeCharta/util/customViewHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+    CCLocalStorage,
+    RecursivePartial, State
+} from "../codeCharta.model"
+import {CustomViewItem} from "../ui/customViews/customViews.component";
+import {CustomViewBuilder} from "./customViewBuilder";
+import {CustomView, CustomViewMapSelectionMode} from "../model/customView/customView.api.model";
+import {CustomViewFileStateConnector} from "../ui/customViews/customViewFileStateConnector";
+
+export class CustomViewHelper {
+    private static readonly CC_LOCAL_STORAGE_VERSION = "1.0.0"
```
RomanenkoVladimir, line 12: what is the version related to?
To codecharta/ visualization or anaylsis?

RomanenkoVladimir (commented):
Right now enabling experimental features triggers a map rerendering, you should check if you can blacklist that action in store.service.ts under dispatch to reduce unnecessary overhead.

And right now loading a scenario looks a bit buggy, due to it triggering multiple events, maybe we could trigger a redraw process, like whne selecting a new file to avoid the "jumping", just as an idea?

Otherwise look good to me, great feature!

RomanenkoVladimir:
On a sidenote you could try increasing the timeout for e2e tests if yours keep timing out too soon, due to needing more time.
It would make the branch green :)

visualization/app/codeCharta/util/customViewBuilder.ts
RomanenkoVladimir: Ok that makes sense, closing the comment

RomanenkoVladimir:
Seems to have broken after a recent update, please take another look

visualization/app/codeCharta/ui/customViews/customViewFileStateConnector.ts
BridgeAR: The has value should always be calculated for a file, not only for a map. Ideally the file contains a precomputed checksum. If it does not (e.g., old cc.json files) calculate it on the fly.

ce-bo: I have had a look at this. We can calculate missing checksums on the fly, if a file has been choosen. We should also adjust the FileDownload, so that we generate a fresh checksum for the downloaded .cc.json file (not to forget: the parsers which generate .cc.json files as well).

For which properties the checksum would be built? Propably for all properties of the `ExportCCFile` Type (including Blacklist, MarkedPackges, etc.).
This will have negative consequences for existing CustomViews. As soon if the blacklist will change and the .cc.json file will be downloaded, all corresponding CustomViews are outdated for the loaded map and for the new downloaded file as well. Note that a CustomView can override the blacklist anyway. This is not very user friendly.

Solution? We could build a checksum during the download process based on the map nodes, projectName, apiVersion, attributeTypes and edges. If a file is uploaded, missing checksums could be generated on the fly.
But that checksum would not be a generic checksum! It would be a kind of "CustomViewChecksum" or anything like that.

We could add a new optional generic `attributes` object as a property of the `ExportCCFile` type. There we could store some kind of meta attributes for the .cc.json file. For example `attributes['customViewChecksum']` could hold the md5 hash of the map nodes, edges, apiVersion, and projectName (but not the blacklist and markedPackages). Any suggestions appreciated.

BridgeAR: Please ignore the current download possibility. In a later stage, the download button should be completely replaced and the cc.json file is not changed at all (and should not be downloaded). Instead, only a configuration file would be downloaded that corresponds with the original cc.json file.

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -53,11 +54,17 @@ export class FileChooserController {
 
 	private addNameDataPair(fileName: string, jsonString: string) {
 		let content: ExportCCFile
+
 		try {
 			content = JSON.parse(jsonString)
+
+			if (content.fileChecksum === undefined) {
```
BridgeAR, line 61:
```suggestion
			if (!content.fileChecksum) {
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+			mapSize: 0
+		}
+	}
+
+	private static deepMapOneToOther<T>(source: any, target: T) {
+		Object.keys(source).forEach(key => {
+			const value = source[key]
+
+			if (key in target) {
+				if (typeof value !== "object" || Array.isArray(value) || typeof value === "undefined" || value === null) {
```
BridgeAR, line 144:
```suggestion
				if (typeof value !== "object" || Array.isArray(value) || value === null || target[key] === undefined) {
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+
+	private static initializeTreeMapSettings(target: CustomView) {
+		target.stateSettings.treeMap = {
+			mapSize: 0
+		}
+	}
+
+	private static deepMapOneToOther<T>(source: any, target: T) {
+		Object.keys(source).forEach(key => {
+			const value = source[key]
```
BridgeAR, line 141:
```suggestion
		for (const [key, value] of Object.entries(source)) {
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+
+			if (key in target) {
```
BridgeAR, line 143:
```suggestion
			if (Object.prototype.hasOwnProperty.call(target, key)) {
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+						// Assign object to target property,
+						// if target property does not specify deeper properties.
+						target[key] = value
+					} else {
+						// We have to map an object with nested properties here.
+						// source and target have the same properties specified for the nested object.
+						// Thus, map properties of the next deeper level.
+						this.deepMapOneToOther(value, target[key])
+					}
+				}
```
BridgeAR, line 158:
```suggestion
				} else {
					// We have to map an object with nested properties here.
					// source and target have the same properties specified for the nested object.
					// Thus, map properties of the next deeper level.
					this.deepMapOneToOther(value, target[key])
				}
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}, {})
```
BridgeAR, line 160:
```suggestion
		}
```

visualization/app/codeCharta/util/customViewBuilder.ts
```diff
@@ -0,0 +1,162 @@
...
+			panelSelection: undefined,
+			resetCameraIfNewFileIsLoaded: false,
+			scaling: undefined,
+			searchPanelMode: undefined,
+			showOnlyBuildingsWithEdges: false,
+			sortingOrderAscending: false,
+			whiteColorBuildings: false,
+			experimentalFeaturesEnabled: false
+		}
+		target.stateSettings.appSettings.mapColors = {
```
BridgeAR, line 85:
```suggestion
			experimentalFeaturesEnabled: false,
			mapColors: {
```

visualization/app/codeCharta/ui/customViews/customViewFileStateConnector.ts
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	getJointMapName(): string {
+		return this.fileNameParts.join(" ")
+	}
+
+	private setMapSelectionMode(fileSelectionState: string) {
+		if (fileSelectionState === FileSelectionState.Partial) {
+			this.mapSelectionMode = CustomViewMapSelectionMode.MULTIPLE
+		} else if (fileSelectionState === FileSelectionState.Comparison || fileSelectionState === FileSelectionState.Reference) {
```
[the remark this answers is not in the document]
ce-bo: I do not think so. There are other modes beside Partial, Reference and Comparision. They will be mapped to SINGLE because `this.mapSelectionMode` defaults to SINGLE mode.

visualization/app/codeCharta/util/customViewHelper.ts
```diff
@@ -0,0 +1,164 @@
...
+	static getCustomViewSettings(viewId: string): RecursivePartial<CustomView> | undefined {
+		return this.customViews.get(viewId)
+	}
+
+	static hasCustomView(
+		mapSelectionMode: CustomViewMapSelectionMode,
+		selectedMaps: string[],
+		viewName: string
+	): boolean {
+		const customViewIdentifier = CustomViewBuilder.createCustomViewIdentifier(
```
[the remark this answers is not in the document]
ce-bo: The identifier is needed indeed. It is used to get the information, if a customView exists based on mode, assigned maps and the view name itself.

Issue #1170, opened 2020-08-25
Prevent mouseflickering when clicking a building


JaiJoPa:
# Feature request

## Description

When a building is clicked, then the mouse quickly changes to the drag icon. 
It would be nice, if we could prevent that it changes the mouse icon when a building is being clicked. 

## Acceptance criteria
- selecting or opening the nodeContextMenu to a building does not change the cursor icon
- when the map is rotated or moved then the cursors should change to the specific cursor icon during the process

## Assumptions & Exclusions

An idea would be to add a threshhold before showing the drag(left-clicked) or the move (left-clicked) icon

BridgeAR:
I just had tested this and ideally, the cursor just checks for being above the map, not about being above a building, since it's also possible to click on folders and the map consistent to 100% of buildings and folders. Thus, there's no area where the cursor should not be changed.

Therefore it's also best to just use a single on hover event (from the most outer map dom element).

After looking closer, it also seems that cursor movements trigger further events while the mouse button is hold down. That is not ideal, since holding down the mouse is already a specific state and there should not be another state change while being in that state.

PR #1411, opened 2020-10-21
Feature/1170/prevent mouse flickering

RomanenkoVladimir:
Prevent mouse flickering

closes: #1170

## Description
-  Fixed flickering with move and cursor icon: they are always prioritized when pressed
-  Fixed issue with cursor sometimes changing to default on folder hover
-  @BridgeAR I did not change the events, as we plan on changing them soon and the hovering needs to fire an event to highlight even in grabbing/moving mode. However, I could disable highlighting in those modes as it makes the map look a little like a firework on dragging it around. If you think this is a good idea?


ce-bo (approved):

RomanenkoVladimir:
We decided to change the icon whever the user performs a right or left click.
This seems more intuitive, providing direct feedback to the user that he could access additional functionalities with those buttons, which would be lost when adding a timeout.

If you dislike this behaviour we can change it in the future

BridgeAR (approved):
LGTM, awesome work!

Issue #1410, opened 2020-10-21
Map flips after first state change


ce-bo:
# Bug
After the map has been loaded initially, if you invert the colors (in the color metric panel), the map flips around.
This should not be the case.

Thanks @JaiJoPa for reporting this to me.

## Steps to Reproduce the Problem
1. Load CodeCharta Visualization
2. Open Color Metric Panel
3. Click on "Invert Colors"

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.60.1
- OS: Windows 10
- Browser: Chrome

PR #1412, opened 2020-10-21
Fix flipping map bug #1410

ce-bo:
# Clone map object before modifying it's order

closes: #1410 

## Description

The map has to be cloned, so that we do not modify the order of the original map by this.applySortOrderChange


RomanenkoVladimir (approved):
LGTM, nice catch.

This also fixes redrawing issues with other ribbon bar components, we should merge this ASAP in my opinion.

However, we should take a look how to restructure the project so that no deep cloning is needed.

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from "../../state/store/appSettings/sortingOrderAscending/sortingOrderAscending.service"
 import { SortingOptionService, SortingOptionSubscriber } from "../../state/store/dynamicSettings/sortingOption/sortingOption.service"
 import { NodeMetricDataService } from "../../state/store/metricData/nodeMetricData/nodeMetricData.service"
+import {klona} from "klona";
```
BridgeAR, line 11: Nit: please use prettier to format the code before commiting. We might want to add that step to the commit hook. It's a script in the main package.json (main folder) and not in the visualization. I am also fine to copy the script as long as we trigger it ;-)

BridgeAR (commented):
In general LGTM but it would be great to have an end to end test to verify the behavior

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.ts
ce-bo: I have added a regression (unit) test. 
I did not know how to test it with an end to end test because the canvas DOM element does not change.

ce-bo: > Nit: please use prettier to format the code before commiting. We might want to add that step to the commit hook. It's a script in the main package.json (main folder) and not in the visualization. I am also fine to copy the script as long as we trigger it ;-)

I created a new issue for that: #1420

BridgeAR:
Seems like we want to remove two files / add them to the gitignore file? :-) Or is it intentional to add more than 400.000 lines of code?

ce-bo:
@BridgeAR 

> Seems like we want to remove two files / add them to the gitignore file? :-) Or is it intentional to add more than 400.000 lines of code?
Of course not! Shouldn't it be the better way to exclude .json files from being reformatted by the formatter script?

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.spec.ts
```diff
@@ -107,5 +108,26 @@ describe("MapTreeViewController", () => {
...
 		})
+
+		it("should clone new map before updating viewModel.rootNode to prevent unwanted changes in original map", () => {
+			const testNode = VALID_NODE_WITH_MULTIPLE_FOLDERS
+			const untouchedOriginalTestNode = klona(testNode)
+
+			mapTreeViewController["_viewModel"].rootNode = null
+
+			mapTreeViewController.onRenderMapChanged(testNode)
+			$timeout.flush(100)
```
JaiJoPa, line 119: Could you please explain me what flush does and why you have to use it here?

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.ts
NearW: @BridgeAR the husky hook is located in root and will trigger for visualization as well (since visualization is a child of the project). Either commit through the terminal or if you use the GUI in IntelliJ check the `run commit hooks` option.
And don't forget to `npm i` on root.
Bildschirmfoto 2020-10-22 um 12 45 13

visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.spec.ts
ce-bo: There are active timeouts in the code. $timeout.flush() flushes all pending timeouts, so that the process executes the callback function of the timeout immediately (if any).

ce-bo: https://docs.angularjs.org/api/ngMock/service/$timeout#flush

BridgeAR (approved):

PR #1421, opened 2020-10-22
Tech/dependabot/adjust dependency update schedule

ce-bo:
# Adjust the schedules of Dependabot dependency updates

## Description
- Travis builds are waiting often for dependabot builds
- Dependabot should check for dependency once early at night
- So that the development process is not affected by long running Dependabot builds


.github/dependabot.yml
```diff
@@ -7,16 +7,22 @@ updates:
...
   schedule:
     interval: daily
+    time: "03:00"
   open-pull-requests-limit: 10
 - package-ecosystem: npm
   directory: "/visualization"
+  # Check the npm registry for updates at 3am UTC
   schedule:
     interval: daily
+    time: "03:00"
```
BridgeAR, line 27: I suggest to change this to e.g., midnight.

ce-bo: I think the problem is caused by something else. For example, dependabot force-pushes changed integrity checksums in the package-lock.json which leads to a new build every time: https://github.com/MaibornWolff/codecharta/pull/1414

Would changing the number of times we check if a PR is ready to merge solve the problems?

ce-bo: The actual problem is, that some dependabot branches will never be merged because the number of retries is to low.
... Resolved

CHANGELOG.md
```diff
@@ -105,6 +105,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Changed
 
+-   Schedules of dependabot dependency updates changed
```
BridgeAR, line 108: This should be moved into `Chore` :-)

BridgeAR (approved):

PR #1424, opened 2020-10-23
Bug/edge metric throws console errors

RomanenkoVladimir:
# Edge Metric broken on main

## Description
The edge metric broke during the last release, this amends the bug.

### DIFF
--- a/visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
+++ b/visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
@@ -78,9 +78,12 @@ export class CodeMapArrowService
 		const { edges } = this.storeService.getState().fileSettings
 
 		for (const edge of edges) {
-			if (edge.visible && edge.visible !== EdgeVisibility.none) {
-				const originNode = this.map.get(edge.fromNodeName)
-				const targetNode = this.map.get(edge.toNodeName)
+			const originNode = this.map.get(edge.fromNodeName)
+			const targetNode = this.map.get(edge.toNodeName)
+			if (originNode && targetNode && edge.visible !== EdgeVisibility.none && edge.visible) {
+				//TODO It seems originNode or targetNode might be undefined here,
+				// I think it results from the method being called multiple times when it might not be available yet
+				// I changed that back to avoid console errors and re-enable the edge-metric, however we should investigate why this is happening
 				const curveScale = 100 * this.storeService.getState().appSettings.edgeHeight
 				const curve = this.createCurve(originNode, targetNode, curveScale)
 				this.previewMode(curve, edge.visible)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 -   Mouse cursor flickering #1170
 -   Fix flipping map when clicking any option in the toolbar #1410
+-   Fix edge metric not working correctly
 
 ### Chore 👨‍💻 👩‍💻
 
--- a/visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.spec.ts
+++ b/visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.spec.ts
@@ -190,17 +190,37 @@ describe("CodeMapArrowService", () => {
 			})
 			codeMapArrowService["previewMode"] = jest.fn()
 		})
-		it("should create and edge Preview of one", () => {
+		it("should create an edge Preview of one", () => {
 			const nodes: Node[] = [OUTGOING_NODE]
 
 			codeMapArrowService.addEdgePreview(nodes)
 
 			expect(codeMapArrowService["map"].size).toEqual(1)
 		})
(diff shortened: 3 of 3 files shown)


visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -125,25 +127,17 @@ export class CodeMapArrowService
 
 		for (const edge of edges) {
 			const originNode = this.map.get(edge.fromNodeName)
-			// TODO: Maps should only have valid edges. If that's not the case, the
-			// internal decoration is likely faulty. Check if only test data is not
-			// correct or what the root cause of these checks actually is.
-			if (originNode === undefined) {
-				continue
-			}
 			const targetNode = this.map.get(edge.toNodeName)
```
ce-bo, line 130:
```suggestion
			const targetNode = this.map.get(edge.toNodeName)
			if (!originNode || !targetNode ) {
				continue
			}
```

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -125,25 +127,17 @@ export class CodeMapArrowService
...
-			if (targetNode === undefined) {
-				continue
-			}
-			if (node.has(originNode.path)) {
+
+			if (originNode && targetNode && node.has(originNode.path)) {
```
ce-bo, line 132:
```suggestion
			if (node.has(originNode.path)) {
```

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -125,25 +127,17 @@ export class CodeMapArrowService
...
 				this.addArrow(targetNode, originNode, true)
-				// TODO: Check if the second if case is actually necessary. Edges should
-				// always have valid origin and target paths. The test data is likely
-				// faulty and should be improved.
-			} else if (node.has(targetNode.path)) {
+			} else if (originNode && targetNode && node.has(targetNode.path)) {
```
ce-bo, line 134:
```suggestion
			} else if (node.has(targetNode.path)) {
```

ce-bo (approved):
LGTM :)

ce-bo (approved):

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -125,25 +127,20 @@ export class CodeMapArrowService
 
 		for (const edge of edges) {
 			const originNode = this.map.get(edge.fromNodeName)
-			// TODO: Maps should only have valid edges. If that's not the case, the
-			// internal decoration is likely faulty. Check if only test data is not
-			// correct or what the root cause of these checks actually is.
-			if (originNode === undefined) {
-				continue
-			}
```
BridgeAR, line 133: This code should be kept as it was before :-)
These continue conditions are already the safeguard :-)

BridgeAR (approved):
LGTM. Just left a nit (the code is fine, one refactoring was not necessary though).

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
RomanenkoVladimir: This will throw type errors for targetNode is undefined

RomanenkoVladimir: It seems we need to check for both again here

BridgeAR: This is handled a few lines below :-)

RomanenkoVladimir: sorry I ment originNode, that must have caused confusion
because in that case the next statement if (node.has(originNode.path)) would throw an error because originNode is undefined and therefore can not be asked for a path

RomanenkoVladimir: Therefore we need to check for both in the upper statement, at least with my local testing we get errors otherwise

BridgeAR: That part is never reached due to continue statement? What I meant is: this function should not have caused issues, only the function above.

RomanenkoVladimir: Yes I had another local change in there, 
I just reset all changes and pulled the branch again, so the result from changing 
`if (!originNode || !targetNode )`  to `if (originNode === undefined) ` will lead to type errors in targetNode  in line 137, so the check  `else if (node.has(targetNode.path))` because targetNode is not defined in that case

RomanenkoVladimir: The case where a originNode exists but has no path and we therefore add an arrow to the target node would be the case for nodes that are not the origin node (the ones receiving the arrow, visualized in light blue)
 if I understand the code correctly, which would fail in that case

Issue #1276, opened 2020-09-21
EdgePreview on Map broken when selecting zero


JaiJoPa:
# Bug

## Expected Behavior

<GIVEN>

WHEN I move the edgeMetric preview slider to zero

THEN no edgePreview should be visible on the map

## Actual Behavior

WHEN I move the slider to zero

THEN the map shows all edge previews on the map

## Screenshots

![image](https://user-images.githubusercontent.com/50145538/93750909-8af71380-fbfc-11ea-90df-6291428093d7.png)

## Steps to Reproduce the Problem
1. Select an edgeMetric
2. Move edgePreview slider to zero 

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.57.3
- OS: Mac
- Browser: Chrome

RomanenkoVladimir:
In the online demo this results in an infinte loading bug in firefox

RomanenkoVladimir:
@BridgeAR do we want a slider that goes to 0 and shows no edges or should 1 be the minimum?
The second option would seem more intiuitive to me, as you can toggle the metric off, however its your choice.

ce-bo:
> In the online demo this results in an infinte loading bug in firefox.

>EDIT: any change to the label slider now triggers an infinte loading for me, can anyone confirm it please @ce-bo @JaiJoPa @DmytroHutkin ?

There are TypeErrors present in the Console.
![image](https://user-images.githubusercontent.com/26900540/96848463-c16ebb00-1454-11eb-8643-7c8f0396a280.png)

RomanenkoVladimir:
Yes, I fixed the issue
When I fix the bug in question as well it should all work as intended

RomanenkoVladimir:
I detached my PR from this to release the fix this week, this will be done in a follow up PR

BridgeAR:
I am going to reopen the issue as it's fixed for the zero case but not for other preview numbers.

![image](https://user-images.githubusercontent.com/8822573/100756547-94b3a900-33ed-11eb-8d95-2e3cde9f88eb.png)

ce-bo:
@BridgeAR Could you please give us more details, so that we can understand the current issue better.

PR #1444, opened 2020-10-28
Bug/1276/edge preview broken when selecting zero

RomanenkoVladimir:
#Edge Preview shows all edges when slectiing 0

closes: #1276

## Description

- Fix the bug with all edges being shown when selecting 0 for the amount.

### DIFF
--- a/visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
+++ b/visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
@@ -55,15 +55,15 @@ export class EdgeMetricDataService implements StoreSubscriber, BlacklistSubscrib
 	}
 
 	getNodesWithHighestValue(metricName: string, amountOfEdgePreviews: number) {
+		if (amountOfEdgePreviews === 0) {
+			return []
+		}
+
 		const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
 		const keys: string[] = []
 
 		if (nodeEdgeMetrics === undefined) {
 			return keys
-		}
-
-		if (amountOfEdgePreviews === 0) {
-			return null
 		}
 
 		for (const key of nodeEdgeMetrics.keys()) {
--- a/visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
+++ b/visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
@@ -55,12 +55,13 @@ export class EdgeMetricDataService implements StoreSubscriber, BlacklistSubscrib
 	}
 
 	getNodesWithHighestValue(metricName: string, amountOfEdgePreviews: number) {
+		const keys: string[] = []
+
 		if (amountOfEdgePreviews === 0) {
-			return []
+			return keys
 		}
 
 		const nodeEdgeMetrics = nodeEdgeMetricsMap.get(metricName)
-		const keys: string[] = []
 
 		if (nodeEdgeMetrics === undefined) {
 			return keys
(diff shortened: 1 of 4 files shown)


visualization/app/codeCharta/ui/codeMap/codeMap.actions.service.ts
```diff
@@ -60,7 +60,7 @@ export class CodeMapActionsService {
 		for (const edge of edges) {
 			edge.visible = EdgeVisibility.none
 
-			if (edge.attributes[edgeMetric] !== undefined) {
+			if (edge.attributes[edgeMetric] !== undefined && edgePreviewNodes !== null) {
```
BridgeAR, line 63: A set may never be null :-)

```suggestion
			if (edge.attributes[edgeMetric] !== undefined) {
```

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -62,6 +62,10 @@ export class EdgeMetricDataService implements StoreSubscriber, BlacklistSubscrib
 			return keys
 		}
 
+		if (amountOfEdgePreviews === 0) {
+			return null
+		}
```
BridgeAR, line 67: I guess it's nicer to return an empty array instead? In that case it's possible to combine to condition with the one above.

visualization/app/codeCharta/state/store/metricData/edgeMetricData/edgeMetricData.service.ts
```diff
@@ -62,6 +62,10 @@ export class EdgeMetricDataService implements StoreSubscriber, BlacklistSubscrib
 			return keys
 		}
 
+		if (amountOfEdgePreviews === 0) {
```
ce-bo, line 65: I would place this condition at the top of the method. If I understand it correctly, we don not need to retrieve the `nodeEdgeMetrics` in this case.
Furthermore, I would just return an empty array like in the case the `nodeEdgeMetrics` are `undefined`. Thus, no further changes would be neccessary and we can revert the change here codeMap.actions.service.ts:63.

ce-bo (commented):
Could you please address my remarks? :)

visualization/app/codeCharta/ui/codeMap/codeMap.actions.service.ts
```diff
@@ -60,7 +60,7 @@ export class CodeMapActionsService {
 		for (const edge of edges) {
 			edge.visible = EdgeVisibility.none
 
-			if (edge.attributes[edgeMetric] !== undefined && edgePreviewNodes !== null) {
```
ce-bo: We should check for a Set initialized with an empty [], if the else-branch in line 73 implements a valid behaviour.

RomanenkoVladimir: You mean for correctness? 
Because while we set visibility for to-edges they are empty as well, so no edges are drawn on the canvas.

RomanenkoVladimir: I guess one could argue that the current result is incorrect, because the function shows
both if both exist
to if its the only existing direction
from else

However in an empty case the last transformation would lead to empty, which is not an expected result according to such a definition

RomanenkoVladimir: I deleted my previous comments
The first if would catch an empty set because from == to == [] then the inner if checks it from is [] and only if it isn't it sets visibility

RomanenkoVladimir: Missed that point as well, but now we can be sure there is no issue

ce-bo: Thanks, there is definitely no bug :)

ce-bo (approved):
👍

BridgeAR (approved):

Issue #1269, opened 2020-09-18
File tree/flattened/excluded overlay visualization is buggy


ce-bo:
# Bug
- MaibornWolff logo overlaps with the actual overlay
- I can not scroll the excluded file list, if a lot of files are excluded (especially on small screens)
- The displayed file path of an excluded file is wrong. It seems to be copied from a previous excluded file (see the screenshot).

## Expected Behavior
- The tree/file overlay is placed "in front of" the MaibornWolff logo.
- I can scroll down/up the excluded/flattened file list.
- The displayed file path of an excluded file is the right one.
<GIVEN>

## Screenshots
![cc-bug-tree-overlay](https://user-images.githubusercontent.com/26900540/93598034-5d238c00-f9bc-11ea-9023-61d4bb8cbadd.PNG)

## Steps to Reproduce the Problem
Just exclude a batch of files with "right click -> exclude" (maybe on a small screen)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes

PR #1445, opened 2020-10-28
Bug/1269/flatten-exlude overlay is buggy

RomanenkoVladimir:
# File tree/flattened/excluded overlay visualization is buggy #1269 

closes: #1269 

## Description

  - fixes display issue with wrong file names being shown in side-bar for excluded/flattened nodes
  - shortens path names whne they get too long to avoid breaks in the pathname like "ome/filename" if the path gets too long
  - fixes logo overlap
  - adds scrolling functionality

### DIFF
--- a/visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.ts
+++ b/visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.ts
@@ -6,6 +6,8 @@ import { removeBlacklistItem } from "../../state/store/fileSettings/blacklist/bl
 import { StoreService } from "../../state/store.service"
 
 export class BlacklistPanelController implements BlacklistSubscriber {
+	private MAX_PATH_LENGTH = 45 //we use a set width for the component as well as a set font size -> empirically gathered
+
 	private _viewModel: {
 		flatten: Array<BlacklistItem>
 		exclude: Array<BlacklistItem>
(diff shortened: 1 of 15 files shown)


ce-bo:
I got this error:
image

1) Load the sample1 map
2) Exlude the thin red building
3) Try to exclude the remaining red building. You will likely run into the same error.

RomanenkoVladimir:
Yes, I had a small array error for tracking the excluded buildings, nice catch

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.ts
```diff
@@ -23,6 +25,15 @@ export class BlacklistPanelController implements BlacklistSubscriber {
 		const excluded: BlacklistItem[] = []
 
 		for (const item of blacklist) {
```
ce-bo, line 27: Should we add a small comment to explain what that algorithm does?
I understand it that way: "Cut enough folders of the path from the beginning so that the display_name is short enough.

ce-bo: I have another idea for a possible improvement.
We could add the original file path as a `title="{{path}}"` attribute, so that the full path is still visible if you hover the shortened name.

ce-bo: Removed previous comment. A title attribute is already implemented.

ce-bo (approved):
LGTM.

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.ts
```diff
@@ -23,6 +25,15 @@ export class BlacklistPanelController implements BlacklistSubscriber {
...
+			const path = item.path
+			item.display_path = path
+
+			if (path.length > this.MAX_PATH_LENGTH) {
+				const shortName = path.slice(-this.MAX_PATH_LENGTH)
+				const replaceBeforePosition = shortName.indexOf("/")
+				item.display_path = `..${shortName.slice(replaceBeforePosition)}`
```
BridgeAR, line 34: Instead of actually cutting of parts of the name, let's use the css functionality instead. It allows to display the dots at the correct spot in all cases: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

Issue #1092, opened 2020-07-10
Sort order in tree-view not applied correctly


NearW:
# Bug

## Expected Behavior

When I open the tree-view, then I expect the folders and files to be sorted correctly by name. `ab.ts` should come before `Bc.ts`.

## Actual Behavior

When I open the tree-view, then the folders and files are not sorted correctly by name. `Bc.ts` comes before `ab.ts`.

## Screenshots

https://user-images.githubusercontent.com/23529226/87176436-64525d80-c2da-11ea-9bce-276a51dfeaa6.jpeg

## Steps to Reproduce the Problem
1. Open tree-view in demo

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.50.0
- OS: Mac and Windows 10
- Browser: Chrome

BridgeAR:
To improve this even further: sorting should ideally sort numbers naturally and accent marks take are taken into account. This is easy to implement by using `localeCompare()`: `stringArray.sort((a, b) => a.localeCompare(b, undefined, { numeric: true }))`.

E.g.:

```js
[
  '1',   '2',   '3',
  '10',  '11',  'abb',
  'abc', 'abC', 'Abc',
  'ábc', 'abd', 'bbc',
  'Bbc'
]
```

RomanenkoVladimir:
@BridgeAR PTAL

I think you already implemented that change: current sorting uses custom comparator functions and not sort()
numerical sorting is done via substraction
alphabetical via a.name > b.name, I think this is the correct functionality?

PR #1452, opened 2020-10-30
Bug/1092/Change string sorting 

RomanenkoVladimir:
# Change string sorting implementation for folder/filename #1092

closes: #1092 

## Description
- Change the string sorting from a.name > b.name to a.localeCompare(b), if that is what you meant @BridgeAR?
- While we still use array.sort we call it with custom comparator functions, so it will ignore the standard sorting behaviour defined in the issue
- While we could change the sorting to not relay on custom comparators, I like the current behaviour as it makes dding new sorting options or changing behaviour really easy, and localeCompare is not defined for numbers

Would that consider the issue closed? 
Becuase the bug has been fixed by older changes from @BridgeAR I think we dont need a Changelog entry for this and would just close the issue


visualization/app/codeCharta/ui/mapTreeView/mapTreeView.component.ts
```diff
@@ -33,7 +33,7 @@ export class MapTreeViewController implements CodeMapPreRenderServiceSubscriber,
 		const compareFunction: CompareFunction =
 			sortingOption === SortingOption.NUMBER_OF_FILES
 				? (a, b) => b.attributes[NodeMetricDataService.UNARY_METRIC] - a.attributes[NodeMetricDataService.UNARY_METRIC]
-				: (a, b) => (b.name > a.name ? -1 : 1)
+				: (a, b) => a.name.localeCompare(b.name)
```
BridgeAR, line 36:
```suggestion
				: (a, b) => a.name.localeCompare(b.name, undefined, { numeric: true, sensitivity: 'base' })
```

Ideally we add a default collator while starting the application (`new Intl.Collator([...])`).

visualization/app/codeCharta/util/dataMocks.ts
```diff
@@ -274,6 +274,106 @@ export const VALID_NODE_WITH_MULTIPLE_FOLDERS_SORTED_BY_NAME: CodeMapNode = {
...
+		{
+			name: "Földer",
+			type: NodeType.FOLDER,
+			attributes: { [NodeMetricDataService.UNARY_METRIC]: 60 },
+			isExcluded: false,
+			isFlattened: false,
+			children: []
+		},
+		{
+			name: "Folder2",
```
BridgeAR, line 351: We'll need the following folders: `Folder1`, `Folder2` and `Folder10` to test the names.

BridgeAR (approved):

visualization/app/codeCharta/util/dataMocks.ts
RomanenkoVladimir: Ok, I thought to test for special charachters like "ö" being treated correctly, but I can change that

PR #1490, opened 2020-11-18
Feature/dynamic margin to maximum

RomanenkoVladimir:
# Set dynamic margin to maximum and change slider to absolute values

## Description

- slider set absolute values instead of %
- change dynamic to be maximum by default
- camera should remain intakt, we see the whole map when loading it
- update maths where % were required (we could split model and passed number, this would make it nicer I think, however now its nicely coupled so not sure)
- update tests and snapshots(UI changed after changing default margins)


BridgeAR:
Is it possible to allow values that prevents maps from showing up? And could we use a zoom factor that always shows the "current view", just with a different margin? And is it possible to always use the full view screen for the initial map size, no matter what to what the margin is set too? :-)

RomanenkoVladimir:
We should take another look at the correct initial view point here, closing for now

BridgeAR:
Is this still a draft? :)

visualization/app/codeCharta/ui/codeMap/codeMap.label.service.spec.ts
```diff
@@ -160,9 +160,11 @@ describe("CodeMapLabelService", () => {
 		})
 
 		it("scaling existing labels should scale their position correctly", () => {
+			const { margin } = storeService.getState().dynamicSettings
 			const SX = 1
 			const SY = 2
 			const SZ = 3
+			const SCALE_CONSTANT_LABEL = 15 / 4
```
ce-bo, line 167: Why 15?

visualization/app/codeCharta/ui/areaSettingsPanel/areaSettingsPanel.component.html
```diff
@@ -30,9 +29,9 @@
 		class="md-primary"
 		ng-model="$ctrl._viewModel.dynamicMargin"
 		ng-change="$ctrl.applyDynamicMargin()"
-		title="Use a fit to width margin to see the whole map"
+		title="Set margin to default (50)."
```
ce-bo, line 32:
```suggestion
		title="Set margin to default (50px)."
```

visualization/app/codeCharta/ui/areaSettingsPanel/areaSettingsPanel.component.html
```diff
@@ -30,9 +29,9 @@
...
 	>
-		Dynamic Margin
+		Default margin (50)
```
ce-bo, line 34:
```suggestion
		Default margin (50px)
```
Should we show the unit like this?

PR #1531, opened 2020-12-01
Spike/html tests

shaman-apprentice:
# Refactoring file count in Attribute-Side-bar

Issue: #

## Refactor out component and add tests
- small components are easier to maintain
- small components are easier to test
- add tests about html structure to figure out how they work in angularjs and to test the introduced logic in the html template

## cc-prefix:
- I have learned to prefix own custom elements to distinguish them from other quickly
- I have learned to prefix own custom elements to prevent name clashes.
- It is probably heavily arguable:
  - if we want to introduce it now for every component
  - if we want to introduce it with every component we edit
  - if we want to dismiss this prefix thing
  - I am fine with everything and don't want to start a huge discussion. Just wanted to share the cc-prefix idea ;)


visualization/app/codeCharta/ui/attributeSideBar/ccNodePath/ccNodePath.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	it("should display no file count information if selected building is a leaf / i.e. a file", () => {
+		const component = $compile("<cc-node-path-component></cc-node-path-component>")($rootScope)
+		$rootScope.$emit("building-selected", {
+			node: { isLeaf: true, path: "some/file.ts" }
+		})
+		$rootScope.$digest()
+
+		expect(component[0].getElementsByClassName("cc-node-file-count").length).toBe(0)
+	})
```
ce-bo, line 60: I think we can combine this two tests.

visualization/app/codeCharta/ui/attributeSideBar/ccNodePath/ccNodePath.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+	it("should insert section for file count if selected building is not a leaf / i.e. a folder", () => {
+		const getFileCountDescriptionSpy = jest.spyOn(ccNodePathComponent.controller, "getFileCountDescription")
+		const component = $compile("<cc-node-path-component></cc-node-path-component>")($rootScope)
+		$rootScope.$emit("building-selected", {
+			node: { isLeaf: false, path: "some/file.ts" }
+		})
+		$rootScope.$digest()
+
+		const fileCountDescriptionElems = component[0].getElementsByClassName("cc-node-file-count")
+		expect(fileCountDescriptionElems.length).toBe(1)
```
ce-bo, line 71:
```suggestion
		const fileCountDescriptionElements = component[0].getElementsByClassName("cc-node-file-count")
		expect(fileCountDescriptionElements.length).toBe(1)
```

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.html
```diff
@@ -16,19 +16,7 @@ <h2 class="node-name-h2">
...
-					count="$ctrl._viewModel.packageFileCount"
-					when="{
-						'0': '(empty)',
-						'1': '(1 file)',
-						'other': '({} files)'
-					}"
-				></ng-pluralize>
-			</span>
-		</p>
+		<cc-node-path-component></cc-node-path-component>
```
ce-bo, line 19: For components, I like the prefix idea. :-)

visualization/app/codeCharta/ui/attributeSideBar/ccNodePath/ccNodePath.component.ts
```diff
@@ -0,0 +1,39 @@
+import { IRootScopeService } from "angular"
+
+import { CodeMapBuilding } from "../../codeMap/rendering/codeMapBuilding"
+import { BuildingSelectedEventSubscriber, ThreeSceneService } from "../../codeMap/threeViewer/threeSceneService"
+import { Node } from "../../../codeCharta.model"
+
+class CcNodePathController implements BuildingSelectedEventSubscriber {
```
ce-bo, line 7: The components/controllers are already placed within .../app/codeCharta/ui/....
Hence, I am not really convinced to prefix the Angular Components as well. It would be duplication, right?

ce-bo (commented):
Thanks for evaluating how to test angular html components 👍

visualization/app/codeCharta/ui/attributeSideBar/ccNodePath/ccNodePath.component.spec.ts
shaman-apprentice: From my point of view they test different pieces of logic and should therefore be kept separated. As they take less than 100ms I would keep them separated. Does that makes sense to you?

PR #1565, opened 2020-12-08
Spike/lint only staged changes

shaman-apprentice:
# Lint only staged changes of visualization via precommit

This increases dev-speed and encourages to do more micro commits.

Don't forget to run `npm i` as it introduces one new dev-dependency


RomanenkoVladimir (approved):
Very nice change, thank you.

I'm a bit confused if all of the package-lock changes are necessary, cause those are a lot of new additions, however if we need all of them it looks good to me.

BridgeAR (approved):
LGTM

PR #1662, opened 2021-01-12
Feature/1314/open code files locally

### LINKED ISSUE #1314: Open code files locally
# Feature request

## Description

As a user, I want to open files on my hard drive from the map in my default editor to easy accessing the correct file.

## Acceptance criteria

- All development notes are implemented

## Development notes (optional Task Breakdown)

- [x] Add a lazy loading mechanism that is triggered in case you double click on a file in a map and there's no remote link attached
- [x] Add the same lazy loading mechanism to the filename while having a file or folder open on the top right
- [x] As soon as the mechanism is triggered for the first time, a window is opened that asks for the directory of the file (ideally, a "open folder" window is opened to let the user find the correct folder)
- [x] The filename is now combined with the mentioned directory
- [x] The file is opened in the browser as `file://directory/file/path.extension`.
- [x] The mechanism is not triggered if the directory is already known. Instead, the file is opened as described before
- [x] On the standalone variation, validate that the folder does indeed exist as expected and print an error otherwise
- [x] Store the location of the repository in the local storage connected to the concrete map
- [ ] Update our cc.json to include the original path to the repository during creation. This path should be checked in the standalone variation. If it's still valid, use this value by default without asking the user.
- [x] (Optional, if more work than ~2 hours) The user should be able to update the path in case the folder is moved during development.

On top as another feature:

- [ ] Let the user choose to also add a link as directory instead
- [ ] The links are now all added to the current map
- [x] Double clicking behaves as it did before with links

## Open questions

- Check that the file is opened in the default editor
- We might want to let the user choose if the file should be displayed in the browser or in the default editor

DmytroHutkin:
# {Implement #1314}

Closes #1314 

## Description

Allow opening files from local drive by double clicking on the buildings or clicking on their names
 - only available in a standalone version!
 - files are opened in associated third-party applications
 - web-links are opened in a browser window

## Test Scenario
Use `npm run build && npm start` to run in standalone mode
Either double-click on a node or click once and then click on it's name on top right and enter a path of the project's root directory. You can create a new folder hierarchy with some sample files on your local PC to simulate a project.
The root path is then combined with a path attached to a selected node.
If path is valid, the corresponding file will be opened at once in a third-party app. Otherwise an error message will be shown.

## Open questions
 - It is still preferable to use modalDialogs insted of _window_ ones. Any help is appreciative.


visualization/app/codeCharta/util/lazyLoader.ts
```diff
@@ -0,0 +1,50 @@
...
+		const exists = await LazyLoader.checkDirExists(directory)
+		if (!exists) {
+			alert(`Unknown or non-existent directory: ${directory}`)
+			return LazyLoader.setDirectory()
+		}
+		const open = await import("open")
+		try {
+			await open.default(`file:///${path}`)
+		} catch (error) {
+			alert(error)
```
BridgeAR, line 29: I personally would rather use our internal error handling. @NearW do you know how to use that here?

It's not a blocker though.

BridgeAR (commented):
Mainly LGTM!

visualization/app/codeCharta/util/lazyLoader.ts
RomanenkoVladimir: I think the idea is to throw an error dialog to the user to provide feedback that it couldn't be opened?

DmytroHutkin: > I think the idea is to throw an error dialog to the user to provide feedback that it couldn't be opened?

Right. However, as discussed before, there is a problem with `async await` while using two dialogs (error and prompt) at the same time.

Issue #1555, opened 2020-12-07
Wrong display of colored folders in legend


Gaugi28:
# Bug

## Expected Behavior

The user can see which color fits to which selected folders: 
- Deleted Comma, 
- one line per entry, 
- color icon is at one line with the first entry of this color
- the color updates when the user changes the color

## Actual Behavior

The user can't see which color fits to which selected folders.

## Screenshots
![image](https://user-images.githubusercontent.com/12883714/101371006-8b718300-38aa-11eb-9b3e-c530ecb206b5.png)

## Steps to Reproduce the Problem
1. Mark one folder in blue, and two in pink.
2. Open the legend.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version:
- OS: macOS Catalina 
- Browser: Safari

PR #1668, opened 2021-01-13
Bug/1555/wrong display of colored folders in legend

RomanenkoVladimir:
# Wrong display of colored folders in legend

closes: #1555

## Description
- The feature was already implemented and tested correctly, however we were tracking elements by $index, which means that angular would not redraw the new information ( we specifically told the DOM that way that this component would not change) -> Removed incorrect tracking
- Fixed incorrect commata placement

### DIFF
--- a/visualization/app/codeCharta/ui/legendPanel/legendPanel.component.html
+++ b/visualization/app/codeCharta/ui/legendPanel/legendPanel.component.html
@@ -25,19 +25,15 @@
 	<cc-map-color-picker map-color-for="outgoingEdge" label="Outgoing Edge"> </cc-map-color-picker>
 	<cc-map-color-picker map-color-for="incomingEdge" label="Incoming Edge"> </cc-map-color-picker>
 
-	<hr ng-show="$ctrl._viewModel.packageLists[0]" />
-	<div
-		class="legend-block"
-		ng-show="$ctrl._viewModel.packageLists"
-		ng-repeat="packageList in $ctrl._viewModel.packageLists track by $index"
-	>
+	<hr ng-if="$ctrl._viewModel.packageLists[0]" />
+	<div class="legend-block" ng-show="$ctrl._viewModel.packageLists" ng-repeat="packageList in $ctrl._viewModel.packageLists">
 		<img ng-src="{{::packageList.colorPixel}}" alt="" />
 		<span
 			class="marked-package {{ $first ? '' : 'set-left-margin' }}"
-			ng-repeat="item in packageList.markedPackages track by $index"
+			ng-repeat="item in packageList.markedPackages"
 			title="{{::item.path}}"
 		>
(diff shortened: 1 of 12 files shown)


RomanenkoVladimir:
Concerning tests, how would we test for this functionality? 
We already have unit test, however the values were always correct, the issue was manually hindering update propagandation.

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.html
```diff
@@ -25,19 +25,15 @@
...
-		ng-show="$ctrl._viewModel.packageLists"
-		ng-repeat="packageList in $ctrl._viewModel.packageLists track by $index"
-	>
+	<hr ng-if="$ctrl._viewModel.packageLists[0]" />
+	<div class="legend-block" ng-show="$ctrl._viewModel.packageLists" ng-repeat="packageList in $ctrl._viewModel.packageLists">
 		<img ng-src="{{::packageList.colorPixel}}" alt="" />
 		<span
 			class="marked-package {{ $first ? '' : 'set-left-margin' }}"
-			ng-repeat="item in packageList.markedPackages track by $index"
+			ng-repeat="item in packageList.markedPackages"
```
BridgeAR, line 33:
```suggestion
			ng-repeat="item in packageList.markedPackages track by item.path"
```

BridgeAR (commented):
It's a best practice to track by unique identifiers when iterating over objects. We mostly use the full path as identifier and thus, it's good to use this. Otherwise the whole object is stringified and that takes a while.

See https://docs.angularjs.org/api/ng/directive/ngRepeat#tracking-and-duplicates

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.html
RomanenkoVladimir: I see, thanks
I changed it to avoid full object stringifying

BridgeAR (commented):
LGTM but ideally, we'd add an e2e test to verify the behavior to prevent future regressions :-)

visualization/app/codeCharta/ui/legendPanel/legendPanel.e2e.ts
```diff
@@ -0,0 +1,22 @@
...
+		legendPanelObject = new LegendPanelObject()
+		fileChooser = new FileChooserPageObject()
+
+		await goto()
+	})
+
+	describe("LegendPanel", () => {
+		it("should highlight a folder and add to legend then react to a changed color", async () => {
+			await fileChooser.openFiles(["./app/codeCharta/ressources/sample1_with_different_edges.cc.json"])
+			await legendPanelObject.open()
```
BridgeAR, line 19: This does not yet test the output, does it? (Correct display of the file names)

BridgeAR (approved):
LGTM

Issue #1093, opened 2020-07-10
Don't break line for secondary metrics span


NearW:
# Bug

## Expected Behavior

Let's not break the line for `Secondary Metrics`. Instead get the string on one line like in primary metrics.

## Screenshots

![image](https://user-images.githubusercontent.com/12533626/87210806-2d4e6d00-c317-11ea-96cb-e4cbec35d209.png)

## Steps to Reproduce the Problem
1. Select a building

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.50.0
- OS: Windows and Mac
- Browser: Chrome

PR #1672, opened 2021-01-15
Fix: remove unnecessary break line for secondary metrics #1093

Shivan-997:
# Don't break line for secondary metrics span

Closes: #1093 

## Description
The words are forced into one line.

## Screenshots or gifs
Screenshot 2021-01-15 at 12 55 48


RomanenkoVladimir:
LGTM, need to quickly fix the merge conflict tho
Small nit, could you change Issue: to closes: # number of your PR so it will automatically close on Zenhub (in the PR request)

RomanenkoVladimir (approved):

BridgeAR (changes requested):
The reason for the line break is that the table uses a single column. Instead, we could just combine the column over the whole table width. That way there's no extra line break either and the left column would keep the current size.

BridgeAR (approved):

Issue #1686, opened 2021-01-20
Labels placed incorrectly on delta maps that share no common nodes


RomanenkoVladimir:
# Bug

## Expected Behavior

<GIVEN>
A map in delta mode where the two maps in comparison have no common nodes

WHEN

loading the map and drawing the initital label

THEN

the labels have the correct height (normalized to 1000 as height = 500, deltaHeight = 500)

## Actual Behavior

The height is calculated to early resulting in huge, raw values.

## Screenshots
![incorrectDeltaHeight](https://user-images.githubusercontent.com/25777197/105184797-0c6d9c80-5b30-11eb-909a-558743e03c3b.PNG)

## Steps to Reproduce the Problem
1. Open the current web visualization
2. Switch to delta mode and generate delta between the two maps

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes | no
- CodeCharta Version: 1.66.0
- OS: Windows 10
- Browser: Mozilla Firefox/ Google Chrome

PR #1687, opened 2021-01-20
Bug/1686/labels placed incorrecly on delta maps that share no common nodes

RomanenkoVladimir:
# Fix incorrect label placement for delta maps that share no common nodes

closes: #1686 

## Description

The issue seems to be that when the two maps share no nodes the deltaHeight was calculated too early, before it is scaled down to 500(the highest height is always 500 in CodeCharta). I moved it from the render service, because when accesing it directly we could run into ThreeJS crashes due to the values not being avalible.

## Screenshots or gifs
correctDeltaHeight

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Fixed 🐞
 
--   fixed wrong max tree map visibility ([#1624](https://github.com/MaibornWolff/codecharta/issues/1624))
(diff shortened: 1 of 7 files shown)


BridgeAR (commented):
Does this still work as expected while showing multiple labels at a time? :-) E.g., 10 labels? It seems like the implementation would now use different "stages" depending on what label was first in the list and if it's a low label, it would be lower than later ones that are placed higher?

RomanenkoVladimir:
No, the list is always sorted, starting with the biggest element.
The stages are only relevant for the case that we draw a temporary label.

BridgeAR (approved):
RSLGTM

ce-bo (approved):
LGTM - I tested it locally :)

Issue #1578, opened 2020-12-11
Excluding in delta mode is broken


BridgeAR:
# Bug

## Expected Behavior

GIVEN two maps in delta mode

WHEN excluding parts

THEN the excluded part is not visible anymore

## Actual Behavior

The excluded part is still visible

## Screenshots
![image](https://user-images.githubusercontent.com/8822573/101919779-43fd3680-3bcb-11eb-8101-494e419f84d9.png)

## Steps to Reproduce the Problem
1. Try to exclude any data

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.63.0

RomanenkoVladimir:
This issue only exists when excluding buildings via the search bar and not using a star "*".
Whne excluding "*.ts" for example the map is excluded correctly. Same for excluding by nodes/folders manually.
Do you think we should also exclude all folders when not using a *selector(this also doesn't work in single mode currently).

RomanenkoVladimir:
If this isn't the issue this seems to have been fixed already, @BridgeAR could you take another look?

RomanenkoVladimir:
Addition: Blacklist/Flatten tracking by index is bugged as well, we should adress this issue as part of the PR, now that the scope is changed to address *(wildcards).

PR #1703, opened 2021-01-27
Excluding in delta mode is broken #1578

Jacob-Kirimi:
(Inclusion and exclusion of files using wildcard searches in the flatten and exclude operations)

### DIFF
--- a/visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
+++ b/visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
@@ -15,10 +15,7 @@
 			</div>
 		</md-list-item>
 
-		<md-list-item
-			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] track by $ctrl._viewModel.flatten.indexOf(item)"
-			title="{{::item.path}}"
-		>
+		<md-list-item ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path']" title="{{::item.path}}">
 			<md-button class="round-button" ng-click="$ctrl.removeBlacklistEntry(item)">
 				<em class="fa fa-minus-square" title="Remove list item"></em>
 			</md-button>
@@ -47,10 +44,7 @@
 			</div>
 		</md-list-item>
 
-		<md-list-item
-			ng-repeat="item in $ctrl._viewModel.exclude | orderBy: ['path'] track by $ctrl._viewModel.exclude.indexOf(item)"
-			title="{{::item.path}}"
-		>
+		<md-list-item ng-repeat="item in $ctrl._viewModel.exclude | orderBy: ['path'] " title="{{::item.path}}">
 			<md-button class="round-button" ng-click="$ctrl.removeBlacklistEntry(item)">
 				<em class="fa fa-minus-square" title="Remove list item"></em>
 			</md-button>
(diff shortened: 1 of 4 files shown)


BridgeAR (commented):
This is already good work! The part that we have to unify is more though. Check `getNodesByGitignorePath` in `codeMapHelper.ts`. I am not certain if the `isPatternBlacklisted` function works as expected as a wildcard should not match a path, but it might just be just the variable name that confuses me?

RomanenkoVladimir:
Three small things:
- Could you use the PR template, it looks a bit nicer?
- Please inlude a closing keyword, aka: closes #1578 in  your PR so it will auto close on PR merging.
- Could you try tracking by item.path, or a similar construction, instead of removing the tracking entirely to improve performance?

Jacob-Kirimi:
@BridgeAR 
The getNodesByGitignorePath is only called during the search and it recognises the wildcard .ts and *.ts for example 
The problem comes in that when excluding only the *.ts is recognised as a wildcard. The way I have understood it is that 
in the exclude,flatten and  search operations, whether it is a file name or a wildcard search , it will be passed or held in the path field of  a BlacklistItem and hence it is available in the Blacklist array.

Jacob-Kirimi:
@RomanenkoVladimir 
Thanx for the corrections. With regards to the track by Index construction, I can take alook if there is a workaround where we can still maintain it. do you have an idea?

RomanenkoVladimir:
The previous construction was using 
'orderBy: ['path'] track by $ctrl._viewModel.exclude.indexOf(item)"'
I would try orderBy: ['path'] track by '$ctrl._viewModel.exclude.indexOf(item.path)"'

RomanenkoVladimir:
Other small nit: prettier with Changelog.md is not correct please revert and format correctly

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
```diff
@@ -16,8 +16,8 @@ <h3 class="header-text">Flattened</h3>
 		</md-list-item>
 
 		<md-list-item
-			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] track by $ctrl._viewModel.flatten.indexOf(item)"
-			title="{{::item.path}}"
+			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path']"
```
RomanenkoVladimir, line 19: should be ` track by $ctrl._viewModel.flatten.indexOf(item.path)"` here,

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
```diff
@@ -16,8 +16,8 @@ <h3 class="header-text">Flattened</h3>
...
+			title="{{::item.path}} track by $ctrl._viewModel.exclude.indexOf(item.path)"
 		>
```
RomanenkoVladimir, line 21:
```suggestion
			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path']"
			title="{{::item.path}} track by $ctrl._viewModel.flatten.indexOf(item.path)"
		>
```

Jacob-Kirimi: I will include the track by without the title.. Thanx

Jacob-Kirimi:
@BridgeAR I have repaired the exclusion Issue. The only remaining issue is the formatting of CHANGELOG.md producing multiple conflicts which we will a look at with Mehdi first thing tommorow.

Issue #1606, opened 2020-12-18
Delta mode showing no changes


BridgeAR:
# Bug

## Expected Behavior

GIVEN a map in delta mode

WHEN switching the two maps

THEN the map indicates the differences between the two maps

## Actual Behavior

The map indicates no changes at all

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/102594148-19086a80-4116-11eb-8e59-38117bd81372.png)

## Steps to Reproduce the Problem

1. Load the default maps
1. Switch maps

![image](https://user-images.githubusercontent.com/8822573/102594335-5836bb80-4116-11eb-9035-0d957af6dc56.png)
![image](https://user-images.githubusercontent.com/8822573/102594375-65ec4100-4116-11eb-9f2f-56dca0710866.png)
![image](https://user-images.githubusercontent.com/8822573/102594407-73093000-4116-11eb-8f81-36238966cf94.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.63.0

PR #1705, opened 2021-01-29
Fix/1606/switching maps in delta shows no change

JaiJoPa:
# Switching maps in Delta Mode

fixes: #1606 

## Description

- This PR fixes the bug, that the differences between two maps were not shown when switching maps.
- Switching the maps did not show any differences because the reference map remained the same while the comparison map became the reference map. So we compared the same maps in delta.
- This happened because there was a wrong allocation of the reference map when changing the origin reference map.

## Screenshots or gifs

switchMaps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Fixed 🐞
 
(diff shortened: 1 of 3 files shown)


BridgeAR (approved):
❤️

RomanenkoVladimir (approved):

PR #1720, opened 2021-02-03
Fixed Exclusion problem in Delta  #1578

Jacob-Kirimi:
# Excluding in Delta Mode is broken

Issue: 1578

Exclusion/Inclusion of some searches was not working particularly in delta and single mode. and finally their removal from the exclude and flatten lists


visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
```diff
@@ -16,8 +16,8 @@ <h3 class="header-text">Flattened</h3>
 		</md-list-item>
 
 		<md-list-item
-			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] track by $ctrl._viewModel.flatten.indexOf(item)"
-			title="{{::item.path}}"
+			ng-repeat="item in $ctrl._viewModel.flatten | orderBy: ['path'] "
+			title="{{::item.path}} track by $ctrl._viewModel.flatten.indexOf(item.path)"
```
BridgeAR, line 20: The `track by` should be part of `ng-repeat` :-) similar below

Jacob-Kirimi: putting it in "ng-repeat" brings in the previous problem of uncertain behaviour during the removal of items from the flatten and exclude lists. I suggest we leave it out.. :)

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
```diff
@@ -48,7 +48,7 @@ <h3 class="header-text">Excluded</h3>
 		</md-list-item>
 
 		<md-list-item
-			ng-repeat="item in $ctrl._viewModel.exclude | orderBy: ['path'] track by $ctrl._viewModel.exclude.indexOf(item)"
+			ng-repeat="item in $ctrl._viewModel.exclude | orderBy: ['path']"
```
RomanenkoVladimir, line 51:
```suggestion
			ng-repeat="item in $ctrl._viewModel.exclude | orderBy: ['path'] track by item.path"
```

RomanenkoVladimir (dismissed):
Track by does not work by indexOf(item), it will create an index itself. The suggestion should fix the incorrect displaying problem.

Other: Could we maybe move the search functionality to another file as to avoid multiple implementations? Or a similar bug would be repeated if we would implement another component that relays on user search-strings tomorrow.

Nit: Please change Issue: #yournumber to closes/fixes in the PR.

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
Jacob-Kirimi: I willl look into it and give you a feedback.

Jacob-Kirimi:
Track by now seems to be working, Im checking the e2e tests now

BridgeAR (commented):
I just had another look at this and it seems like it's possible to simplify the code quite a bit while making it more robust. I suggest to check `getNodesByGitignorePath`: it's logic should mainly be mimicked and it could be abstracted into two functions. One that just plainly returns the excluded and flattened part and another one that filters. It would pretty much be the `unifyWildCard` function as it's declared here.

We can also try to do some pair programming, if you like.

Jacob-Kirimi:
Sorry  @BridgeAR Im just now seeing the comment. I just finished writing the tests. Yeah I would be great to do pair programming
I havent understood though exaclty coz the functions are in diferent places.

visualization/app/codeCharta/ui/blacklistPanel/blacklistPanel.component.html
Jacob-Kirimi: Done :)

BridgeAR:
@Jacob-Kirimi all good, I sent you an invite for tomorrow.

Issue #1605, opened 2020-12-18
Move to github actions


BridgeAR:
# Tech request

## Description

Travis has a low bandwidth free for open source projects and as such we frequently wait for Travis to finish.
There's also a change in the amount of runs we are allowed to have as open source project.
As such, instead of continuing using travis, let's use github actions.

https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions#:~:text=Travis%20CI%20lets%20you%20set,step%2C%20job%2C%20or%20workflow

## Acceptance criteria
- Github actions run
- Travis is fully removed
- Sonar keeps on working

PR #1730, opened 2021-02-04
Tech/1605/move to GitHub actions

RomanenkoVladimir:
# Move to github actions

closes: #1605

## Description

- this PR removes travis and associated scripts from the project
- all previous functionality moved to github-actions
- sonar -analysis is run on branch-basis showing the expected changes in code-duplication and coverage in regards to main
- split release and push-requests into separate workflows

Known Issues:
- npm install keeps failing at nw-builder, however this is not a failure of github-actions but on the nw-side

Limitations:
- not tested: deployment to docker, release of gh-pages, last part of the npm release, aka releasing the packets itself(version and login work correctly)
- using third-party for upload as the official action is no longer supported
- using third-party for docker deplyoment due to convenience
- script to create gh-pages not inlined to keep a clearer code structure
- E2E - Tests are not included in sonar-coverage, however would be easy to change after changing the sonar configuration (beyond the scope of the issue)


RomanenkoVladimir:
After some consideration I think that running a sonar analysis on every change to every branch might be overkill.
The last change would limit github actions to only run on any change to master as well as changes to branches that have an open PR only.
What do you think?

Med-H:
looks pretty good.

RomanenkoVladimir:
> 
> 
> looks pretty good. how fast is the deployment vs travis ?

We need 8 minutes, from recevining to providing output(see Actions Tab up top) travis needs around 20 in my experience(although it can be a lot more if we have a lot of requests). 

Right now the pure workload is a bit slower, 8 for github actions to 7:30 minutes for travis, however this is due to the added analysis factor. Pure testing is faster by about 2 minutes and the request handling is much better as well.

RomanenkoVladimir:
Also sometimes travis fails to execute all three tasks in parallel, which makes it so slow, this is not the case for github actions.
Once recevied 8-9 minutes should be our worst case.

.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,158 @@
...
+    - name: Install Node
+      run: npm install
+      working-directory: ${{env.working-directory}} 
+
+    - name: Recache-Node     
+      if: failure()
+      run: |
+        npm cache clean --force
+        npm cache verify
+      working-directory: ${{env.working-directory}}
```
BridgeAR, line 45: This seems rather surprising to me? Do we need this part?

.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,158 @@
...
+        uses: SonarSource/sonarcloud-github-action@v1.5
+        with:
+          projectBaseDir: ./visualization
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+      
+      
+
```
BridgeAR, line 158:
```suggestion

```

.github/workflows/release_gh_pages.yml
```diff
@@ -0,0 +1,114 @@
+name: Release Gh-Pages
+
+on: 
+  push: 
+    tags: '1.*'
```
BridgeAR, line 5: We should likely be careful with this tag: if we bump the version, we would not release the gh pages anymore?

BridgeAR (approved):
There's quite a bit of redundancy in our script that could likely be unified and improved. It should also be possible to speed up our builds if we combine some steps but it's not a blocker. We should just have a look at these things again.

.github/workflows/github-actions.yml
RomanenkoVladimir: I will combine a bit more, thats a good point
Concerning the marked line, it was a work around for the nw-builder failure as simply caching it would result in the build failing again when it was restarted manually. We can remove this line now

RomanenkoVladimir: Concerning redundancy, we felt that it would make the jobs more modular as every job would handle his own problem, independent of others.
But maybe that is not the correct philosophy for this task.

.github/workflows/release_gh_pages.yml
RomanenkoVladimir: Yes, that is a problem
However, we could not find another way to handle releases, a release event needs to be explicitly triggered, as we do farther below, and in our case a release is simply a tagged push to main.

For testing purposes we used tags of the form v1.0... and so on, which allowed us to use "v*" wildcards. We could change it to '*', but this felt more representative of a tag.

.github/workflows/github-actions.yml
BridgeAR: If I understood correct, the whole re-cache part may be removed?

BridgeAR:
@RomanenkoVladimir I decided to merge this as is. It would be great if you could follow-up on my comments so that they are not lost :)

RomanenkoVladimir:
> @RomanenkoVladimir I decided to merge this as is. It would be great if you could follow-up on my comments so that they are not lost :)

Yes, I will open a follow-up PR later in the evening

PR #1739, opened 2021-02-05
Rollback nw to 0.50.3 version

RomanenkoVladimir:
… last update

# Rollback nw to 0.50.3 version

## Description
The current update results in multiple failed builds in our pipelines, lets roll it back for now
Also formatting of the package-lock.json is broken on main, see other branches having the same formatting change, lets fix this as well.


visualization/package.json
```diff
@@ -1,174 +1,174 @@
...
+	"window": {
+		"icon": "app/icon.png",
+		"toolbar": true,
+		"frame": true,
+		"width": 1024,
+		"height": 768
+	},
+	"main": "dist/webpack/index.html",
+	"build": {
+		"nwVersion": "0.49.0",
```
BridgeAR, line 39: We might want to update this as well :-)

RomanenkoVladimir: Oh, you are right
I'm not sure if that is what causes the issue in the first place now
That would explain why there are no mentiones about this on their official github page?

RomanenkoVladimir: So we just have too much of a version difference between the build and the used package and thats what makes it fail, could that be possible?

RomanenkoVladimir: @BridgeAR wdyt?

RomanenkoVladimir: Ok setting both to 0.51 did not solve the issue, I guess we need to roll back then

visualization/package.json
```diff
@@ -1,174 +1,174 @@
...
+		"ignore": "^5.0.2",
+		"jquery": "^3.3.1",
+		"klona": "^2.0.4",
+		"lodash.camelcase": "^4.3.0",
+		"lodash.debounce": "^4.0.8",
+		"lodash.isequal": "^4.5.0",
+		"lodash.throttle": "^4.1.1",
+		"md5": "^2.3.0",
+		"nouislider": "^14.6.0",
+		"nw": "^0.50.3",
```
BridgeAR, line 111: Is there no change in our package-lock necessary?

RomanenkoVladimir: no, it seems like it was never changed to the new version to begin with
package-lock is still at 0.50.3, I think the builder update was kind of erroneous

BridgeAR (approved):

PR #1759, opened 2021-02-11
Fix unicorn issues

BridgeAR:


visualization/app/codeCharta/model/files/files.helper.spec.ts
```diff
@@ -90,9 +90,9 @@ describe("files", () => {
 		})
 
 		it("should return the first fileState found if multiple files match the fileName", () => {
-			files.concat([{ file: TEST_DELTA_MAP_A, selectedAs: FileSelectionState.None }])
+			const expectedFiles = [...files, { file: TEST_DELTA_MAP_A, selectedAs: FileSelectionState.None }]
 
-			const result = getFileByFileName(files, TEST_DELTA_MAP_A.fileMeta.fileName)
```
BridgeAR, line 95: This is an actual fix, the test should have failed without this change. I guess the test does not work as expected?

visualization/app/codeCharta/util/usageDataTracker.ts
```diff
@@ -248,7 +248,7 @@ export function trackEventUsageData(actionType: string, state: State, payload?:
 	}
 }
 
-function buildEventTrackingItem(actionType: string, payload?: any): EventTrackingItem | null {
```
RomanenkoVladimir, line 251: This is much more concrete as the previous version, will this not limit current/further functionality?

BridgeAR: We should never use `any`. Otherwise there's no point in using TypeScript :)

If we change a type, we have to make sure it works as expected everywhere and that's only possible by passing through the expected type.

visualization/app/codeCharta/model/files/files.helper.spec.ts
RomanenkoVladimir: Yes,I don't understand the behaviour either 
We should mark it to investigate

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.spec.ts
```diff
@@ -478,19 +479,19 @@ describe("nodeContextMenuController", () => {
 	describe("onBodyLeftClickHideNodeContextMenu", () => {
 		it("should not hide if a click on color-picker occurs", () => {
 			const broadcastHideEventSpy = jest.spyOn(NodeContextMenuController, "broadcastHideEvent")
-			const mockedMouseEvent = {
+			const mockedMouseEvent: any = {
```
RomanenkoVladimir, line 482: why do we define and not cast to any?

visualization/app/codeCharta/util/nodePathHelper.ts
```diff
@@ -26,5 +26,5 @@ export function getParent<T>(hashMap: Map<string, T>, path: string): T {
 		if (node) {
 			return node
 		}
-	} while (path !== CodeChartaService.ROOT_PATH && path.length)
+	} while (path !== CodeChartaService.ROOT_PATH && path.length > 0)
```
RomanenkoVladimir, line 29: how did this work in the past, seems like we never checked correctly?

visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.spec.ts
BridgeAR: I tried to specify the concrete type but it was difficult to so, so I went back to any. Defining something is always better than casting it, since we might otherwise override things that we did not intend to do (with any there's no difference but with other types there might be).

visualization/app/codeCharta/util/nodePathHelper.ts
BridgeAR: It is identical, it just looks nicer.

RomanenkoVladimir (approved):
LGTM

visualization/app/codeCharta/util/nodePathHelper.ts
RomanenkoVladimir: Ah ok, I didn't know that thx

Issue #674, opened 2019-08-09
Export current view as .png and make background transparent


NearW:
# Feature request

Since we usually screenshot views of the map and include it in PP-Presentations, it would be nice to have a button that does that for you and makes the background transparent.
## Description

*When* clicking a button to export the view as .png,

*then* ...
- the viewcube should be hidden,
- the center-map-button should be hidden,
- the toolbar should be hidden,
- the mouse cursor should be hidden,
- ~~the legend-panel should be hidden,~~
- ~~the logo should be hidden,~~ 
- change the background to default,
- the current view should be exported as a .png-file
- ~~make a screenshot of the canvas with minimum size that fits the map + some offset~~
- replace all pixels with default-background-color with transparent one

The button can be placed next to the download button with this [icon](https://fontawesome.com/v4.7.0/icon/camera)

Example for such a screenshot with offset:
![Screen Shot 2019-08-09 at 14 40 11](https://user-images.githubusercontent.com/12533626/62779480-a57f3380-bab3-11e9-8269-93891735441b.png)

BridgeAR:
This comes in handy in combination with custom configs. This is in fact one of the necessary steps to automatically export all screenshots.

BridgeAR:
The map should only optionally be resized though and the logo should be kept as is.

BridgeAR:
The legend should also not be hidden by default. It might be configured somewhere to do so.

PR #1765, opened 2021-02-12
Export current view as .png

Med-H:
# Export current view as .png and make background transparent

closes: #674 

#decription 

Right now just click on the link marked in yellow
image


Med-H:
I'm not sure if it fullfills the requirements

ce-bo (commented):
LGTM :)
It would be great, if we could export the PNGs by a keyboard shortcut. 
So you could higlight folders by mouse and make a screenshot of that.

RomanenkoVladimir (approved):

Issue #1783, opened 2021-02-17
visual artifacts on maps with low building height


Med-H:
# Bug

<GIVEN>

WHEN opening a map with low building heights

THEN there should be no visual artifacts

## Screenshots
actual :
![image](https://user-images.githubusercontent.com/74670211/108214710-c73d8a00-7130-11eb-8fe8-c987b4a40130.png)

expected:
![image](https://user-images.githubusercontent.com/74670211/108214938-08359e80-7131-11eb-8eb3-334a7ccd04dd.png)

## Steps to Reproduce the Problem
1. open firefox json file

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.68.0
- OS: windows 10
- Browser: chrome

Med-H:
#1534 should have fixed it

PR #1784, opened 2021-02-17
Fix visual artifacts on map with low building height

Med-H:
# fix visual artifacts on map with low building height

closes: #1783 

## Description

added per camera position height adjustment in vertex shader

## Screenshots or gifs
(image is based on the sharpened version)
image


Med-H:
no unit tests have been added.

visualization/app/codeCharta/ui/codeMap/rendering/boxGeometryGenerationHelper.ts
```diff
@@ -138,6 +138,18 @@ export class BoxGeometryGenerationHelper {
 		}
 	}
 
+	private static isTopSide(side: number, node: Node) {
```
ce-bo, line 141: I guess this function returns an array to indicate which vertice is on a top side. 
What do you think about another name like `determineTopSideVertices`

visualization/app/codeCharta/ui/codeMap/rendering/boxGeometryGenerationHelper.ts
```diff
@@ -164,31 +177,35 @@ export class BoxGeometryGenerationHelper {
 				uvs[intermediateIndexBL],
 				color,
 				subGeomIndex,
-				deltaRelativeToHeight
+				deltaRelativeToHeight,
+				isTopSide[0]
```
ce-bo, line 181: Does it make sense to use the enum from line 6 for the side index?
`sides.left` in this case.

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
```diff
@@ -41,11 +42,27 @@ void getDirectionalDirectLightIrradiance(const in DirectionalLight directionalLi
...
 void main() 
 {
-    vec4 worldPosition = modelMatrix * vec4(position, 1.0);
+    vec3 adjustedHeightPos = position;
+    // height position will be adjusted based on camera position in order to prevent 
+    // z-fighting even in far away zoom levels when position is positive.
+    // preferred multiplication over if branching for the isHeight variable 
+    // the single line could be written this way : 
+    //
+    // adjustedHeightPos = position
```
ce-bo, line 51: I would just use the more extensive if-statement instead of the one-liner. So we can remove the comment and don't have to adjust the comment and the code itself on every change.

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
```diff
@@ -6,6 +6,7 @@ attribute vec3 defaultColor;
 attribute vec3 deltaColor;
 attribute highp float subGeomIdx;
 attribute highp float delta;
+attribute lowp float isHeight;
```
ce-bo, line 9: Where does it come from? Is it a boolean?

ce-bo (commented):
Some unit tests would be really great. :)

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
Med-H: Sure, I just don't like branching branching in the GPU side as it might lead to more computation. There should be no performance issue in this case. I will change it and remove the comments.

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
Med-H: This attribute is sent from the geometry building function. It is a float number, as no Boolean attributes can be sent to the shaders.

Med-H:
> Some unit tests would be really great. :)

It would be hassle 4 now, as most files don't have unit tests. I would prefer if it is done in another pr.

visualization/app/codeCharta/ui/codeMap/rendering/boxGeometryGenerationHelper.ts
Med-H: I'm unsure if the enum would match the numbering, I will further investigate it

ce-bo: I think it wouldn't make much sense. Imagine: `isTopSide(sides.Left)`
So it's good to keep the current implementation.

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
ce-bo: Ah ok, I didn't know that. :)

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
ce-bo: Thanks for the explenation. I thought that isHeight is a `number[]` coming from `isTopSide(x)`. Multiplying it with a scalar wouldn't make much sense.

ce-bo:
> > Some unit tests would be really great. :)
> 
> It would be a hassle 4 now, as most related files don't have unit tests. I would prefer if it is done in another pr.

Hm, at least the `isTopSide(x)` in `boxGeometryGenerationHelper` looks very easy to test. :)

visualization/app/codeCharta/ui/codeMap/rendering/codeMapVertexShader.glsl
Med-H: should I leave it as it is or change it ? It's up to you :)

ce-bo: Just as you like. ;) I am fine with both solutions.

Med-H:
closing it as it is incorporated in image sharpening pr.

Issue #1785, opened 2021-02-17
Fix e2e Visualisation 'Node is either not visible or not an HTMLElement'


Med-H:
# Bug

## Expected Behavior

e2e test should always pass

## Actual Behavior
at least this example will fail 

   > FAIL app/codeCharta/ui/fileChooser/fileChooser.e2e.ts (42.877 s)
      FileChooser › should keep the old map if opening a file was cancelled
      Node is either not visible or not an HTMLElement

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes | no
- CodeCharta Version: 1.68.0
- OS: windows 10
- Browser: chrome

JaiJoPa:
fixed

PR #1786, opened 2021-02-17
testing the root cause of the e2e test failure

Med-H:
# testing the root cause of the e2e test failure

Issue: #1785


BridgeAR (commented):
This is definitely the correct approach! It is in fact the same as documented here https://github.com/MaibornWolff/codecharta/blob/6872a31a8624d4a150fa61dd5684a5b58e090c61/visualization/app/puppeteer.helper.ts#L6-L8

Thus, we should remove the `{ timeout: 3000 }` in all of the `.toClick()` calls. And please also update the documentation here: https://github.com/MaibornWolff/codecharta/blame/main/gh-pages/_posts/dev-guide/2020-07-31-e2e-testing-with-puppeteer.md#L122

LGTM otherwise! :-)

BridgeAR (approved):

BridgeAR:
@Med-H thanks a lot for fixing the tests!

PR #1803, opened 2021-02-22
Feature/1650/change dependabot update policy

### LINKED ISSUE #1650: Change dependabot update policy
Please update our dependabot settings to use a new versioning-strategy. We should only receive a PR if a change to our `package.json` is necessary, not if it only changes the `package-lock.json`

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#versioning-strategy

Jacob-Kirimi:
change dependabot update policy

closes #1650

This branch merges the creation of a new file config.yml, That is used to test if PRs will be created as per the described format. If it works we can change it to allow for automerging updates that do not affect the package.json.
Affter merge I will create the branch again as it is required for testing purposes.


BridgeAR (dismissed):

ce-bo (approved):

.dependabot/config.yml
```diff
@@ -0,0 +1,11 @@
+version: 1
```
BridgeAR, line 1: It should actually be version 2: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

.dependabot/config.yml
```diff
@@ -0,0 +1,11 @@
...
+
+update_configs:
+   - package_manager: "javascript"
```
BridgeAR, line 4: I guess we want npm here?

BridgeAR (changes requested):
It should actually update the existing dependabot config file in the .github folder :)

.dependabot/config.yml
Jacob-Kirimi: Using the dependabot.yml, one cannot automerge updates that do not affect package.json file. The option to automerge can be achieved using the config.yml that has an option to automerge  updates that affect only the package-lock.json and create PRs for version bumps that affect the package.json.  see https://dependabot.com/docs/config-file/

.dependabot/config.yml
Jacob-Kirimi: I beleive javascript means updating javascript dependencies. In this way dependabot will take a look at package.json and package-lock.json. This is the only way to select it and one cannot specify npm.

.dependabot/config.yml
BridgeAR: The documentation from dependabot.com is the old version 1. But dependanbot moved to github and won't be developed further as version 1 :-)

If I remember correct, automerging is done by other github actions when using dependabot version 2. The merging itself is not an issue for now but only what changes we get. We only want updates that are related to the package.json instead of the package-lock.json.

Jacob-Kirimi: sorry my bad.. I will read the documents again to see how we can achieve that using version 2.

Jacob-Kirimi: @Ruben Hoping that I have not missed anything, I think dependabot does not offer this posibility as per now. 
The versioning strategy documentation is not really very clear about
using 'widen' or 'increase' and I beleive it does not guarantee that the PRs of updates that only affect the package.jsonfile will be created, rather it will be in addition to those that are in package-lock.json.
I suggest we try  setting the versioning strategy to widen or increase using a target branch and monitor the pull requests that will be generated.

BridgeAR:
Closing as won't fix due to the limitation on side of dependabot

Issue #1728, opened 2021-02-04
Decrease gpu utilization


Med-H:
# Feature request

As a user , I want my gpu to render on demand.

## Description

The current rendering is done even when no changes are made, which consumes my GPU ressources and battery.

## Acceptance criteria

- [ ] rendering should only be done when moving the mouse, changeing parameters , ... 
- [ ] no performance drop when using on large maps

PR #1806, opened 2021-02-22
Feature/1728/on demand rendering

Med-H:
# ondemand rendering

Issue: #1728 

## Description

  - With this PR the gpu utilization will be near 0% when in idle mode
  - The render request animation has been removed in favor of manual update when needed

## Screenshots or gifs

without gpu on demand
image

with gpu on demand
image


BridgeAR:
Seems like the prettier check does not check the exit code properly. Some files have the different line ending again :)

BridgeAR (commented):
Mainly LGTM. It does feel like we could easily miss something though and we have to be very careful to really rerender everything properly. Thus, I would like to have others closely review this, especially by trying it out in all kinds of situations.

Going forward is also more difficult with this solution as we have to pay attention to call update manually. It would be good to rework the internals in a way that this is all done for us automatically.

visualization/app/codeCharta/ui/codeMap/codeMap.component.spec.ts
```diff
@@ -85,6 +98,34 @@ describe("ColorSettingsPanelController", () => {
...
+			expect(threeViewerService.autoFitTo).toHaveBeenCalled()
+		})
+
+		it("should call threeViewerService animate", () => {
+			expect(threeViewerService.animate).toHaveBeenCalled()
+		})
+
+		it("should call threeViewerService animateStats", () => {
+			expect(threeViewerService.animateStats).toHaveBeenCalled()
+		})
```
ce-bo, line 126: Small idea (nit): We could do that in one test function by iterating over an array of the function names and add the expects automatically.

ce-bo (approved):
LGTM but I had not the time to test it locally.

visualization/app/codeCharta/ui/codeMap/codeMap.component.spec.ts
Med-H: sure :)

visualization/app/codeCharta/ui/codeMap/codeMap.preRender.service.ts
```diff
@@ -99,6 +99,8 @@ export class CodeMapPreRenderService
 		) {
 			this.debounceRendering()
 			this.debounceTracking(actionType)
+		} else {
+			this.codeMapRenderService.update()
```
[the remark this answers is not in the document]
Med-H: I think debouncerendering does set a new map mesh etc, while this.codeMapRenderService.update() is only redrawing the frame.

ce-bo:
I tested it locally and it works. However, I experienced a little performance issue when showing a lot of labels and hovering them. Not sure, if this is caused from the new changes.

visualization/app/codeCharta/ui/codeMap/codeMap.preRender.service.ts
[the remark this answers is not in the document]
BridgeAR: But do we need to redraw anything in case there's no update to the map? Why would we trigger the function otherwise?

visualization/app/codeCharta/ui/codeMap/codeMap.component.spec.ts
```diff
@@ -85,6 +98,29 @@ describe("ColorSettingsPanelController", () => {
...
+		]
+		beforeEach(() => {
+			codeMapController.onSharpnessModeChanged()
+		})
+
+		for (const test of sharpnessTests) {
+			it(`should call threeViewerService ${test.func}`, () => {
+				expect(threeViewerService[`${test.func}`]).toHaveBeenCalled()
+			})
+		}
```
BridgeAR, line 117: I personally would have used a single `it` and used a for loop inside that, but that's just my personal style and not blocking.

BridgeAR (approved):

Issue #1807, opened 2021-02-22
enforce simple linebreaks breaks linter on windows


Med-H:
linter is broken under windows when using "linebreak-style": ["error", "unix"]

Med-H:
not needed

PR #1820, opened 2021-02-23
fix broken linter under windows

Med-H:
# Broken linter under windows

Issue: #1807

## Description

unix `linebreak-style` linter option breaks windows linter. with this pr, the line-break depends on the operating system.


Med-H:
no need as it can be configured under the editor settings

PR #1821, opened 2021-02-23
Tech/1795/enforce line endings 3

BridgeAR:
This contains the editorconfig, the gitattributes file and an improved changelog while also fixing prettier issue detection on our CI runs and makes sure files in python are saved with \n instead of \r\n.


.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
```
ce-bo, line 6: I am not sure. Do we use indent_size of 4 at the moment?

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+quote_type = double
+
+[*.kt]
+indent_size = unset
+quote_type = unset
```
ce-bo, line 14: What was the reason for unsetting those values? :)

.editorconfig
BridgeAR: We use ktlint for kotlin files and the indentation is not 2 but partially 3 and 4.

.editorconfig
BridgeAR: I was not sure anymore if the quotes are different due to ktlint. It's not necessary to define it here, as ktlint is going to complain about them anyway, if they are wrong.

ce-bo (approved):

PR #1849, opened 2021-03-03
ci: use deploy token with more privileges and prevent recursive builds

NearW:
# Use token with more privileges to deploy gh-pages

## Description

- Prevent recursive triggering of pipelines triggered by master push
- Use a dedicated deploy token for gh-pages provided by the tech-user

## ToDo
- Before merging this, we need a token named `DEPLOY_TOKEN` in our secrets created by the codecharta-user with the priviliges to push to a protected branch.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -10,13 +10,13 @@ on:
 
 jobs:
  check_prettier:
+    if: "!contains(github.event.head_commit.message, 'Deploying to main')"
     name: Prettier Test
     runs-on: ubuntu-latest
     env:
       working-directory: ./
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


BridgeAR (approved):
I am not super happy with the duplication and the check itself is brittle (checking for a string could cause issues later on) but it seems to solve the problem for now.

NearW:
@BridgeAR The duplication is required because you only do conditions on each job, not for the whole flow.

Another slightly better option would be to configure the gh-pages-deploy-action to use a custom commit message containing a prefix of `[ ci-skip ]` that we can check for as a string. This will prevent the workflow from breaking in case they'd change their default commit message.

Also, we still need the DEPLOY_TOKEN in our secrets.

PR #1882, opened 2021-03-19
Fix auto approve issue

BridgeAR:
This should fix the `Resource not accessible by integration` issue coming up with dependabot auto approval. This is now triggered as soon as the `Tests` actions are done and should have the correct privileges. I am not certain about the latter though.

### DIFF
--- a/.github/workflows/auto-approve-and-merge.yml
+++ b/.github/workflows/auto-approve-and-merge.yml
@@ -1,29 +1,32 @@
 name: Auto approve and merge PRs by dependabot
 
 on:
-  pull_request
+  workflow_run:
+    workflows: ["Tests"]
+    types:
+      - completed
 
 jobs:
   autoapprove:
     name: Auto Approve a PR by dependabot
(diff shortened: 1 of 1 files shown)


RomanenkoVladimir:
According to the following discussion the only change required is changing pull_request to pull_request_target.

Discussion in question: https://github.com/hmarr/auto-approve-action/issues/183

The mentioned issue popped out in the most recent version update it seems and the change is the approved solution by the actions creator.

I'm not sure if we wont run into privilidge issue using the changes in this PR

BridgeAR:
My suggestion is based upon https://securitylab.github.com/research/github-actions-preventing-pwn-requests. It was not completely clear how the rights where set in the example but to me it seemed the solution as implemented here would be sufficient.

RomanenkoVladimir:
After taking a closer look at the article you provided I agree, it should have the required permissions in our case.

Should it happen that it fails for some reason we could use pull_request_target and restrict it to a specific label, e.g. "dependencies" which was another solution proposed in the article.

Issue #1966, opened 2021-04-28
Sonarcloud action always fails on first attempt


NearW:
# Bug

## Expected Behavior

GIVEN a triggered dependabot pipeline

WHEN the sonarcloud task runs for the first time

THEN it succeeds 

## Actual Behavior

GIVEN a triggered dependabot pipeline

WHEN the sonarcloud task runs for the first time

THEN it fails no matter what

BridgeAR:
This should also be useful for renovate.

PR #2015, opened 2021-05-20
Proposal: Allow inbuild dependabot to read secrets again(see comment)

RomanenkoVladimir:
# Dependabot cant read secrets following last update

closes: #1966

## Description

- According to this discussion:  https://github.com/dependabot/dependabot-core/issues/3253 and this migration guide: https://hugo.alliau.me/2021/05/04/migration-to-github-native-dependabot-solutions-for-auto-merge-and-action-secrets/#share-your-secrets-with-dependabot the following PR will allow dependabot to read secrets again 

- However: This represents a workaround: This proposal uses on.pull_request_target for dependabot PRs which is less secure (we can read all secrets like this). When github-actions eventually implemenents a working solution we will have to change that again if the PR is accepted. 
- Concerning dependabot secrets (new option in Settings -> Secrets): I tried it, as did people in the linked discussion and it is not working properly as of today.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
     - main
-  pull_request:
+  pull_request_target:
     branches:
     - main
 
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -4,9 +4,12 @@ on:
   push:
     branches:
     - main
-  pull_request_target:
+  pull_request:
     branches:
     - main
+  pull_request_target:
+      branches:
+      - main
 
 jobs:
  check_prettier:
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -1,10 +1,7 @@
 name: Tests
 
 on:
-  push:
-    branches:
-    - main
-  pull_request_target:
+  pull_request_target
 
 jobs:
  check_prettier:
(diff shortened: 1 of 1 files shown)


RomanenkoVladimir:
This is a proposal, I think in the current project and given the differentiation to only use pull_request_target on dependabot PRs we minimize the risk a bit but its still a problem:

https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-842471128

and we should decide whether or not it is worth the risk in our case.

As for other solutions: Pushing empty commits/ restarting workflows by hand solves the issue but is a very ugly way to do it.
We could also migrate back away from the inbuild solution for now, while we wait for a proper implementation by github

RomanenkoVladimir:
I took another look and found out that it is indeed connected to the automerge/autoapprove issue
We are faced with the same problem, according to this comment by the dependabot author: 
https://github.com/dependabot/dependabot-core/issues/3253

I add a short update to that in a moment

RomanenkoVladimir:
Hm, so seems like this is more of a problem then anticipated
I would really like to avoid running auto-merges as a fork due to security reasons, also there is some discussion about it breaking the pipeline due to missing secrets on main.
Furthermore our automerge action is rather outdated, what aobut using: "https://github.com/ahmadnassri/action-dependabot-auto-merge" instead (this would however require a fork context)
Another option would be to try "https://github.com/koj-co/dependabot-pr-action", they seem to get around the issue, but I would need to investigate some more

BridgeAR:
@RomanenkoVladimir are you certain the latter works? Their own dependabot PRs are not merged anymore?

RomanenkoVladimir:
Closing this PR due to the solution being too insecure and superseeded by https://github.com/MaibornWolff/codecharta/pull/2018

Issue #353, opened 2019-02-15
Dragging/Rotating the map does not work when mouse is near viewcube


mrpatpat:
# Bug

Dragging/Rotating the map does not work when mouse is near viewcube. The right events need to be propagated to the codemap and/or the three orbit controls

## Expected Behavior

Dragging the map near the viewcube rotates the map.

## Actual Behavior

Dragging the map near the viewcube does nothing.

NearW:
https://stackoverflow.com/questions/46667197/three-js-orbitcontrols-not-working-when-dom-element-is-positioned-over-the-scene

RomanenkoVladimir:
@DmytroHutkin this currently breaks any user interactions PTAL

PR #2028, opened 2021-05-31
Fix/353/map rotation not working near cube

Shivan-997:
# map rotation not working near cube

closes: #353 

## Description
  - Dragging/Rotating the map does not work when mouse is near viewcube.


RomanenkoVladimir (approved):

RomanenkoVladimir:
still needs fixing

PR #2053, opened 2021-06-20
fix: wsl pre-hook

shaman-apprentice:
otherwise sh is used despite "#!/usr/bin/env bash" which has no fancy if "[["


ngormsen (approved):
LGTM

PR #2065, opened 2021-06-23
Feature/1954/loading map should not replace old ones

### LINKED ISSUE #1954: Loading a new map should add that map to known ones instead of replacing the former one
# Feature request

## Description

As a user, I want to keep old maps so that I safe time when comparing maps in multiple mode.

## Acceptance criteria
- Loading a new map does not remove already existing maps anymore
- The new maps are the only "active" maps (as implemented currently)
- The map dropdown list receives a garbage bin icon to remove loaded maps to safe memory

ngormsen:
# Loading new files does no longer replace old ones.

Closes #1954 

## Description
    -   Opening new files does no longer remove old ones
    -   Already loaded files can be individually removed.
    -   The 'Multiple' view will select the latest files.


RomanenkoVladimir (commented):
Talked about a few smaller changes in pairing.
LGTM else

PR #2079, opened 2021-06-30
Fix building selection #2070

### LINKED ISSUE #2070: Selection of a building doesn't open its sidebar
# Bug

## Expected Behavior

When I click on a building the side bar with its metrics is opened

## Actual Behavior
Nothing happens

## Screenshots

image

1. Checkout main
2. `npm ci && npm run dev`
3. Click on any building of the default map

## Hints
- reverting to before https://github.com/MaibornWolff/codecharta/commit/95e18e2ef92265c497ee46b7e88f8452665736f0, it works fine

ngormsen:
# Restore correct behaviour when clicking on a building

closes #2070 

## Description

Quick fix as suggested by @shaman-apprentice in #2070 to restore the correct behaviour when clicking on a building.

I will open a separate issue for adding tests to rule out the bug in the future.


ngormsen:
Newly created issue for adding tests to avoid the bug in the future: #2080

BridgeAR (commented):
Thanks for the quick fix. I understand that the intention is to fix the issue immediately and then add tests. The issue has not been released though and as such I would rather have a test added right away.

RomanenkoVladimir:
I would advise against this change if possible
updateHovering will call a lot of functions in cascading function, which would be really expensive
what update introduced this bug?

RomanenkoVladimir (approved):

PR #2099, opened 2021-07-07
Feature/1520/show max value instead of infinite

### LINKED ISSUE #1520: Label on Code Coverage shows infinite instead of 100
# Max value of line_coverage shows as 'infinite' instead of '100'

## Legend should show e.g. >67 to 100
Select line_coverage as one of the dimensions of the charta charta

## Actual Behavior
Legen shows 'infinite' instead of 100

## Screenshots
Legend

See above

ngormsen:
# Replace infinite in legend panel with max metric value

Closes #1520 

## Description

  - Replace infinite with the same max metric value that is shown in the threshold slider
  - Some minor visual improvements as mentioned in the issue

## Screenshots or gifs

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Added 🚀
+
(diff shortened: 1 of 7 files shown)


RomanenkoVladimir (commented):
LGTM, however we extended the issue to include the color metric top bar as well, to avoid potential confusion
Also please add a Changelog entry

RomanenkoVladimir (approved):
LGTM, however we have a lot of duplication between colorSettingsPanel and legendPanel, while keeping each component as separate as possible is important we currrenly need to update information twice, as noticed in the first review, which results in an increased errror probability.
Maybe we should refactor this in the future?

PR #2100, opened 2021-07-08
Fix/1312/calculate metrics only once

### LINKED ISSUE #1312: Only calculate metrics once
# Bug

## Expected Behavior

GIVEN a map

WHEN it's loading or the user e.g., blacklists something

THEN the metrics are calculated once

## Actual Behavior

The metrics are currently calculated six times during loading and this is one of the main performance issues.

1. Profile
1. Load a map
1. Check the profile

## Specifications
- CodeCharta Version: 1.57.4

ngormsen:
# Calculate metrics only once

Closes #1312 

## Description
 
- The metrics are currently calculated six times during loading and this is one of the main performance issues as noted in #1312.
- With debounce the metrics are only calculated once which reduces the time of `loadFiles` with a big map like [Apache OpenOffice](https://maibornwolff.github.io/codecharta/showcase/aoo/) from 4.5 seconds to 2.6 consistently during my tests.
- This seems like an easy fix. If anyone sees any possible problems that could occur by introducing this change please let me know.

## Screenshots or gifs
image
image

Sorry for the screenshot quality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Added 🚀
 
--   Custom metric scenarios which include the 'Color-Metric' (rloc) will now also save any changes made to the color scheme.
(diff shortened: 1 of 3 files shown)


RomanenkoVladimir (approved):
Great workaround, although we will need to fix this properly by adressing our event handling in the future

PR #2298, opened 2021-07-14
Refactor/migrate first component

### LINKED ISSUE #2270: Migrate /app/codeCharta/ui/sortingButton/sortingButton.component.ts

### LINKED ISSUE #2290: Enable dual bootstrapping of AngularJS and Angular
As a Developer I want to be able to partial replace AngularJS through Angular, so that:
- We always have a working version
- Main and migration work don't diverge
- Work can be parallelized

For that we need to dual bootstrap AngularJS and Angular.

### LINKED ISSUE #2291: Nice way of connecting Angular to Redux
There is currently no nice way to connect an Angular component to Redux. Using [angular-redux-connect](https://github.com/rynrn/angular-redux-connect) we still need to manually add and type properties, what is unnecessary boiler code and adds potential error sources.

The goal is to have an utility like [react-redux's connect](https://react-redux.js.org/api/connect) with strong types.

### LINKED ISSUE #2293: Improve tests for new Angular components
As a developer I want tests with no false positive nor false negative. Current test approach of AngularJS controllers has the following flow described by the example of **app/codeCharta/ui/sortingButton/sortingButton.component.spec.ts**:

## False positive
If its template would e.g. bind `ng-click="$ctrl.onButtonClickWithTypo()"` the tests would still be green.

## False negative
If we would e.g. rename `onButtonClick` to `onClick` the tests would fail, as they try to directly call implementation detail `onButtonClick`.

# Solution: Utilize @testing-library/angular
[@testing-library/angular](https://testing-library.com/docs/angular-testing-library/intro/) is an approach to avoid exactly that flaw. The goal of this ticket is to integrate @testing-library/angular for new Angular components.

shaman-apprentice:
# Migrate first component (sortingButton) for real

Closes #2270 
Closes #2290
Closes #2291
Closes #2293

This PR migrates the sortingButton.component.ts (#2270) as first migrated component and introduces:
- Dual booting of AngularJS and Angular (#2290)
- A nice way for connecting Angular to redux store (#2291)
- [@testing-library/angular](https://testing-library.com/docs/angular-testing-library/intro/) (#2293)


visualization/app/codeCharta/ui/sortingButton/sortingButton.component.spec.ts
```diff
@@ -1,57 +1,23 @@
...
-	}
+	it("should toggle SortingOrderAscending on click", async () => {
+		await render(SortingButtonComponent)
+		const initialSortingOrder = sortingOrderAscendingSelector(Store.store.getState())
+		expect(initialSortingOrder).toBe(false)
 
-	function rebuildController() {
-		sortingButtonController = new SortingButtonController($rootScope, storeService)
-	}
+		const button = screen.getByTitle("Toggle sort order (currently descending)")
```
NearW, line 17: move title to variable.

use `screen.getByRole("button")` instead

visualization/app/codeCharta/ui/sortingButton/sortingButton.component.ts
```diff
@@ -1,34 +1,18 @@
...
-		SortingOrderAscendingService.subscribe(this.$rootScope, this)
-	}
-
-	onSortingOrderAscendingChanged(sortingOrderAscending: boolean) {
-		this._viewModel.orderAscending = sortingOrderAscending
-	}
+const ConnectedClass = connect((state: CcState) => ({ orderAscending: sortingOrderAscendingSelector(state) }), { setSortingOrderAscending })
 
+@Component({
+	selector: "cc-sorting-button",
```
NearW, line 11: Why the `cc-` prefix? use `sorting-button` or `sorting-button-component` instead

visualization/package.json
```diff
@@ -118,13 +125,17 @@
 		"shelljs": "^0.8.4",
 		"three": "^0.126.1",
 		"three-orbit-controls": "^82.1.0",
-		"typescript-json-schema": "^0.50.0"
+		"typescript-json-schema": "^0.50.0",
+		"zone.js": "^0.11.4"
```
NearW, line 129: is this required?

NearW (approved):

visualization/app/codeCharta/ui/sortingButton/sortingButton.component.ts
shaman-apprentice: if I recall correctly product / company prefix is best practice for custom elements. They have to be unique and it "avoids" name clashes in case of multiple custom elements from different providers. It is also less verbose than component and ensures that each component has a dash, which is required.

visualization/package.json
shaman-apprentice: yes, zone.js is a peer dependency, which needs to be loaded before Angular.

PR #2309, opened 2021-07-19
tech: switch to node 16

### LINKED ISSUE #2301: Synchronize / update Node.js version
[@types/nodes](https://github.com/MaibornWolff/codecharta/blob/main/visualization/package.json#L136) points to v15. But our ci uses v12 [here](https://github.com/MaibornWolff/codecharta/blob/main/.github/workflows/release_gh_pages.yml) and [here](https://github.com/MaibornWolff/codecharta/blob/main/.github/workflows/test.yml). CI and developers should use the same node version, to avoid different node / npm behaviour.

As further improvement we should use [npm ci](https://docs.npmjs.com/cli/v7/commands/npm-ci) over `npm install` in pipeline. Otherwise the build pipeline might run with different packages / package-lock.json, than present in the repository, as possible changes by `npm install` in package-lock.json are not commited by the pipeline.

The goal should be to use node v16, which will become LTS soon.

shaman-apprentice:
closes #2301


NearW (changes requested):
LGTM in general, but we can enforce a node version through the `package.json` by specifying engines. We should do that in all of our `package.json` files

see https://stackoverflow.com/a/29349773

NearW (approved):

BridgeAR:
In general: ideally, we stick to the latest Node.js LTS version. It seems like this might have broken our `plop` scripts. We could add some tests to the scripts as well to detect these issues.

Issue #2352, opened 2021-08-30
Create guideline "how-to-changelog"


BridgeAR:
# Feature request

## Description

As a developer I want to know how to write changelogs so that the changelog entries are written in a user friendly way that is also presentable in our "whats new" section.

## Acceptance criteria
- Create a documentation that outlines how, where and when to create changelog entries (technical)
- Make sure to include what the changelog entry should contain:
  - a PR / issue link
  - a comprehensive description
  - screenshots of changed functionality
  - ...

PR #2358, opened 2021-09-06
Add Changelog Guidelines #2352

IhsenBouallegue:
# Add Changelog Guidelines

Issue: #2352 

## Description

  - The changelog entries do not follow any pattern and are sometimes confusing.
  - Add changelog guidelines for the developers to help further clarify how and when to add to the changelog file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ To unify the appearance of all commit messages we only accept commit messages us
 
 ###How to write a good description?
 
--   The description should be precise and short. Please keep it in one sentence (Max. 2).
+-   The description should be precise and short to provide the user with all necessary information. If it is needed, add some precise notes about the usage of the new feature.
 -   One changelog entry should describe one change.
 -   Avoid writing technical descriptions
 -   Start with a verb in the present tense
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,3 +79,44 @@ To unify the appearance of all commit messages we only accept commit messages us
 -   Name the PR like its branch name (e.x. `Tech/123/my branch name`). The name is usually created by GitHub automatically
 -   Add the correct labels
 -   The PR Assignee is only used by the reviewer to see who is reviewing it
+
+## Changelog Guidelines
+
+###Why should you we write a changelog?
+
+-   A changelog is vital for the developers to keep track of their work. But most importantly now that it appears on every version update to every user, we should make sure that it is also user-friendly. And for that, we need guidelines to help us.
+
+###When should you add to the changelog file?
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -79,3 +79,44 @@ To unify the appearance of all commit messages we only accept commit messages us
...
+-   [Description of what you changed] [Link to the pull request]
+    [One picture (specify the width and/or height)]<br/>
+    Example:<br/>
+    `- Add a changelog dialog with the latest additions to CodeCharta that appears on version update ([#1315](pull-request-linkl)) <img src="image-link" width="350px">`
+    The image should always have a width and/or a height attribute.
+    Example: width=”500px”
+
+###How to write a good description?
+
+-   The description should be precise and short. Please keep it in one sentence (Max. 2).
```
ce-bo, line 105: Ruben suggested a while ago to describe the behavior of (for example) a new feature in detail. I like this idea and personally would keep it instead of writing just one sentence :)

CONTRIBUTING.md
```diff
@@ -79,3 +79,44 @@ To unify the appearance of all commit messages we only accept commit messages us
...
+
+## Changelog Guidelines
+
+###Why should you we write a changelog?
+
+-   A changelog is vital for the developers to keep track of their work. But most importantly now that it appears on every version update to every user, we should make sure that it is also user-friendly. And for that, we need guidelines to help us.
+
+###When should you add to the changelog file?
+
+-   You should only add a changelog entry after the merge (or approval) of your feature/fix not before.
```
ce-bo, line 91: Normally, we develop new features and/or bug fixes in separate feature branches. So we can add the changelog entry directly in that branch. Only after the merge of the feature, the changelog entry will appear in the main branch.

IhsenBouallegue: I realized that after writing it. I fixed it.

CONTRIBUTING.md
IhsenBouallegue: That makes sense from a developer point of view. But as a user, I want a straight to the point description. What I mean is the changelog can give you a quick glance on something. If one wants to understand the details, they can click on the link for the Pull Request/Issue/documentation. That's what I noticed at least from researching changelogs of other projects.

ce-bo: Personally, as a User I would be happy about tipps regarding the usage of the new feature. Therefore I wouldn't restrict the changelog description to be as short as possible. What do you think about something like this: "The description should be precise and short to provide the user with all neccessary information. If it makes sense, add some precise notes about the usage of the new feature."

IhsenBouallegue: That sounds good! Okay I will change it to that

ce-bo (approved):

Issue #2326, opened 2021-07-30
Screenshot improvements


BridgeAR:
# Feature request

## Description

As a user, I want to see all parts of the view on my screenshot so that I have the same information on the screenshot as I do when having a open map.

## Acceptance criteria

- ~~Add a new hotkey to save the screenshot to the clipboard instead of saving it to a file.~~
- ~~Add a section in the global configuration to allow changing default screenshot behavior (allowing to switch between different output behaviors - either saving to file or to clipboard)~~
- ~~Add a option to add the logo to the screenshot output and use that as new default behavior~~ (a separate [issue](#1990) exists for that part)

- Add the legend to the screenshot output (will be done in #3080 

## Development notes (optional Task Breakdown)

Please add one PR per acceptance criteria, building one feature on top of the other.

BridgeAR:
Please be careful about PRs closing issues that they should not :-) // @Shivan-997 @ce-bo

ce-bo:
I changed the scope of this issue to only provide the legend in the screenshot output. This is the most important thing to do.
The other requirements are out sourced to #3080.

ce-bo:
- On Download: Download two screenshots (one for map, one for legend).
- Rename existing copy function to "Copy Map to Clipboard"
- Add new function "Copy legend to clipboard".

ce-bo:
Closed as completed. Adding the legend to screenshot will be done in #3080.

PR #2369, opened 2021-09-13
Feature/2326/add legend to the screenshot output

Shivan-997:
# add legend to the screenshot output

Issue: #2326

## Description

Add a option to add the legend to the screenshot output and use that as new default behavior


Shivan-997:
was created for testing purposes

Issue #2372, opened 2021-09-14
Flaky E2E Tests


ce-bo:
Global Settings - Change Display Quality Test is flaky so that development is blocked.

![image](https://user-images.githubusercontent.com/26900540/133271799-8513dbd9-9126-4a33-a781-994d5fca912a.png)

ce-bo:
Let's discuss regarding PR #2382

PR #2381, opened 2021-09-21
Spike: Apply slowMotion to stabilize test #2372

ce-bo:
# Stabilize E2E Tests by enabling slow motion

Issue: #2372 

## Description
By setting `slowMo` to a low value, e2e tests seem to run more stably.
This PR could fix our current problems with flaky e2e tests, but this does not address the root cause at all.

The question is, why it works with a slow motion of 25ms!


ce-bo:
Outdated by #2382

Issue #2377, opened 2021-09-20
Improve/Provide metric descriptions in legend


ce-bo:
# Feature request
As a UX Engineer, I want to prevent empty metric descriptions in the legend, so that Users of CodeCharta are not supposed to be confused.

## Acceptance criteria
- The metric label is not shown, if no description is available.
- More basic metric descriptions are provided. Especially for `loc` and `comment_lines`
- 
- 

![image](https://user-images.githubusercontent.com/26900540/133969279-9124e89d-2002-4143-957d-237b38c00a34.png)

PR #2421, opened 2021-10-04
Fix/2377/improve metric descriptions in legend

jannikr:
# Improve metric descriptions in legend

closes: #2377 

## Description

-The metric label is not shown, if no description is available.
-More basic metric descriptions are provided. Especially for loc and comment_lines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Fixed 🐞
+
+-   Hides legend block if label description is not available and adds new metric descriptions ([#2377](https://github.com/maibornwolff/codecharta/issues/2377)).
+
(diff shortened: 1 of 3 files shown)


ce-bo (commented):
Thanks for the PR and the fix.
Since I see the solution now, I would rather hide the following part `: <empty description>`. The label should be still visible, even if no description is available. Thus, the legend would provide information of the choosen metrics all the time.

Would you please so kind and change this?

jannikr:
@ce-bo I've been thinking about that, too. I just wonder then what added value the legend brings. The selected metric can also be seen in the bar at the top of the screen.

ce-bo:
@jannikr That's true. For the case, that we take a screenshot, we need the metric information to be included in the legend.

Shivan-997 (approved):

Shivan-997:
Notice: do not Squash and merge. there is some changes coming.

jannikr:
@ce-bo Please note my new commit.
legend_improvement
The screenshot shows the behavior when descriptions are missing.

ce-bo (approved):
Thank you @jannikr 👍 
I think you could have done it by applying the `ng-if` to the span. This would be less "html-verbose".

PR #2458, opened 2021-10-18
Refactor/2318/migrate map tree view

shaman-apprentice:
**Be aware that folder now always stay on the top, no matter how sorted. If sorted by files count, the file order will not change anymore, when changing sorting order ascending / descending**


ce-bo:
@BridgeAR I like that the folders are not placed at the bottom anymore when the sort order is changed. Would you like to keep the original behaviour or the improved one?

Issue #2434, opened 2021-10-07
Suspicious Metrics feature performs calculations even if experimental mode is off


ce-bo:
# Bug

Suspicious Metrics feature performs calculations even if experimental mode is off. On top of that, the calculations are really expensive, so that we should do the calculations only if experimental mode is turned on.

PR #2471, opened 2021-10-29
Fix/2434/no-calculations-of-suspicious-metrics-when-experimental-mode-disabled

Hall-Ma:
# Running calculations of suspicious metrics only if experimental mode is turned on

Closes: #2434 

## Description

- Fixes unwanted calculations of the suspicious metrics feature when experimental mode is disabled
- adds condition when calculations shall be performed


visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -103,6 +103,12 @@ export class ArtificialIntelligenceController implements FilesSelectionSubscribe
 	}
 
 	private calculate() {
+		const { experimentalFeaturesEnabled } = this.storeService.getState().appSettings
```
ce-bo, line 106: This won't work: If the web application is loaded with a clear local storage, then the experimental features are disabled by default. Turning them on in the global settings will not change the value of the constant `experimentalFeaturesEnabled `. Instead, please subscribe the state by `ExperimentalFeaturesEnabledSubscriber`.

ce-bo (commented):
Let's write a regression test for this fix.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
ce-bo: I am sorry, I was wrong! You're totally right that you always have the current enabled/disabled value.
My (manual) test failed, because the calculation has not been triggered on activating the experimental features.

That means, keep the if as it is and we just have to subscribe the `ExperimentalFeaturesEnabledSubscriber` and call the `debounceCalculation()` method.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Fixed 🐞
 
+-   Fix unwanted calculations of suspicious metrics when experimental features are disabled ([#2434](https://github.com/MaibornWolff/codecharta/issues/2434))
```
ce-bo, line 12:
```suggestion
-   Fix unwanted calculations of suspicious metrics when experimental features are disabled ([#2471](https://github.com/MaibornWolff/codecharta/pull/2471))
```

ce-bo (approved):

Issue #2318, opened 2021-07-27
 Migrate all Angular Components and Services


shaman-apprentice:
This is a living todo list. Feel free to resort or add items to it. If you start with an item please move it to the top and add your name behind it. If necessary, we can discuss in the comments section further steps or create further tickets. 

- [x] ./codeCharta/ui/sortingButton/sortingButton.component.ts (#2298) (@shaman-apprentice )
- [x] ./codeCharta/ui/metricDeltaSelected/metricDeltaSelected.component.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/mapTreeView/mapTreeView.level.component.ts (@shaman-apprentice )
- [x]  ./codeCharta/ui/mapTreeView/mapTreeView.component.ts (@shaman-apprentice )
- [x] ./codeCharta/state/store/appSettings/sortingOrderAscending/sortingOrderAscending.service.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/attributeSideBar/nodePath/nodePath.component.ts (@IhsenBouallegue )
- [x] ./codeCharta/ui/matchingFilesCounter/matchingFilesCounter.component.ts (@shaman-apprentice )
- [x] ./codeCharta/state/store/dynamicSettings/searchedNodePaths/searchedNodePaths.service.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/sortingOption/sortingOption.component.ts (@shaman-apprentice )
- [x] ./codeCharta/state/store/dynamicSettings/sortingOption/sortingOption.service.ts (@shaman-apprentice )
- [x] ./codeCharta/state/nodeSearch.service.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/attributeSideBar/attributeSideBar.component.ts (@shaman-apprentice )
- [x] ./codeCharta/state/store/appSettings/secondaryMetrics/secondaryMetrics.service.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/attributeTypeSelector/attributeTypeSelector.component.ts (@shaman-apprentice )
- [x] ./codeCharta/ui/legendPanel/legendMarkedPackages/legendMarkedPackages.component.tsx  (@shaman-apprentice )
- [x] ./codeCharta/ui/customColorPicker/storeColorPicker/storeColorPicker.component.ts  (@shaman-apprentice )
[...]

PR #2472, opened 2021-10-30
Refactor/2318/migrate sorting option component

### LINKED ISSUE #2107: Evaluate use of Angular Material
# Do we need Angular(JS) Material?

Currently we use [AngularJS Material](https://material.angularjs.org/latest/), which cannot be easily migrated. We have the following two options:

1. Utilize [Angular Material](https://material.angular.io/)
2. Simply use some own CSS / Sass

## Advantages of Angular Material
- Theming (do we need it?)
- Accessibility (do we need it?)
- Responsive Design (do we need it?)
- It comes with pre-build components like `mat-select`, directives like `cdkPortal`, ect

## Disadvantages of Angular Material
- New framework to learn
- Additional dependency (to maintain and build time)

### LINKED ISSUE #2473: Toolip of sorting order shows wrong sorting direction
image

With the migration of the map-tree-component the sorting algorithm (which is already in main but not released) this was already fixed. As an unwanted side effect the results are now sorted initially descending instead of ascending. 

Therefore we need to make the default sorting order ascending.

shaman-apprentice:
refs: #2318
Resolve #2107 
Close #2473


shaman-apprentice:
Sonar is red due to "less than 80% test coverage" - all uncovered lines are `NgModule` := import overhead and can safely be ignored. So I declare the pipeline as green 😎

jannikr (changes requested):
If I understand correctly, the Sorting Option function is responsible for sorting the project folders in the File/Node Explorer. However, I noticed here that sorting by name does not work.
sorting-option

shaman-apprentice:
> If I understand correctly, the Sorting Option function is responsible for sorting the project folders in the File/Node Explorer. However, I noticed here that sorting option has a different behavior. sorting-option
> 
> **Setup:** Node v16.12.0 MacOs using Chrome v95.0.4638.54

Good catch. This is currently a bug in released version, that it displays wrong title. I will change the tooltip instead of the sorting order and fix with it #2473.

jannikr (approved):

PR #2495, opened 2021-11-09
Refactor/2318/migrate node serach service

shaman-apprentice:
Store should only contain unique atomic data. Searched nodes path can be derived from searchPattern. Therefore the migration of `nodeSearch.service` was to extract the logic into a selector and remove it 😎


CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/)
 
-## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
+## [unreleased] (Added 🚀 | Changed | Deprecated | Removed 🗑 | Fixed 🐞 | Security)
+
+### Changed
+
+-   Chore: Remove `nodeSearch.service.ts` and `searchedNodePaths` from store as they can be derived from `searchPattern`.
```
ce-bo, line 12:
```suggestion
## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)

### Chore 👨‍💻 👩‍💻

-   Remove `nodeSearch.service.ts` and `searchedNodePaths` from store as they can be derived from `searchPattern`.
```



Personally, I would keep the categories as they are. A removing change will most probably affect the UI only, so `Removed` would be the more appropriate one. However, I like your suggestions to add `Deprecated` to cover it for future cases. Thanks!

Let's revert this change and follow the current format. I will open a separate PR, because I know that the "What's new" feature has to be adjusted as well to match the new categories. We should improve this feature, so that new/removed changelog categories will automatically be discovered. Sorry for the constant back and forth.

ce-bo: Here is the new issue for further discussions/implementation #2497

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Chore 👨‍💻 👩‍💻
+
+-   Remove `nodeSearch.service.ts` and `searchedNodePaths` from store as they can be derived from `searchPattern`.
```
IhsenBouallegue, line 12: Shouldn't that be under Removed?
And also there should be a link to the PR (or Issue)

IhsenBouallegue (commented):
Awesome Work 😎

CHANGELOG.md
shaman-apprentice: Removed is for a removed Feature. For the user everything behaves the same. It is an internal (chore) change.

But I added a link to the PR

IhsenBouallegue (approved):

Issue #2419, opened 2021-10-04
Excluding patterns causes console error


ce-bo:
# Bug

Entering and excluding the pattern `!*.ts` causes the following console error:

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/135811855-0dd6c8e0-9b37-464f-bc51-f25f6dba5c89.png)

## Steps to Reproduce the Problem
1. Enter `!*.ts` in the file node explorer and
1. click exclude

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.79.0
- OS: Windows
- Browser: Chrome

PR #2505, opened 2021-11-15
Update blacklist item property structure to add only relevant properties #2419

Shivan-997:
# Update blacklist item property structure to add only relevant properties

closes: #2419 

## Description

Fix Issue regarding duplicate blacklist objects by updating blacklistItem interface.


visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.spec.ts
```diff
@@ -441,6 +441,15 @@ describe("nodeContextMenuController", () => {
 
 			expect(dialogService.showErrorDialog).toBeCalled()
 		})
+
+		it("should prevent duplicate blacklist object regarding issue #2419", () => {
```
ce-bo, line 445: It is not necessary to refer to the issue number. The reference is placed in the commit message anyway. Furthermore, I don't want to review the issue to be able to understand the test case :-)

BridgeAR: You are correct that the code references the issue. The test itself does not and if I want to know more about that, it's easier to identify the corresponding issue from a failing test case. :-)
Having self explanatory tests is always ideal. For regression tests, it's often still useful to know why a test case exists.

ce-bo: The test case is well described, so I am fine with keeping the (helpful) reference, of course.
It's just a personal preference.

BridgeAR (approved):

BridgeAR:
The e2e tests are failing.

Issue #2455, opened 2021-10-18
Showing "yellow" Labels not working on "Weighted Gradient"


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN clicking "show labels" for the yellow buildings

THEN the labels should appear

## Actual Behavior

THEN no label pops up

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/137688887-fbb96e5d-5d0a-465b-9075-4321c93b1f8e.png)

## Steps to Reproduce the Problem
1. Load Demo Map
1. Click "Show Labels" next to yellow color setting

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.81.0
- OS: Windows
- Browser: Chrome

BridgeAR:
This does not work with either gradient.

IhsenBouallegue:
@BridgeAR Should we move it back to the Backlog if no one is actively working on it? Because it will block the pipeline otherwise

PR #2511, opened 2021-11-16
Fix/2455/show labels for gradients properly

BridgeAR:
# Few fixes to the color gradients and code cleanup

Issue: #2455

## Description

```
fix: simplify and fix gradient color range calculation

This simplifies the gradient ranges and fixes the ranges from being
inconsistent about inclusion or exclusion of the current value.

It also fixes the colors being used for neutral to negative values
while using the "true gradient" mode.
```

```
fix: show yellow labels for gradients

This is a hotfix that does not really fix the underlying issue. Now
instead of detecting no yellow at all, too many entries are detected
while using the gradient.

Refs: https://github.com/MaibornWolff/codecharta/issues/2455
```

```
refactor: clean up code for less complexity and better types
```

## Screenshots or gifs

TBD


visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -259,24 +262,11 @@ export class CodeMapMouseEventService implements ViewCubeEventPropagationSubscri
 	}
 
 	private drawTemporaryLabelFor(codeMapBuilding: CodeMapBuilding, labels: Object3D[]) {
-		const appSettings = this.storeService.getState().appSettings
-		const showLabelNodeName = appSettings.showMetricLabelNodeName
-		const showLabelNodeMetric = appSettings.showMetricLabelNameValue
+		const { showMetricLabelNodeName, showMetricLabelNameValue } = this.storeService.getState().appSettings
 
-		let displayLabelMetricName = true
+		const displayLabelMetricName = !showMetricLabelNameValue || showMetricLabelNodeName
```
Hall-Ma, line 267: Enforcement is only necessary when showMetricLabelNameValue and showMetricLabelNodeName are disabled.

```suggestion
		const enforceLabelMetricName = !showMetricLabelNameValue && !showMetricLabelNodeName ? true : false
```

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -338,8 +327,7 @@ export class CodeMapMouseEventService implements ViewCubeEventPropagationSubscri
 			CodeMapMouseEventService.changeCursorIndicator(CursorType.Grabbing)
 		}
 		this.mouseOnLastClick = { x: event.clientX, y: event.clientY }
-		this.unhoverBuilding()
-		$(document.activeElement).blur()
+		;(document.activeElement as HTMLElement).blur()
```
Hall-Ma, line 330: Why is a semicolon at the beginning of this line?

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -413,9 +400,26 @@ export class CodeMapMouseEventService implements ViewCubeEventPropagationSubscri
...
 	}
 
+	private isGrabbingOrMoving() {
+		return this.isGrabbing || this.isMoving
+	}
+
 	private hoverBuilding(hoveredBuilding: CodeMapBuilding) {
-		if (hoveredBuilding) {
-			this.hoverBuildingAndChildren(hoveredBuilding)
+		CodeMapMouseEventService.changeCursorIndicator(CursorType.Pointer)
```
Hall-Ma, line 408: Why we don't need isGrabbingOrMoving() here?

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: This is a JS specific thing: JS does not require semicolons at the end of statements and implicitly tries to intelligently place them between statements.
Few statements require a semicolon, otherwise the output is ambiguous and that's also the case here. This would be interpreted as function call otherwise. Semicolons may be inserted anywhere between the statements. No matter if it's at the end of a statement, in-between or before the next statement (as done here).

To give you an example:

```js
const abc = [1,2,3]
(() => console.log(abc))()
// TypeError: [1,2,3] is not a function
```

```js
const abc = [1,2,3]
;(() => console.log(abc))()
// [ 1, 2, 3 ]
```

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: I guess your version might be easier to follow due to focusing on the enforced part instead of when to display the metric name. Returning `true` or `false` explicitly is not required though. That's already the case due to the statement itself being a boolean. :-)

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: There is only a single [call site](https://github.com/MaibornWolff/codecharta/pull/2511/files/a72c4acaf7286c8f71d0a983a06d2411b0b6d91c#diff-c80ef911d9938116f514f83e9a6652169f64075c015fefcee96ea555bf1a54a1L140) for this function at the moment and it has some further logic that is not required in case `isGrabbingOrMoving()` returns true. Thus, it already checks that in that function.

https://github.com/MaibornWolff/codecharta/pull/2511/files/a72c4acaf7286c8f71d0a983a06d2411b0b6d91c#diff-c80ef911d9938116f514f83e9a6652169f64075c015fefcee96ea555bf1a54a1R135-R137

Hall-Ma (approved):
LGTM, but please fix the tests and update change log.
Overall thanks a lot for cleaning up the code!

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
Hall-Ma: Thanks for your nice explanation!

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
Hall-Ma: I see. I leave it up to you whether you want to change it or not.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: I checked again and you were right, that it's not necessary in this specific case. I removed it due to that.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: I refactored the code a bit more so that the label enforcing becomes clearer.

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
BridgeAR: It is in fact added by our linter, so let's keep it as is.

visualization/app/codeCharta/ui/codeMap/codeMap.label.service.ts
```diff
@@ -41,12 +41,29 @@ export class CodeMapLabelService implements CameraChangeSubscriber {
...
+		const { appSettings, dynamicSettings, treeMap } = this.storeService.getState()
 
-		const { scaling, layoutAlgorithm } = state.appSettings
-		const { margin, heightMetric } = state.dynamicSettings
+		const { scaling, layoutAlgorithm, showMetricLabelNodeName, showMetricLabelNameValue } = appSettings
+		const { margin, heightMetric } = dynamicSettings
+
+		let labelText = ""
+
+		if (showMetricLabelNodeName || (enforceLabel && !showMetricLabelNameValue)) {
```
ce-bo, line 53: Can we please rename the variables?

```suggestion
		if (showNodeName || (enforceLabel && !showMetric)) {
```

visualization/app/codeCharta/ui/codeMap/codeMap.label.service.ts
```diff
@@ -41,12 +41,29 @@ export class CodeMapLabelService implements CameraChangeSubscriber {
...
+			labelText = `${node.name}`
+		} else if (!showMetricLabelNameValue) {
+			return
+		}
+
+		if (showMetricLabelNameValue) {
+			if (labelText !== "") {
+				labelText += "\n"
+			}
+			labelText += `${node.attributes[heightMetric]} ${heightMetric}`
```
ce-bo, line 63: Wouldn't be the color metric more appropriate?

Hall-Ma (approved):

visualization/app/codeCharta/ui/codeMap/codeMap.label.service.ts
BridgeAR: That would change the behavior from the current one. I personally would keep it as is for now.

visualization/app/codeCharta/ui/codeMap/codeMap.label.service.ts
BridgeAR: I agree that new names would be better. This PR is already quite big though. I would rather do such change later.

PR #2514, opened 2021-11-18
Refactor/2318/prepare migration of metric data services

shaman-apprentice:
This PR moves the calculation of metric data into selectors, so that we can easily access them from Angular components to come during migration.

The only "real" change, is that on `onAttributeTypesChanged` `nodeMetricData` and `edgeMetricData` aren't recalculated anymore, as they are independent of `attributeType`.


visualization/app/codeCharta/state/store/metricData/nodeMetricData/nodeMetricData.service.ts
```diff
@@ -1,77 +1,69 @@
...
 	}
 
 	getMetrics() {
-		return this.storeService.getState().metricData.nodeMetricData.map(x => x.name)
+		return this.nodeMetricData.map(x => x.name)
 	}
 
 	isMetricAvailable(metricName: string) {
-		return this.storeService.getState().metricData.nodeMetricData.some(x => x.name === metricName)
+		return this.nodeMetricData?.some(x => x.name === metricName)
```
BridgeAR, line 44: This changes the return type in case `nodeMetricData` has not yet been set from `false` to `undefined`. I suggest to just use an empty array as default value. That way the return type is not changed and there's no need for the optional chaining anymore.

visualization/app/codeCharta/ui/codeMap/codeMap.preRender.service.spec.ts
```diff
@@ -66,8 +66,10 @@ describe("codeMapPreRenderService", () => {
 		storeService.dispatch(addFile(ccFile))
 		storeService.dispatch(setSingleByName(fileStates[0].file.fileMeta.fileName))
 		storeService.dispatch(setBlacklist())
-		storeService.dispatch(calculateNewNodeMetricData(storeService.getState().files, []))
-		storeService.dispatch(calculateNewEdgeMetricData(storeService.getState().files, []))
+		// storeService.dispatch(calculateNewNodeMetricData(storeService.getState().files, []))
+		// storeService.dispatch(calculateNewEdgeMetricData(storeService.getState().files, []))
```
BridgeAR, line 70:
```suggestion
```

BridgeAR (approved):
Great work!

Just left two minor comments.

Issue #2393, opened 2021-09-27
Risk Profiles: Total Sum of percentages exceeds 100%


ce-bo:
# Bug - Risk Profiles: Total Sum of percentages exceeds 100%

## Expected Behavior

GIVEN loaded codecharta.cc.json sample map

WHEN opening Suspicious Metrics Panel

THEN the sum of percentages should be 100%

## Actual Behavior

THEN the sum of percentages exceeds 100%

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/134864482-bee12bb7-cdfa-4754-b36c-9aa5556ed41c.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.79.0
- OS: Windows
- Browser: Chrome

Hall-Ma:
We already fixed this problem in commit 3bdc263a585cdb92b63bfdca329b869aee822caa from PR: fix/2376/improve-ui-of-suspicious-metrics-feature https://github.com/MaibornWolff/codecharta/pull/2496.
But maybe we should write an additional test to check the risk calculation.

PR #2516, opened 2021-11-18
Fix/2393/risk-profile-total-sum-of-percentages

Hall-Ma:
# Sum of all percentages of the risk profile exceeds 100
Close: #2393

## Description

  - The sum of all percentages of the risk profile can exceed 100 percent under certain circumstances
  - Now package https://www.npmjs.com/package/percent-round is used and the sum of all percentages are always 100 (rounding is based on largest remainder method)


ce-bo (approved):
LGTM! 🚀

BridgeAR (commented):
Thanks for adding the tests. Please use a largest remainder method to always achieve 100%. [Percent round](https://github.com/super-ienien/percent-round) should work fine.

ce-bo (approved):

Issue #2460, opened 2021-10-20
Improve type guarding


shaman-apprentice:
# Feature request / Refactoring / Developer experience

As developer I want the TypeScript definitions to be as strict as possible, to have best possible support and to reduce unnecessary if else blocks.

## Description

When an optional property is further accessed, it has to be done guarded or TypeScript will thrown an error. E.g.:
```ts
type SomeType = { children?: number[] }
const instance: SomeType = {}
instance.children.map // this should throw an TypeScript error
```

## Acceptance criteria
- [This access without guard](https://github.com/MaibornWolff/codecharta/pull/2458#discussion_r732457181) throws an error, or `children` isn't an optional property anymore of `squarifiedNode` 
- In tsconfig.json we set `"strictNullChecks": true`
- In tsconfig.json we set `"strictPropertyInitialization": true`

## Open questions
- Needed effort for setting all the strict properties to true is unclear. In case of too much effort, we should adress this step by step.

PR #2527, opened 2021-11-28
Refactor/2318/migrate attribute side bar

### LINKED ISSUE #2539: Delta values of secondary metrics are not displayed in attribute side bar
# Bug

## Expected Behavior

GIVEN a map in delta mode

WHEN I look at secondary metrics in attribute side bar

THEN I can see delta values of secondary metrics 

## Actual Behavior
Delta values of secondary metrics are not shown

## Screenshots
image

## Specifications
- CodeCharta Version: 1.84.0

## Development notes (optional Task Breakdown)
- Note that edge metrics are stored in `edgeAttributes` of a given node

shaman-apprentice:
This PR:
- fixes #2539
  Just fixed wrong property passed down into the component, leading to values not shown. Therefore it looks now as indented and lets ignore for this fix, that it is maybe not the prettiest:
image
- migrates attribute-side-bar-component to Angular
- removes `secondaryMetrics` from store as they can be derived from selected node
- moves attribute-side-bar only related components into attributeSideBar folder

close #2539
ref #2318


visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const secondaryMetricsSelector: (state: CcState) => SecondaryMetric[] = createSelector(
+	[selectedNodeSelector, primaryMetricNamesSelector],
+	(selectedNode, primaryMetricNames) => {
+		if (!selectedNode) return []
+
+		const primaryMetricNamesList = Object.values(primaryMetricNames)
+		const secondaryMetricNames = Object.keys(selectedNode.attributes)
+			.filter(metricName => metricName !== "unary" && !primaryMetricNamesList.includes(metricName))
+			.sort()
```
BridgeAR, line 19: I very well see why you changed it. This does however change the actual behavior. The former behavior gathered all metrics from all nodes and combined them in a list. Now, only the individual node is looked at. That way, if one node contains different metrics than another node, we'll also display other metrics. Before, the unknown metrics were set to zero.

I personally am fine with this change as I can not think of a good use case for inconsistent metrics for now. We should however outline this behavior change in the changelog.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarPrimaryMetrics/primaryMetrics.selector.ts
```diff
@@ -0,0 +1,44 @@
...
+	area: Metric
+	height: Metric
+	color: Metric
+	edge: { name: string; incoming: number; outgoing: number } | undefined
+}
+
+export const primaryMetricsSelector: (state: CcState) => PrimaryMetrics | undefined = createSelector(
+	[selectedNodeSelector, primaryMetricNamesSelector],
+	(selectedNode, primaryMetricNames) => {
+		if (!selectedNode?.attributes) return
```
BridgeAR, line 17: Is this case ever triggered outside of tests? If not, it's more a kind of assertion, right?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarPrimaryMetrics/primaryMetrics.selector.ts
```diff
@@ -0,0 +1,44 @@
...
+			color: {
+				name: primaryMetricNames.nameOfColorMetric,
+				value: selectedNode.attributes[primaryMetricNames.nameOfColorMetric],
+				showAttributeTypeSelector
+			},
+			edge: {
+				name: primaryMetricNames.nameOfEdgeMetric,
+				incoming: selectedNode.edgeAttributes[primaryMetricNames.nameOfEdgeMetric]?.incoming,
+				outgoing: selectedNode.edgeAttributes[primaryMetricNames.nameOfEdgeMetric]?.outgoing,
+				showAttributeTypeSelector
```
BridgeAR, line 40: See comment below.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/nodePath.component.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+	it("should display an empty p tag, if no building is selected", async () => {
+		const { container } = await render(NodePathComponent, { componentProperties: { node: undefined } })
+		const pTag = container.querySelector("p")
+
+		expect(pTag).not.toBe(null)
+		expect(pTag.textContent).toBe("")
+	})
+
+	it("should display only node path, when a file is selected", async () => {
+		const node = { isLeaf: true, path: "some/file.ts" } as unknown as Node
```
BridgeAR, line 22: Do we have a full actual (fixture) node that we could use instead of casting the object? The same applies to the other one.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/fileCountDescription.selector.ts
```diff
@@ -0,0 +1,15 @@
+import { createSelector } from "../../../../state/angular-redux/createSelector"
+import { selectedNodeSelector } from "../../../../state/selectors/selectedNode.selector"
+
+type Node2FileCountDescription = { attributes?: { unary?: number } }
+
+export const node2fileCountDescription = (node: Node2FileCountDescription | undefined) => {
+	if (!node) return
```
BridgeAR, line 7: May this selector really be called with an undefined node?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
```diff
@@ -1,292 +1,35 @@
...
-			]
-
-			expect(attributeSideBarController["_viewModel"].secondaryMetrics).toEqual(expectedSecondaryMetrics)
-		})
-	})
+	// test display filename
+	// test primary and secondary metrics (sorted), with attribute type change
+	// test close side bar
+	// test click on leaf calls LazyLoader.openFile = jest.fn()
+	// TEST_NODE_LEAF
```
BridgeAR, line 34: I guess this is something that is still worked upon?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
shaman-apprentice: Thanks for the question. I doubled checked and came to the following conclusion:
- Previous implementation also took only attributes of current selected node into account
- But edge values where always secondary metrics, even if it is selected as primary metric as well (see screenshot below). I think new behavior is as expected and I would add an entry to change section of changelog. But I would be fine with re-implementing old behavior as well.
image


For reference previous implementation:
```ts
private updateSortedMetricKeysWithoutPrimaryMetrics(buildingSelected: boolean) {
		if (this._viewModel.node) {
			const metricValues = new Set(Object.values(this._viewModel.primaryMetricKeys.node))

			const secondaryMetricKeys = Object.keys(this._viewModel.node.attributes)
				.filter(key => !metricValues.has(key))
				.sort()

			const secondaryMetrics = []

			for (const metricKey of secondaryMetricKeys) {
				// Temporary solution to remove unary as viewable metric
				if (metricKey === "unary") {
					continue
				}

				const metricType =
					metricKey in this._viewModel.node.edgeAttributes
						? this.edgeMetricDataService.getAttributeTypeByMetric(metricKey)
						: this.nodeMetricDataService.getAttributeTypeByMetric(metricKey)
				secondaryMetrics.push({ name: metricKey, type: metricType } as SecondaryMetric)
			}
			if (!buildingSelected) {
				this.storeService.dispatch(setSecondaryMetrics(secondaryMetrics))
			}
			this._viewModel.secondaryMetrics = secondaryMetrics
		}
	}
```
with 
```
this._viewModel.primaryMetricKeys := {
                node: {
			area: string
			color: string
			height: string
		}
		edge: {
			edge: string
		}
	}
```

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Metric, shouldShowAttributeTypeSelector } from "../util/metricHelper"
+
+export type SecondaryMetric = Metric & {
+	showDeltaValue: boolean
+}
+
+export const secondaryMetricsSelector: (state: CcState) => SecondaryMetric[] = createSelector(
+	[selectedNodeSelector, primaryMetricNamesSelector],
+	(selectedNode, primaryMetricNames) => {
+		if (!selectedNode) return []
```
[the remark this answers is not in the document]
shaman-apprentice: secondary metrics are calculated by node of selected building. If there is no building selected in the map, there is no `selectedNode`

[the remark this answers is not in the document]
BridgeAR: I understood that. I would just expect the secondary metrics to only be required if a building is selected?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
BridgeAR: Thanks for cross checking! The pictures looks rather odd to me though (both, the old and the new one): the edge metric has no indication if it's a sum or the median. The values also deviate in the old version: as primary value different values are visible than the same entry as secondary value?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
```diff
@@ -0,0 +1,28 @@
...
+		const primaryMetricNamesList = Object.values(primaryMetricNames)
+		const secondaryMetricNames = Object.keys(selectedNode.attributes)
+			.filter(metricName => metricName !== "unary" && !primaryMetricNamesList.includes(metricName))
+			.sort()
+
+		return secondaryMetricNames.map(secondaryMetricName => ({
+			name: secondaryMetricName,
+			value: selectedNode.attributes[secondaryMetricName],
+			showAttributeTypeSelector: shouldShowAttributeTypeSelector(selectedNode),
+			showDeltaValue: Boolean(selectedNode.deltas)
```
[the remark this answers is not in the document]
shaman-apprentice: > Instead, it should only be done once upfront. It not something that the individual metric may turn of or on. 

done 👍 

> I would expect the property not to be exported in here at all

I am not sure about this one. We could easily create two selectors which calculate only this value. But I am not sure, if it is a single concern on its own, or if the single concern is better encapsulated as complete secondary metrics. Maybe we can have a quick call later.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarPrimaryMetrics/primaryMetrics.selector.ts
shaman-apprentice: According to [defined type here](https://github.com/MaibornWolff/codecharta/blob/main/visualization/app/codeCharta/codeCharta.model.ts#L331) `attributes is always present`. I was afraid due to #2460. But I checked previous implemantation, which doesn't include this check. So good catch :) Nullish check for `atttributes` will be removed with next push

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/nodePath.component.spec.ts
shaman-apprentice: I see your point but I am a little torn here. I also don't like to let loose complete node type if I am only interested in two properties. I made the type of the component more specific, which is now `@Input() node?: Pick<Node, "path" | "isLeaf">`. This removes the need for the cast and I can still see inline immediately the values of the test.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/fileCountDescription.selector.ts
shaman-apprentice: yes. This node comes from selected building, which can be undefined, if no building is selected in the map.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
shaman-apprentice: yes. The meaning isn't clear to me as well. The screenshot is from the default map with local development, which is not a real map, as far as I know. I also checked a real map and it doesn't make sense to me yet. Maybe we need an investigation task for it? I would like to restrict migration PRs to migration.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/fileCountDescription.selector.ts
BridgeAR: I understood that. I just wonder if the functionality itself is ever called in case no building is selected? I would expect this to trigger on building selection?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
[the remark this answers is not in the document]
shaman-apprentice: yes it is. But I strongly think a selector should never throw if given a valid store state. As selected building can be undefined in store, I think we should handle it here as well, regardless if the selector will only be called within the sidebar, which should only be rendered if there is a node.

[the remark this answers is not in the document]
BridgeAR: I understand your point. Yet, what would the receiving side do in case it receives an empty array? It might be able to handle it gracefully but maybe it does not. As such, let's go for a middle ground: throw an error in case it's called without selected node. That way we are able to let our selves be reminded how to use the functionality correct and the receiving side is never going to receive things that it's not able to handle. WDYT?

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/nodePath.component.spec.ts
BridgeAR: SGTM.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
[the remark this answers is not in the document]
shaman-apprentice: done

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
shaman-apprentice: Resolve conversation with: 
- I will adjust the hand crafted cc.json for local development some time
- created #2528

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
shaman-apprentice: done

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarPrimaryMetrics/primaryMetrics.selector.ts
shaman-apprentice: done

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarSecondaryMetrics/secondaryMetrics.selector.ts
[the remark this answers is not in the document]
shaman-apprentice: I would be OKish with adding jsdocs for throwing, or throwing an explicit exception. Although I still think selectors should be able to handle every possible state gracefully.

Besides an empty list is a valid secondary metric. A map could consist of only 3 metrics which are all primary metrics. Therefore every component using secondary metrics have to deal with that case.

Anyway I just tried out what happens when not doing this check (this nullish check of node was also done in none migrated implementation). Currently selecting another building results in deselecting first (which crashes the side bar without the check) and then selecting the new building. So lets be save with the check for now.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/fileCountDescription.selector.ts
shaman-apprentice: same as [above](https://github.com/MaibornWolff/codecharta/pull/2527#discussion_r757949475)

BridgeAR (changes requested):
There is a regression when it comes to expanding the side bar. So far it opens and closes smoothly. With this change it is a hard abrupt change.

shaman-apprentice:
> There is a regression when it comes to expanding the side bar. So far it opens and closes smoothly. With this change it is a hard abrupt change.

Thx for the catch (fixed with a116996)! I guess I must be more careful about those "fun" css things 👼

BridgeAR (approved):
LGTM. The transition seems somewhat faster than before? I guess the code is executing much faster and therefore the transition seems faster as well? It might be nice to add ~0.1 seconds to it or something like that.

shaman-apprentice:
> LGTM. The transition seems somewhat faster than before? I guess the code is executing much faster and therefore the transition seems faster as well? It might be nice to add ~0.1 seconds to it or something like that.

You are right. But I am not good with such UI things, as I only care about functionality and efficiency, when I use a software myself. 0.1s felt to fast for me. I added 0.05s and will merge with it. Feel very free to tweak it as you like :)

PR #2536, opened 2021-11-29
docs: remove duplicated chore section

shaman-apprentice:
as each section should occur only once per release for an easy overview.

ref: #2318


ce-bo (approved):
Thanks!

Issue #2480, opened 2021-11-08
Loading formerly loaded files


BridgeAR:
# Bug

## Expected Behavior

GIVEN a broken cc.json.

WHEN loading that file, an error is reported. Fix the cc.json and load again.

THEN the file opens properly.

## Actual Behavior

The file is reported as already loaded.

Such an error is not useful under any circumstances. Even already loaded maps should load fine. The map view should just be replaced. Verifying what map is loaded should only check the hash and not the filename. If a map is loaded with a different hash but the same name, just add a number to the end of the name and visualize that map as well.

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/140728963-79a14413-6ddd-4e6d-88ae-e42919b62283.png)

## Steps to Reproduce the Problem

1. Load a file
1. Try to load a file with the same name (identical and not identical!)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.82.0
- OS: all
- Browser: all

PR #2548, opened 2021-12-08
Fix/2480/loading formerly loaded files

Hall-Ma:
# Loading formerly loaded files

Close: #2480 

## Description

Problem:
- User could not load identical files (equal file name and MD5 hash) into CodeCharta

Solution:
- Now user can upload identical files (equal file name and MD5 hash) without any error messages into CodeCharta, but only one file is showed
- Now user can upload files with different files names but same MD5 hashes, replacing the previous file name with the new one
- Now user can upload files with equal file names but different MD5 hashes, adding a number to the file names to better distinguish the files

### DIFF
--- a/visualization/app/codeCharta/assets/sample1.cc.json
+++ b/visualization/app/codeCharta/assets/sample1.cc.json
@@ -1,7 +1,7 @@
 {
 	"projectName": "Sample Project with Edges",
 	"apiVersion": "1.2",
-	"fileChecksum": "invalid-md5-sample",
+	"fileChecksum": "valid-md5-sample1",
 	"nodes": [
 		{
 			"name": "root",
--- a/visualization/app/codeCharta/codeCharta.service.spec.ts
+++ b/visualization/app/codeCharta/codeCharta.service.spec.ts
@@ -12,6 +12,7 @@ import { CCValidationResult, ERROR_MESSAGES } from "./util/fileValidator"
 import packageJson from "../../package.json"
 import { clone } from "./util/clone"
 import { nodeMetricDataSelector } from "./state/selectors/accumulatedData/metricData/nodeMetricData.selector"
+import { klona } from "klona"
 
 const mockedNodeMetricDataSelector = nodeMetricDataSelector as unknown as jest.Mock
 jest.mock("./state/selectors/accumulatedData/metricData/nodeMetricData.selector", () => ({
@@ -32,6 +33,7 @@ describe("codeChartaService", () => {
 		withMockedDialogService()
 
 		validFileContent = clone(TEST_FILE_CONTENT)
+
 		metricData = clone([
 			{ name: "mcc", maxValue: 1, minValue: 1 },
 			{ name: "rloc", maxValue: 2, minValue: 1 }
@@ -57,6 +59,8 @@ describe("codeChartaService", () => {
 	}
 
 	describe("loadFiles", () => {
+		mockedNodeMetricDataSelector.mockImplementation(() => metricData)
+
 		const expected: CCFile = {
 			fileMeta: {
 				apiVersion: packageJson.codecharta.apiVersion,
@@ -121,10 +125,10 @@ describe("codeChartaService", () => {
 			}
 		}
 
-		it("should load a file without edges", () => {
+		it("should load a file without edges", async () => {
 			validFileContent.edges = undefined
 
-			codeChartaService.loadFiles([
+			await codeChartaService.loadFiles([
 				{
 					fileName,
 					content: validFileContent,
@@ -136,8 +140,8 @@ describe("codeChartaService", () => {
 			expect(isSingleState(storeService.getState().files)).toBeTruthy()
 		})
 
-		it("should load a valid file and update root data", () => {
-			codeChartaService.loadFiles([
+		it("should load a valid file and update root data", async () => {
+			await codeChartaService.loadFiles([
 				{
 					fileName,
 					content: validFileContent,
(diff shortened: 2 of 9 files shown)


Christian-Eberhard (approved):

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -67,7 +66,55 @@ export class CodeChartaService {
...
+	private findFileName(currentFileName: string, isDuplicate: boolean, storedFileNames: Map<string, string>, currentFileChecksum: string) {
+		if (storedFileNames.get(currentFileName) === currentFileChecksum) {
+			return currentFileName
+		}
+
+		let nameFound = false
+		let fileNameOccurrence = 1
+		let newFileName = currentFileName
+		while (!nameFound) {
+			newFileName = currentFileName.replace(currentFileName, `${currentFileName.split(".")[0]}_${fileNameOccurrence}.cc.json`)
```
ce-bo, line 101: Do you need this line? In Line 103 the newFileName is already set.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -67,7 +66,55 @@ export class CodeChartaService {
...
+			const endOfNameIndex = currentFileName.indexOf(".")
+			newFileName =
+				endOfNameIndex >= 0
+					? [currentFileName.slice(0, endOfNameIndex), "_", fileNameOccurrence, currentFileName.slice(endOfNameIndex)].join("")
+					: `${currentFileName}_${fileNameOccurrence}`
+			// resolve if uploaded file has identical checksum and different already occurring name
+			if (isDuplicate && storedFileNames.has(newFileName) && storedFileNames.get(newFileName) === currentFileChecksum) {
```
ce-bo, line 108:
```suggestion
			if (isDuplicate && storedFileNames.get(newFileName) === currentFileChecksum) {
```

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -67,7 +66,55 @@ export class CodeChartaService {
...
+				nameFound = true
+			}
+			if (!storedFileNames.has(newFileName)) {
+				nameFound = true
```
ce-bo, line 112: I think you can `return newFileName` immediately, in those cases (nameFounde = true). Then, the wrapping `else` is redundant and can be removed.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -67,7 +66,55 @@ export class CodeChartaService {
...
+			this.recentFiles[0] = currentFileName
+			this.recentFiles.push(currentFileName)
+			return
+		}
+
 		this.fileStates.push({ file: ccFile, selectedAs: FileSelectionState.None })
+		this.recentFiles.push(currentFileName)
+	}
+
+	private findFileName(currentFileName: string, isDuplicate: boolean, storedFileNames: Map<string, string>, currentFileChecksum: string) {
```
ce-bo, line 92: I would rename this function. `find` indicates that not always a file name can be found.

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -154,39 +158,85 @@ describe("codeChartaService", () => {
...
+			validContent3ndFile.fileChecksum = "validHash_2"
+
+			await codeChartaService.loadFiles([{ fileName: "FirstFile", content: validFileContent, fileSize: 42 }])
+
+			await codeChartaService.loadFiles([
+				{ fileName: "FirstFile", content: validContent2ndFile, fileSize: 42 },
+				{ fileName: "FirstFile", content: validContent3ndFile, fileSize: 42 }
+			])
+
+			expect(getCCFiles(storeService.getState().files).length).toEqual(3)
```
ce-bo, line 192: Minor note: It would be a little shorter to store the files in a variable instead of getting all CC-Files again and again :-)

visualization/app/codeCharta/codeCharta.service.ts
Hall-Ma: I didn't change that. Otherwise you would have `while(true)`. But I agree that we can delete `else`

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -154,39 +158,85 @@ describe("codeChartaService", () => {
...
 		})
 
-		it("should not replace the last files when loading new files", () => {
-			codeChartaService.loadFiles([
-				{
-					fileName: "FirstFile",
-					content: validFileContent,
-					fileSize: 42
-				}
+		it("should replace files with equal file name and checksum when loading new files", async () => {
```
ce-bo, line 161: According the bug description of the issue, we should add a regression test to handle that case (uploading broken file, fix it manually, upload again).

visualization/app/codeCharta/assets/sample1.cc.json
```diff
@@ -1,7 +1,7 @@
 {
 	"projectName": "Sample Project with Edges",
 	"apiVersion": "1.2",
-	"fileChecksum": "invalid-md5-sample",
+	"fileChecksum": "valid-md5-sample1",
```
ce-bo, line 4: Nit picking: Intentionally, "invalid-md5-sample" should express that the sample value is not a valid md5. So, "valid-md5-sample" is a bit confusing here. I agree that the former is not that obvious. Therefore, we could just omit the prefix "valid".

ce-bo (commented):
LGTM! However, we could remove the recentFiles handling from the store because it is nowhere else needed and it just makes no sense to dispatch this. The recentFiles array is used to get access to the first uploaded file, so that it can be marked as selected. This can be done locally in codeCharta.service.ts.

ce-bo (approved):

PR #2549, opened 2021-12-10
chore: activate strictFunctionTypes typescript check

BridgeAR:
This activates the strictFunctionTypes check to verify type parameter
positions contravariantly.
It also fixes issues found due to that check.


visualization/app/codeCharta/state/angular-redux/createSelector.ts
```diff
@@ -20,18 +20,22 @@ export function createSelector<S1, S2, S3, S4, S5, P, State = CcState>(
...
+export function createSelector<S1, S2, S3, S4, S5, P, State = CcState>(
+	selectors: [Selector<S1, State>, Selector<S2, State>?, Selector<S3, State>?, Selector<S4, State>?, Selector<S5, State>?],
+	projector: (s1: S1, s2?: S2, s3?: S3, s4?: S4, s5?: S5) => P
 ): (state: State) => P {
 	let lastValues = []
 	let memorizedResult: P
 
 	return (state: State) => {
-		const values = selectors.map(s => s(state))
+		const values: [S1, S2?, S3?, S4?, S5?] = [selectors[0](state)]
```
shaman-apprentice, line 31: Not worth the discussion but out of curiosity: I like the shorter `map` more, as it can be easier read and needs less local variables. Why do you prefer the `for`-loop?

shaman-apprentice (approved):
LGTM

visualization/app/codeCharta/state/angular-redux/createSelector.ts
BridgeAR: In TS it's a tuple, not an array but the map would return an array. TS is not yet smart enough to understand it otherwise.

shaman-apprentice: indeed. How annoying.

ce-bo (approved):

PR #2550, opened 2021-12-10
chore: skip sonar-java-plugin updates with renovate

BridgeAR:
As discussed. I am not certain if it works as expected but this is likely a good starting point.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,6 +4,7 @@
 		{
 			"matchPaths": ["build.gradle"],
 			"matchUpdateTypes": ["minor", "patch", "pin", "digest"],
+			"excludePackageNames": ["org.sonarsource.java:sonar-java-plugin"],
 			"groupName": "gradle dependencies not Major update",
 			"enabled": true
 		},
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -29,6 +30,12 @@
 			"matchUpdateTypes": ["minor", "patch", "pin", "digest"],
 			"groupName": "other updates",
 			"enabled": true
+		},
+		{
+			"matchPackageNames": ["@types/three", "org.sonarsource.java:sonar-java-plugin", "three"],
+			"matchUpdateTypes": ["minor", "patch", "pin", "digest"],
```
ce-bo, line 36: Shouldn't major be also in this list?
```suggestion
			"matchUpdateTypes": ["major", "minor", "patch", "pin", "digest"],
```

BridgeAR: I did not check but I thought major would be handled separately anyway or where is that configured for our other dependencies?

ce-bo: Because there is no explicit configuration for major PRs Renovate opens a separete one for every major update.
So it would make sense to me to cover those updates in this new (disabled) group also.

ce-bo: By the way, `matchUpdateTypes` (with major included) would match all update types. Thus, I assume we can omit it.

Shivan-997: Renovate will consider by default all types
```suggestion
```

Hall-Ma (approved):

Issue #1661, opened 2021-01-12
Prevent popups from rendering out of view


shaman-apprentice:
# Feature request

## Description

As a user I want popups to be always completely visible in the view, so that I don't have to shift things around before opening a popup, as shifting things around just for opening a popup is annoying.

Example issues:
- node context menu and its color-picker are each partially out of view:<br/>
  ![image](https://user-images.githubusercontent.com/3596742/104308228-a1d5b480-54d0-11eb-9e22-20b6b3c7bd47.png)
- color-picker and legend are overlapped by ribbon-bar<br/>
  ![image](https://user-images.githubusercontent.com/3596742/104308432-f416d580-54d0-11eb-9b13-373f27ea73f4.png)

## Acceptance criteria
- The two examples from above are not reproducable

## Development notes (optional Task Breakdown)
- We could custom implement a solution via [getBoundingClientRect](https://developer.mozilla.org/de/docs/Web/API/Element/getBoundingClientRect) (probably a bad idea)
- We could utilize components like [material.angularjs.org](https://material.angularjs.org/latest/demo/dialog)
- We could tackle this issue once after we have moved to Angular. In the long run this is in total probably the least amount of work

BridgeAR:
I am currently cleaning up the backlog. As this does not seem like it's something we are going to address anytime soon, I am going to close it. As soon as we finished the migration, we might pick this up again.

shaman-apprentice:
I actually planned to tackle it, when migrating context menu and color picker with the help of Angular Materials CDK. Mind if I reopen it?

BridgeAR:
The need for it is currently pretty low. Please do not invest too much time into it. If it's difficult to resolve, let it be for now.

shaman-apprentice:
FYI: The node context menu has still the problem ;) But I am OK with leaving the issue closed. Then I will ninja fix it when migrating the component.

PR #2552, opened 2021-12-10
refactor: migrate legendMarkedPackages

shaman-apprentice:
- migrate legendMarkedPackages
- introduce an Angular color-picker for it

Previous a right-click on the map closed the color picker. Now it opens default browser behavior. Should we keep the old behavior?
image

ref #2318
ref #1661


shaman-apprentice:
superceded by #2560

Issue #2489, opened 2021-11-09
Improve custom configs


BridgeAR:
# Feature request

## Description

As a user, I want custom configs to be easily understandable and easy to use so that I do not have to invest a lot of time using them.

## Acceptance criteria
Please address the tasks (when in makes sense) in separate PRs to deliver value to the customer in small interations.

### Minimum set before releasing the feature as stable:
- [x] Remove notification when removing entries (just remove the entry by collapsing the list)
- [x] "AI" entries should not be kept in the local storage. They are recalculated as soon as the map is loaded anyway. As discussed [here](https://github.com/MaibornWolff/codecharta/issues/2435), this will resolve #2435 and #2436.
- [x] Do not close Custom Config menu after deleting a config. 
- [x] Do not show non-matching entries by default! Only list matching entries for the loaded maps. All others should be in a separate category.  
- [x] Rename everything related to custom configs/map to "Custom view"
- [x ] Separate applicable and not-applicable entries.
- [x] Add a "Show not-applicable entries" button below the list of applicable views and hide those by default.
- [x] Add a subtitle in the Custom Views Modal to explain in one sentence for what the User can use it (e.g. Store your current Settings/Configuration, Download and Import them.). Add a link to the docs for more information.
- [x] Partially applicable entries (only a single map is loaded and there's a custom view for that map in multiple mode in combination with another map) count as applicable. #3090
- [x] Partially applicable entries should be clickable but greyed out. While hovering over those, show what is missing to apply that custom view (e.g. map X is not loaded;map X is loaded in standard mode, but config is created for delta mode). On click, the config should still be applied, although it is greyed out. #3090
- [x] Add icons/names/descriptions of `metrics` (area: rloc, height: foo, color: bar) and `mode` (single,multiple,delta) below the custom view name. (#3119)
- [x] "Somehow" show the used colors (at least if they deviate from the default [green, yellow, red]) in the view  (#3119)
- [x] Rename "multiple" to "standard" mode in list view (e.g. "Custom View(s) in mutliple mode for ...."). (#3119)

### Tasks to release the feature as a stable one:
------
- [x] Mark custom configs as stable (should be shown also when experimental mode is disabled)
- [x] Add a How To article in our documentation. Describe how to use it, how to Download/Import/Share custom configs. Describe how and why it is replacing the download button.
- [x] Show a deprecation warning in the changelog that the download button is going to be removed soon. Link to the How To article.
[...]

PR #2553, opened 2021-12-10
Feature/2489/remove-notification-when-deleting-custom-config

Hall-Ma:
# Removing of notification dialog when custom config is deleted

Issue: #2489 

## Description

  - When a user deletes a custom config then a notification dialog (that something is deleted) appears that disrupts the user experience.
  - Removing this notification dialog solves this problem.


ce-bo (approved):
LGTM. Please update the Changelog :-)

PR #2558, opened 2021-12-13
Refactor/2318/migrate legend panel

shaman-apprentice:
continuation of #2552

ref #2318
ref #1661

PR #2559, opened 2021-12-13
Refactor/2818/replace store color picker

shaman-apprentice:
continuation of #2558

ref #2318
ref #1661


shaman-apprentice:
superceded by #2560

Issue #475, opened 2019-04-09
Visualization can export 3D printer files


Richargh:
In addition to the regular cc.json export, we could also provide a 3d printer export. Then people could print their software and watch it grow, year after year. 

![Printer-Export.drawio.png](https://images.zenhubusercontent.com/5cd16a929795a865159e70ff/c0712b35-65e1-4ab3-8f37-9b1d2d60f4b5)

## Open Questions:
What should we do with the floor labels?
How should we prevent floating buildings/floors?
What should we do for tiny buildings?

BridgeAR:
While it's a nice idea, I am not convinced we should implement something like that. Do you see an actual need for anyone to print code maps as a 3D model? It's probably not something many would need but mainly something to play around with?

Richargh:
We had this feature on a branch at one point. @mrpatpat knows more. As far as I know it didn't take long to implement. A day maybe?

It's a vanity/gimmick for sure. I love the idea to be able to print the software. You could print it year after year and watch it change. After an audit you could give this as a final present (probably with modified colors). I did consider putting it in CC if the implementation cost was acceptable but then edges took priority and I forgot... :|

mrpatpat:
As far as I remember we used a already implemented solution. I think it was based on the [STL-Exporter included in Three.js ](https://github.com/mrdoob/three.js/blob/master/examples/js/exporters/STLExporter.js). A usable example for a test print is located [here](https://threejs.org/examples/misc_exporter_stl.html)
It takes a scene as input and outputs a STL file which can be printed. 
I do not think integrating it is a complex task but there may be some needed work in order to make the codecharta scene itself ready for printing (e.g excluding hard to print structures like arrows or labels). 
I hope this helps. And yes it is a gimmick but it always sparked joy when we talked with people about it.

BridgeAR:
The initial description for the feature request is to add it to the export feature. Let's discuss where to put the download button as the current one is actually planned for removal.

IhsenBouallegue:
@BridgeAR I like doing it that way.
Well, it also means that we really are pushing this feature because it will be immediately visible. But I find this a good thing. As people will love it and share their 3D-printed maps with others and even the whole company. Good marketing for us.
We can have instead of the download button a 3D icon, and it converts the map into an STL file. The trickiest part is exporting it in a way that can be 3D printable (eliminating small buildings while still keeping the general look of the map)




> 
> 
[...]

PR #2561, opened 2021-12-15
Feature/475/export 3d printer files

IhsenBouallegue:
# Introduction of 3D printing feature

Close: #475 

## Description

Now we can export .stl files from visualization. These files are then used with a 3D slicer software to generate the appropriate G-code for the 3D printer.

With the current implementation, the opened map will be downloaded as an .stl file after clicking the export 3D button.

## How-to

1. Click the 3D export button
grafik

2. Save the STL file
grafik

3. Import file into your chosen slicer software and configure (Cura 3d shown below)
grafik

4. Print! 🖨️ 🥳 
grafik


Christian-Eberhard (approved):

visualization/app/codeCharta/ui/codeMap/threeViewer/threeSceneService.ts
```diff
@@ -35,6 +35,8 @@ export class ThreeSceneService implements CodeMapPreRenderServiceSubscriber, Map
 	mapGeometry: Group
 	private readonly lights: Group
 	private mapMesh: CodeMapMesh
+	/** Temporary workaround until mapMesh can be derived from store */
```
ce-bo, line 38: The TODO Prefix will highlight this comment in IDEs :-)
```suggestion
	/** TODO: Fix temporary workaround as soon as mapMesh can be derived from store */
```

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Mesh } from "three"
+
+@Component({
+	selector: "cc-export-threed-map-button",
+	template: require("./export3DMapButton.component.html")
+})
+export class Export3DMapButtonComponent {
+	private fileName: string
+	private files: FileState[]
+	exporter = new STLExporter()
```
ce-bo, line 21: Can this property be private?

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
```diff
@@ -0,0 +1,38 @@
...
+	constructor(@Inject(Store) store: Store) {
+		store.select(accumulatedDataSelector).subscribe(accumulatedData => {
+			this.fileName = accumulatedData.unifiedFileMeta?.fileName
+		})
+		store.select(filesSelector).subscribe(files => {
+			this.files = files
+		})
+	}
+
+	downloadStlFile() {
```
ce-bo, line 32: I would add some tests for this logic. Some of my ideas: Propably for edge cases like Empty ThreeMesh, One Selected File, Multiple Selected Files, Delta File. And I would test that the right stl file will be downloaded if the mode changes or another map was uploaded.

ce-bo (commented):
The code is LGTM! :-) 
Please add some tests and adjust the changelog. Thanks!

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
IhsenBouallegue: I would love some help with that. I tried but didn't succeed, as I was constantly stuck.

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.scss
```diff
@@ -0,0 +1,27 @@
...
+	.button-icon {
+		font-size: 16px;
+		line-height: 1.3;
+	}
+	span {
+		padding: 0;
+		margin: 0;
+	}
+	.mat-mini-fab .mat-button-wrapper {
+		height: 25px;
```
shaman-apprentice, line 22: width and height are redundant with above `.mat-mini-fab` and can be removed here, aren't they?

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
```diff
@@ -0,0 +1,38 @@
...
+	selector: "cc-export-threed-map-button",
+	template: require("./export3DMapButton.component.html")
+})
+export class Export3DMapButtonComponent {
+	private fileName: string
+	private files: FileState[]
+	private exporter = new STLExporter()
+
+	constructor(@Inject(Store) store: Store) {
+		store.select(accumulatedDataSelector).subscribe(accumulatedData => {
```
shaman-apprentice, line 24: we must unsubscribe, if we subscribe manually

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
shaman-apprentice: The download doesn't depent on the CodeCharta state, does it? As far as I understand it, it takes the Three.js model and generates a binary for the 3d printer out of it. Therefore we don't need to test different CodeCharta states like multiple or delta mode...

I would write a test that the FileDonloader gets called. The FileDownlaoder then could be tested through pure unit tests, I hope.

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.scss
IhsenBouallegue: Resolved in the next push

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
ce-bo: Ah, I see. You are right!
@shaman-apprentice Does unifiedFileMeta mean that it is a combination of all selected files?

shaman-apprentice: yes. But I am not to sure about the naming and calculation - it is extracted by one of the services into a selector, as it is purely dependend from store state and can then easily be used by Angular components (and has better separations of concers, as the services do too much anyway)

shaman-apprentice: We added a test which verifies the template and controller binding. 

The download logic is still untested. It uses mostly external libs which are awkward to test with current design. We discussed a better approach, which is possible after Angular migration. Current implemented approach is still better than our "normal" Controller tests. According to the boy scouts rule I am OKish with it for now  (although the rule only kind of applies, as it is new code).

shaman-apprentice (approved):
Looks OK to me

visualization/app/codeCharta/ui/export3DMapButton/export3DMapButton.component.ts
ce-bo: Personally, I would try to spy on the STLExporter:parse() and the FileDownloader:downloadData() function. Then, I would just test that the file name is set correctly.

ce-bo: I tested locally to download the STL file for 5 selected maps in multiple mode. The resulting file name is very long: `file_aggregation_of_sample1_and_sample2_and_codecharta_2021-12-16_14-52_and_junit4.source.modified_and_codecharta_edges_and_codecharta_analysis_2021-12-18_12-45.stl`. 

@BridgeAR Do you have any requirement for that?

ce-bo (approved):
LGTM! Thanks for completing the Tests. Would you please Update the Name of the new Test? Something like „should Download stl file Witz proper Name“ Would be More appropriate.

PR #2563, opened 2021-12-15
Feature/2489/remove suspicous metrics from custom config

Shivan-997:
# remove suspicous metrics from custom config

issue: #2489


ce-bo:
Continued here: #2564

Issue #772, opened 2019-11-08
ccsh argument inconsistency


NearW:
# Feature request

## Description

The argument description is inconsistent. Sometimes it says --projectName and sometimes --project-name.
We should agree on camelCase or kebab-case.

alexhunziker:
This may be off-topic but: Speaking of the projectName.... in the analysis we pretend that this attribute is super important (by default the merging of two project fails if the names are not matching for example) but in the visualization we do almost/absolutely nothing with this attribute. 

Therefore the question: What is the purpose of this field? If there is none, let's remove it. If there is one, we might want to show the project name in the visualization as well.

alschmut:
I have actually been asking this question to myself since I am part of CodeCharta and I would say the original reason to introduce it, was to avoid the merge of two different projects. But to be honest the projectName is not always set correctly and depends (if I am right) mostly on the importer. I think we can evaluate if files should be merged or not without administrating the projectName, simply by comparing some of the top nodes in the cc.json map.
- If the top nodes seems to be mostly identical, we can merge it
- If they don't we could abort and suggest to double check the files with the structure modifier

alexhunziker:
Yes, you are right, if the projectName is not set manually, it depends on the importer (i guess with the exception of the SonarImporter). With #567 we have introduced Log messages, which inform the user to what degree the merged projects were matching up. This is in my opinion enough information to the user so that he can see if the  merge yielded the desired result or not. 

If there is no other purpose of the projectName I say let's remove it completely.

alschmut:
Are there any new thoughts on this issue? I am actually really excited to have this cleaned up 😃

alexhunziker:
When cleaning up the arguments, i suggest that in the SCMLogParser we remove the --git and --svn arguments, and instead always ask for the input format. 

Why? First --git does nothing else than set --input-format to GIT_LOG and second GIT_LOG is just one of the candidates for a git log file and may not be appropriate. If both arguments are set, the --input-format is ignored. So if you do somehting along the lines of `ccsh scmlogparser foo.log --git --input-format=GIT_LOG_NUMSTAT_RAW` the parser will discard GIT_LOG_NUMSTAT_RAW and instead use GIT_LOG, which is unexpected behaviour in my opinion. Therefore i suggest removing --git and --svn altogether. That way the interface becomes more understandable.

alschmut:
@alexhunziker I agree with removing the simplified `--git` and `--svn` parameters

[...]

PR #2564, opened 2021-12-17
Feature/2489/Remove-suspicious-metrics-from-custom-config

Hall-Ma:
# Removing suspicious metrics from custom config

ref: #2489

## Description

- To reduce the load on the user's local storage, the suspicious metrics are no longer stored in the custom configs.
- Overall, this leads to better performance.


visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -201,49 +216,29 @@ export class ArtificialIntelligenceController
 	private createCustomConfigSuggestions(fileState: FileState, programmingLanguage) {
```
ce-bo, line 216: Please remove the custom Config Reference.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.html
```diff
@@ -43,8 +43,8 @@
 			<ul class="suspicious-metrics">
 				<li class="suspicious-metric" ng-repeat="item in $ctrl._viewModel.suspiciousMetricSuggestionLinks">
 					<md-button
-						class="metric-button {{ item.outlierCustomConfigId.length > 0 ? 'high-risk' : '' }}"
-						ng-click="$ctrl.applyCustomConfig(item.generalCustomConfigId)"
+						class="metric-button {{ item.isOutlier ? 'high-risk' : '' }}"
```
ce-bo, line 46: Why do we need this css class (high-risk)?

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -96,8 +86,33 @@ export class ArtificialIntelligenceController
 		}
 	}
 
-	applyCustomConfig(configId: string) {
-		CustomConfigHelper.applyCustomConfig(configId, this.storeService, this.threeCameraService, this.threeOrbitControlsService)
+	applySuspiciousMetric(metric: MetricSuggestionParameters, isOutlier: boolean) {
```
ce-bo, line 89: I would rather remove the second parameter of `applySuspiciousMetric()`. We can get this information from the `MetricSuggestionParameters`.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -96,8 +86,33 @@ export class ArtificialIntelligenceController
...
+				from: metric.from,
+				to: metric.to,
+				max: 0,
+				min: 0
+			}
+
+			if (metric.isOutlier === isOutlier) {
+				mapColors.positive = "#ffffff"
+				mapColors.neutral = "#ffffff"
+				mapColors.negative = "#A900C0"
```
NearW, line 103: This will mutate the state, since mapColors is a reference to an object inside the state.
The state should only be changed through dispatching an action. Instead, create a new object by spreading the mapColors from the state inside it, mutate the new object and dispatch it.

```
const mapColors = { ...this.storeService.getState().appSettings.mapColors }
```

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.html
Hall-Ma: No, I will delete it.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
Hall-Ma: The problem is that we need to pass an additional parameter into the method, because the user may want to look at a suspicious metric or the very high risk view (only colors are changed). A suspicious metric may have `true` in `isOutlier`, but the user only wants to see the normal suspicious view. Do you maybe have an idea for this problem?

IhsenBouallegue (approved):
LGTM 🎉

ce-bo (approved):

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
ce-bo: Ah, I just noticed that very high risk configs do not exist asindependent configs. In Line 228 the `isOutlier` flag of an existing config is set.

analysis/import/SCMLogParser/src/test/resources/expected_git.json
```diff
@@ -1,35 +1,49 @@
...
-          ]
-        }
-      ]
-    }
-  ]
-}
\ No newline at end of file
+  "checksum": "foo",
+  "data":{
+    "projectName": "",
```
ce-bo, line 4: Do you know where the projectName normally comes from?
What is the reason that it is empty now?

ce-bo: I did a little research. In https://github.com/MaibornWolff/codecharta/issues/772 the projectName was intended to be removed. But this has never happend (completely). A `DUMMY_PROJECT_NAME` has been introduced which is just an empty string: de.maibornwolff.codecharta.model.ProjectBuilder::DUMMY_PROJECT_NAME

Thus, the empty projectName here is ok.

ce-bo (approved):

PR #2588, opened 2022-01-01
refactor: move duplicated logic into cc-color-picker

shaman-apprentice:
the duplication for opening the color picker in context-menu and the other color pickers in a menu annoyed me. So I refactored it out into the common cc-color-picker component.


Hall-Ma (approved):
LGTM :)

PR #2590, opened 2022-01-04
do not close custom config menu when try deleting a config #2489

Shivan-997:
# Custom config stay open after deletion
Issue: #2489 

## Description
Do not close Custom Config menu after deleting a config.

##  Question
after deletion the items group drop-down get closed, is that acceptable behavior


visualization/app/codeCharta/ui/customConfigs/customConfigs.component.ts
```diff
@@ -76,6 +76,7 @@ export class CustomConfigsController implements FilesSelectionSubscriber {
 
 	removeCustomConfig(configId) {
 		CustomConfigHelper.deleteCustomConfig(configId)
+		this.loadCustomConfigs()
```
ce-bo, line 79: In a perform manner I would remove the deleted Config from `this._viewModel.dropDownCustomConfigItemGroups` instead of accessing the localStorage API which to avoid I/O.
But this will do the job and I think it is not performance critical. So, I am fine with it :-)

visualization/app/codeCharta/ui/customConfigs/customConfigs.component.html
```diff
@@ -67,6 +67,7 @@
 							ng-click="$ctrl.removeCustomConfig(customConfig.id)"
```
ce-bo, line 67: I think it is possible to prevent collapsing the list on remove.
You could try to append a call to `$event.stopPropagation();` within the `ng-click` attribute (untested).

Hall-Ma:
Issue is already fixed in https://github.com/MaibornWolff/codecharta/pull/2591

Issue #2523, opened 2021-11-24
"Know more" button in What's new section doesn't work


Hall-Ma:
# Bug

## Expected Behavior

When "Know more" button is clicked then news page should open a new tab.

## Actual Behavior

"Know more" button doesn't link to news page

## Screenshots

![Bildschirmfoto 2021-11-24 um 12.08.18.png](https://images.zenhubusercontent.com/615eb4f76331b83ced10ffe0/2f345f49-8468-4d79-90d7-0f1420621128)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.84.0
- OS: all
- Browser: Chrome

Hall-Ma:
My suggestion is to use 
`a href="#" onClick="window.open('https://maibornwolff.github.io/codecharta/news/', '_blank')" class="md-primary">Know more</a>` in dialog.changelog.component.ts

ce-bo:
I would be very suprised, if the `<a>` tag does not work without JavaScript in this case.
Adding `target="_blank"` works for me.

I just noticed that in the current implementation the link does not work but the URL has changed.
(Windows, Chrome)

Hall-Ma:
I wonder that the current implementation works local but not in the deployed version. Furthermore it would be more convenient when a new tab would open the news page.

PR #2602, opened 2022-01-11
Fix/2523/Relink to news page when clicking 'Know more' button in changelog dialog

Hall-Ma:
# Relink to news page when clicking 'Know more' button in changelog dialog

Closes: #2523

## Description

- when clicking 'Know more' button in changelog dialog it will open the news page in a new tab


ce-bo (approved):

Hall-Ma:
@ce-bo I really don't know why the previous implementation worked only on localhost, but I implemented a solution that seems to work (we use it in our global settings (links to Wiki, Github, ...)).  I hope that will solve this issue. because I cannot test it correctly.

Issue #2526, opened 2021-11-28
User-friendly renaming of display quality in global settings


Hall-Ma:
# Feature request

## Description

As a user, I want to have a concise labeling of the graphic settings so that I don't have to look up what "Pixel Ration With FXAA Antialiasing" means, for example.

## Acceptance criteria
- Rename current display quality descriptions in a more generic way like "standard", "low", "high" and "best"

PR #2603, opened 2022-01-12
Fix/2562/Selected-color-building-in-delta-mode

Hall-Ma:
# Color change of the selected building in delta mode 

Closes: #2526

## Description

- When a building is selected (left mouse click) in delta mode, the whole building is now colored (with selected color)

### DIFF
--- a/visualization/app/codeCharta/ui/codeMap/rendering/codeMapMesh.ts
+++ b/visualization/app/codeCharta/ui/codeMap/rendering/codeMapMesh.ts
@@ -47,6 +47,7 @@ export class CodeMapMesh {
 
 	selectBuilding(building: CodeMapBuilding, color: string) {
 		building.setColor(color)
+		building.setDeltaColor(color)
 		this.setVertexColor(building.id, building.getColorVector(), building.getDefaultDeltaColorVector())
 		this.updateVertices()
 	}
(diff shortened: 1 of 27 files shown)


visualization/app/codeCharta/ui/codeMap/rendering/codeMapMesh.ts
```diff
@@ -47,6 +47,7 @@ export class CodeMapMesh {
 
 	selectBuilding(building: CodeMapBuilding, color: string) {
 		building.setColor(color)
+		building.setDeltaColor(color)
```
ce-bo, line 50: I think this could introduce a hidden side effect because the delta color is never resetted (see `clearSelection()`).

Hall-Ma: ClearSelection() calls resetColors() ([See here](https://github.com/MaibornWolff/codecharta/blob/b0b663699f156bf7f538b58958351f9097a79439/visualization/app/codeCharta/ui/codeMap/rendering/codeMapBuilding.ts#L71)) and there color and delta color are resett.

ce-bo (approved):
Code LGTM (but not tested locally on my machine) :-)

IhsenBouallegue:
Tested locally, works. But there seems to be a weird problem with gradle. I relaunched the GitHub actions. Hopefully that does the trick

Issue #2094, opened 2021-07-06
All files are rejected if one throws an error


ngormsen:
# Bug

## Expected Behavior

GIVEN multiple files where some are valid and at least one is not

WHEN I load all of them at the same time

THEN I want every file that does not have an error to be loaded.

## Actual Behavior

All files are rejected if a single file throws an error.
 
## Steps to Reproduce the Problem
1. Load multiple files with only some of them being valid

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.75.0
- OS: macOS
- Browser: Chrome

PR #2604, opened 2022-01-14
Fix/2094/all files are rejected on import

IhsenBouallegue:
# Improve file validation handling

Close: #2094 

## Description

  - All files used to be rejected on import if one is erroneous.
  - This PR enables all correct files (also ones containing just warnings) to be loaded, regardless of the erroneous files.
  - This PR implements better and more descriptive error messages on file import

## Screenshots or gifs

grafik


visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -35,41 +35,57 @@ export const ERROR_MESSAGES = {
 export function validate(nameDataPair: NameDataPair) {
```
ce-bo, line 35: I do not think `validate()` needs to be exported.

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -35,41 +35,57 @@ export const ERROR_MESSAGES = {
...
 	const file = nameDataPair?.content
 	const result: CCValidationResult = { error: [], warning: [] }
+
+	result.error = checkErrors(file)
+	result.warning = checkWarnings(file)
+
+	if (result.error.length > 0) {
+		throw result.error
+	}
+}
```
ce-bo, line 45: I propably would call `checkErrors` and `checkWarnings` in loadFiles directly. The `validate` function is not needed anymore. And instead of throwing errors and warnings from one function to another, I would just process the return values of the two new check functions.

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -35,41 +35,57 @@ export const ERROR_MESSAGES = {
...
+
+function checkWarnings(file: ExportCCFile) {
+	const warnings = []
+	switch (true) {
+		case fileHasHigherMinorVersion(file):
+			warnings.push(`${ERROR_MESSAGES.minorApiVersionOutdated} Found: ${file.apiVersion}`)
+			break
+	}
+	return warnings
```
ce-bo, line 54:
```suggestion
	const warnings = []
	if (fileHasHigherMinorVersion(file)) {
			warnings.push(`${ERROR_MESSAGES.minorApiVersionOutdated} Found: ${file.apiVersion}`)
	}
	return warnings
```

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -35,41 +35,57 @@ export const ERROR_MESSAGES = {
...
-			result.error.push(ERROR_MESSAGES.majorApiVersionIsOutdated)
-			break
-		case fileHasHigherMinorVersion(file):
-			result.warning.push(`${ERROR_MESSAGES.minorApiVersionOutdated} Found: ${file.apiVersion}`)
+			errors.push(ERROR_MESSAGES.majorApiVersionIsOutdated)
 			break
 	}
 
-	if (result.error.length === 0) {
+	if (errors.length === 0) {
```
ce-bo, line 71: The JSON Schema validations is a separate concern and should therefore be extracted into a separate function.

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Awesome that would simplify the code even further!

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Will implement in the next commit.

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Implemented in the next commit.

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Implemented in the next commit.

visualization/app/codeCharta/util/fileValidator.ts
```diff
@@ -32,42 +32,50 @@ export const ERROR_MESSAGES = {
...
 			break
 	}
+	if (errors.length === 0) errors.push(...checkJsonSchema(file))
+	return errors
+}
 
-	if (result.error.length === 0) {
+function checkJsonSchema(file: ExportCCFile) {
+	const errors: string[] = []
+	if (errors.length === 0) {
```
ce-bo, line 65: This is always true, isn't it?

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -25,28 +25,26 @@ export class CodeChartaService {
...
+
 		for (const nameDataPair of nameDataPairs) {
-			try {
-				validate(nameDataPair)
+			errors.push(...checkErrors(nameDataPair.content))
+			if (errors.length > 0) {
+				this.fileStates.filter(file => getCCFile(nameDataPair) !== file.file)
+				this.recentFiles.filter(fileName => fileName !== nameDataPair.fileName)
+			}
+			if (errors.length === 0) {
```
ce-bo, line 37: The behavior before was different. Errors and warnings were shown at the same time. Now, Warnings are shown only if no errors are present.

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Missed that! I will clean it up, thanks 😃

visualization/app/codeCharta/codeCharta.service.ts
IhsenBouallegue: That is true. But it didn't make sense to me. Let's say a file has an error, which means something is seriously wrong with it. Why should we bother checking for warnings at this point? Plus, in the future if we decide to add warnings that assume that the file structure is correct, the old implementation is not flexible enough.
When we show the user the validation results, I think if a file has an error, the user won't care about its warnings. But once the error is fixed, the warnings will be checked for and shown if found.

Basically, the warnings of a certain file will be shown only when it has no errors, meaning the structure is correct. 
What do you think?

visualization/app/codeCharta/util/fileValidator.ts
IhsenBouallegue: Fixed in next commit

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -25,29 +25,19 @@ export class CodeChartaService {
...
+
+		this.getValidationResults(nameDataPairs, fileValidationResults)
 
-		if (this.fileStates.length > 0) {
+		this.storeService.dispatch(setIsLoadingFile(false))
+
+		if (
+			fileValidationResults.flatMap(validation => validation.errors).length > 0 ||
+			fileValidationResults.flatMap(validation => validation.warnings).length > 0
+		)
```
ce-bo, line 37: It would move line 30 into getValidationResults(). Thus, you could just return an empty array (on no errors) or a filled one with errors or warnings. Line 34-37 would then just be a check on `fileValidationResults.length > 0` without the two flatMaps.

visualization/app/codeCharta/codeCharta.service.ts
```diff
@@ -25,29 +25,19 @@ export class CodeChartaService {
...
+			await this.dialogService.showValidationDialog(fileValidationResults)
+
+		if (this.recentFiles.length > 0) {
```
ce-bo, line 40: It looks like that recentFiles is not set anymore,

visualization/app/codeCharta/ui/dialog/dialog.service.ts
```diff
@@ -38,26 +38,45 @@ export class DialogService {
...
+
+		if (fileValidationResults.flatMap(validation => validation.errors).length > 0) {
+			htmlMessages.push("<h2>Errors</h2>")
+			for (const fileValidationResult of fileValidationResults) {
+				if (fileValidationResult.errors.length > 0) {
+					const fileErrorMessage = this.buildFileErrorMessage(fileValidationResult)
+					htmlMessages.push(fileErrorMessage)
+				}
+			}
+		}
```
ce-bo, line 52: I would not iterate twice over the validation results. Here a suggestion to save one iteration:
```suggestion
		const errors = fileValidationResults.filter((validationResult) => { return validationResult.errors.length > 0 })
		if (errors .length > 0) {
			htmlMessages.push("<h2>Errors</h2>")
			for (const error of errors ) {
					const fileErrorMessage = this.buildFileErrorMessage(error )
					htmlMessages.push(fileErrorMessage)
			}
		}
```

ce-bo (commented):
Would you please fix the two failing quality gates and check my last comments? :-)

IhsenBouallegue:
> Would you please fix the two failing quality gates and check my last comments? :-)

I am working right now on the tests. Not my strong suit. But as soon as I get everything green, I will look into the comments! :D

visualization/app/codeCharta/codeCharta.service.ts
IhsenBouallegue: it is being set in method addFile

visualization/app/codeCharta/ui/dialog/dialog.service.ts
IhsenBouallegue: Doesn't filter iterate over the array too? Or is it less expansive than flatmap?

ce-bo: Yes, it does. But your second loop in line 46 will iterate all validation results again (instead of just the errors). Furthermore, it must be checked if there are any errors present a second time  .

visualization/app/codeCharta/codeCharta.service.ts
IhsenBouallegue: Resolved in the next commit

visualization/app/codeCharta/ui/dialog/dialog.service.ts
IhsenBouallegue: Resolved in the next commit

ce-bo (approved):
Code LGTM! :-)

PR #2608, opened 2022-01-15
refactor: move data of right clicked node into store

shaman-apprentice:
refactor: move data of right clicked node into store as the sync between AngularJS, Angular and document dispatching was quite hard to reason about. This helps migrating nodeContextMenu.component.


visualization/app/codeCharta/ui/nodeContextMenu/nodeContextMenu.component.ts
```diff
@@ -57,29 +50,37 @@ export class NodeContextMenuController implements ShowNodeContextMenuSubscriber,
...
 		"ngInject"
-		NodeContextMenuController.instance = this
 		MapColorsService.subscribe(this.$rootScope, this)
-		document.addEventListener("building-right-clicked", this.onBuildingRightClicked)
-		NodeContextMenuController.subscribeToShowNodeContextMenu(this.$rootScope, this)
-		NodeContextMenuController.subscribeToHideNodeContextMenu(this.$rootScope, this)
+		Store.store.subscribe(() => {
+			const state = Store.store.getState()
+			const rightClickedNodeData = rightClickedNodeDataSelector(state)
+			if (this.rightClickedNodeData === rightClickedNodeData) return
```
Hall-Ma, line 57: What must happen for this condition to become true?

shaman-apprentice: `rightClickedNodeDataSelector` calculates a memorized value. If the **related** input of state doesn't change, the same memorized  value / object is returned. Otherwise a new value.

Hall-Ma (approved):

PR #2615, opened 2022-01-19
fix: node retrivial in case it has no building

shaman-apprentice:
what can be "achieved" through clicking in map tree view

ref #2318


ce-bo (approved):

BridgeAR:
@shaman-apprentice would you be so kind and add a test case as well? :-)

PR #2622, opened 2022-01-23
Refactor/2318/improve custom dispatch 4

shaman-apprentice:
refactor: introduce `TrackEventUsageDataEffect` for:

- Better separation of concerns
- Reduce custom dispatch logic, which impedes Angular migration

Note that initial calls are only almost the same. I guess this is fine? @ce-bo lets make an appointment for a discussion and review :)


shaman-apprentice:
Why do we need `onColorRangeFromUpdated` and `onColorRangeToUpdated` if we already fire on `ColorRangeActions`?

ce-bo (approved):
Code LGTM! Nice clean ups :-)

PR #2625, opened 2022-01-24
Feature/2489/Rename custom config to custom view only visible for user

Hall-Ma:
# Rename custom config to custom view

Ref: #2489

## Description

- Rename all visible fields of custom configs to custom views


IhsenBouallegue (approved):
LGTM 🥳

Issue #697, opened 2019-08-16
Invert AreaMetric


alschmut:
# Feature request

## Description

We have options to invert the height and the color to adapt to metrics like test-coverage, which are interpreted to be better the higher the values are. If I want to set a metric like this to areaMetric, I would like to invert the amount of area, which is used for the nodes.

## Required steps
- [x]  Implement feature in AngularJS
- [x] Migrate feature to Angular

BridgeAR:
This could be done by using `<ctrl> + <left click>` on a building or folder. That should be pretty straight forward.

PR #2626, opened 2022-01-24
add invert area #697

Shivan-997:
# add invert the amount of area, which is used for the nodes

issue: #697

## Screenshots or gifs
image

### DIFF
--- a/visualization/app/codeCharta/util/algorithm/treeMapLayout/treeMapGenerator.ts
+++ b/visualization/app/codeCharta/util/algorithm/treeMapLayout/treeMapGenerator.ts
@@ -248,19 +248,17 @@ function isOnlyVisibleInComparisonMap(node: CodeMapNode, dynamicSettings: Dynami
 }
 
 // Only exported for testing.
-export function calculateAreaValue(node: CodeMapNode, state: State, maxWidth: number) {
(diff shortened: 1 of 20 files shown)


visualization/app/codeCharta/util/algorithm/treeMapLayout/treeMapGenerator.ts
```diff
@@ -241,17 +248,19 @@ function isOnlyVisibleInComparisonMap(node: CodeMapNode, dynamicSettings: Dynami
 }
 
 // Only exported for testing.
-export function calculateAreaValue(node: CodeMapNode, { dynamicSettings }: State) {
+export function calculateAreaValue(node: CodeMapNode, state: State, maxWidth: number) {
```
ce-bo, line 251: I would keep the deconstruction of `dynamicSettings`. You can do this wihtin the method signature (parameter list) or within the method body. Example for the former one:
```suggestion
export function calculateAreaValue(node: CodeMapNode, { dynamicSettings, appSettings }: State, maxWidth: number) {
```

ce-bo (commented):
This is actually a new component. So, we should migrate this to Angular.
I would be fine to merge this PR (but do not close the issue) and to create another follow up PR for the migration.

CodeCharta (approved):

PR #2643, opened 2022-02-06
Refactor/2318/id to node service

shaman-apprentice:
extract idToNode via a selector as store should only contain atomic non derivable data.

This also "migrates" / removes idToNode.service


visualization/.eslintrc.js
```diff
@@ -1,86 +1,87 @@
 module.exports = {
```
IhsenBouallegue, line 1: It seems like nothing changed here. Was this a formatting/linting mistake? or was a specific rule changes?

IhsenBouallegue (approved):
LGTM!

visualization/.eslintrc.js
shaman-apprentice: I added `"unicorn/no-useless-undefined": "off"`. Otherwise e.g. `function foo(a: string | undefined)` is always unhappy as `foo()` is missing one parameter and `foo(undefined)` has an "usless" undefined. In my strong opinion second call is the correct way, so I disabled this rule.

Issue #2617, opened 2022-01-20
Import GameObjects File


ce-bo:
# Feature request

## Description

As a User, I want to import the attached Game Object Json File ([gameObjectPositions.zip](https://github.com/MaibornWolff/codecharta/files/7906328/gameObjectPositions.zip)), so that I can explore them in the 3D visualization.

## Acceptance criteria
- [x] A prototype script can import the game objects of the attached file: [gameObjectPositions.zip](https://github.com/MaibornWolff/codecharta/files/7906328/gameObjectPositions.zip)
- [x] The prototype creates and exports a valid .cc.json file
- [x] The .cc.json file can be imported and visualized in CodeCharta Visualization
- [x] The prototype can import the dependencies included in the attached file

## Assumptions & Exclusions
- Dependency Information will be attached soon.
- The GameObject JSON Files will include Java Classes only
- Class Names do not have dots included

## Development notes (optional Task Breakdown)
- It would be sufficient to write a prototype script (e.g. written in Node.js) to import the data
- The x, y and z coordinates and the scaling must be transformed into our format

## Screenshots from GameObject Tool
![image](https://user-images.githubusercontent.com/26900540/150517703-94f6e911-fdcb-4e54-bb33-a18ea988ccf0.png)
![image](https://user-images.githubusercontent.com/26900540/150517709-d09a3a8d-081a-47c1-b9ec-2ba820038827.png)

IhsenBouallegue:
What are the dependencies? Can anyone clarify that a bit?

IhsenBouallegue:
Should it be called GameObject? Or should we find a more descriptive name?

ce-bo:
@IhsenBouallegue The dependencies list will be a from-to list. This relationships are used to draw the Edges/Arrows in CodeCharta. E.g. it is used to depict that two buildings are coupled due to any reason. As soon as the JSON strucutre and one example for the list is provided, we can start to implement this as a second step.

The naming of GameObject is not final yet. It could change in the future but for the Prototype I would keep it, because it is a known term.

IhsenBouallegue:
Another question about the coordinates. The x, y and z are in relation to what? Where is the origin point of the map. And do they tell us the position of the middle of the buildings or the corner?

IhsenBouallegue:
Yet another question,
[...]

PR #2646, opened 2022-02-07
Feature/2617/import gameobjects file

IhsenBouallegue:
# Import GameObjects file

Close: #2617

## Description
  - Parse GameObjects file structure into cc.json
  - Import GameObjects file on upload

## Screenshots or gifs
partial_Output cc(1)


visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
```diff
@@ -38,7 +40,10 @@ export class FileChooserController {
 				}
 
 				reader.onload = event => {
-					content = (isCompressed ? zlib.unzipSync(Buffer.from(<string>event.target.result)) : event.target.result).toString()
+					const result: string = event.target.result.toString()
```
NearW, line 43:
```suggestion
					const result = event.target.result.toString()
```
I'd recommend to use type inference as much as possible. Especially if you call a function like `toString()` it's quite obvious, what the type is.

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.spec.ts
```diff
@@ -163,4 +164,5 @@ describe("fileChooserController", () => {
 			expect(fileChooserController["files"][1].content.fileChecksum).toBe("a35dd32bd850841c47894ac9b8e1287a")
 		})
 	})
+	describe("", () => {})
```
NearW, line 167: Remove empty test

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
+import md5 from "md5"
+
+function isFile(names) {
+	return names.length === 0
+}
```
NearW, line 5: I assume that names is a list of strings. This file is missing type declarations on each function parameter.
We lose our typescript advantages, if we don't specify the types here.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+
+function addNodeRecursively(names, nodes, gameObjectPosition, parentNodeName, gameObjectPositions) {
+	if (names.length === 0) return
+	// get current path name
+	const name = names.shift()
+	let node
+
+	// create node
+	node = { name, type: isFile(names) ? "File" : "Folder", attributes: {} }
```
NearW, line 18:
```suggestion
	// create node
	let node = { name, type: isFile(names) ? "File" : "Folder", attributes: {} }
```

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+		node.fixedPosition = { left, top, width, height }
+	}
+	if (name === "root") {
+		node.fixedPosition.top = Math.floor(50 - node.fixedPosition.height / 2)
+		node.fixedPosition.left = 0
+	}
+
+	const newParentName = parentNodeName === "base" ? node.name : `${parentNodeName}.${node.name}`
+	addNodeRecursively(names, node.children, gameObjectPosition, newParentName, gameObjectPositions)
+	return
```
NearW, line 56: obsolete empty return at the end of a function

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+			z: longEdge
+		}
+	})
+
+	for (const gameObjectPosition of gameObjectPositions) {
+		const names = gameObjectPosition.name.split(".")
+		if (names[0] !== "base") addNodeRecursively(names, nodes[0].children, gameObjectPosition, "base", gameObjectPositions)
+	}
+
+	if (cycles)
```
NearW, line 129: I'd use a default value to an empty list for cycles in case this is null. You can simply call a for loop then, that will be skipped if cycles is empty.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+		for (const cycle of cycles) {
+			edges.push(createEdge(cycle))
+		}
```
NearW, line 132:
```suggestion
		const edges = cycles.map(cycle => createEdge(cycle))
```
Always ask yourself, if you can use map or filter, if you use something like `push` in a for loop. Array functions are lovely!

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+
+function createAttributeTypes() {
+	return {
+		edges: {
+			coupling: "relative"
+		}
+	}
+}
+
+export function parseGameObjectsFile(data) {
```
NearW, line 82: Move public accessible functions (export) to the top

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+		checksum: "",
+		data: {
+			projectName: "ScriptProjectName",
+			fileChecksum: "",
+			apiVersion: "1.3",
+			nodes: []
+		}
+	}
+
+	const nodes = []
```
NearW, line 95: nodes never contains more than one object in this function (dummyNode). You can just create the list with that element in it.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function nodeAlreadyExists(nodes, name) {
+	return nodes.some(node => node.name === name)
+}
+
+function addNodeRecursively(names, nodes, gameObjectPosition, parentNodeName, gameObjectPositions) {
+	if (names.length === 0) return
+	// get current path name
+	const name = names.shift()
```
NearW, line 14:
```suggestion
	const [name] = names
```
You should check out array destructuring! This retrieves the first element WITHOUT mutating the original array. `names` is passed down again later one and it's not that obvious, that you removed the first element here. In general it's not a good idea modify function parameters. This can result in really nasty bugs.
Instead create new arrays and pass them down. I'd do this call at the end then:

```
	addNodeRecursively(names.slice(1), node.children, gameObjectPosition, newParentName, gameObjectPositions)
```

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.ts
IhsenBouallegue: Thanks :D !
Corrected in the next commit.

visualization/app/codeCharta/ui/fileChooser/fileChooser.component.spec.ts
IhsenBouallegue: Oh, slipped under the radar!
Corrected in the next commit.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: If I declare it there. I need to turn node.children to node["children"]. Is that a bad practice?

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: Thanks!
Corrected in the next commit.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: That wouldn't have occurred to me in a million years. Thank you for your input!
Will fix it in the next commit

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: I found another way
```suggestion
	// create node
	let node = { name, type: isFile(names) ? "File" : "Folder", attributes: {}, children: undefined}
```
Then I can assign children if it's needed (in case of a folder) let me know if this is not good practice.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: That saved many lines.
Corrected in the next commit.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: Thank you for the tip! Will use them more often 😄

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: You are totally right!
Corrected in the next commit.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: Hey, a short follow-up question. The rest of the method relies on the array becoming shorter for other checks. So after taking the first element it shouldn't be in the array anymore. Should we do `const [name] = names; names = names.slice(1)` But isn't that the same as `const name = names.shift()`? Or should I copy it and work with a copy?

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function nodeAlreadyExists(nodes: CodeMapNode[], name: string) {
+	return nodes.some(node => node.name === name)
+}
+
+function wrapFileInAFolder(nodeName: string, node: CodeMapNode, gameObjectPosition) {
+	const childNode = { ...node }
+	childNode.attributes = { height: gameObjectPosition.scale.y }
+	const ParentNode: CodeMapNode = { name: nodeName, type: NodeType.FOLDER, attributes: {}, children: [childNode] }
```
ce-bo, line 73: Variables should be named in camelCase.
```suggestion
	const parentNode: CodeMapNode = { name: nodeName, type: NodeType.FOLDER, attributes: {}, children: [childNode] }
```

visualization/package.json
```diff
@@ -190,7 +190,7 @@
 		"typescript": "^4.3.5",
 		"webpack": "^5.33.2",
 		"webpack-cli": "^4.7.2",
-		"webpack-dev-server": "^4.0.0",
+		"webpack-dev-server": "3.11.2",
```
ce-bo, line 193: Was that your intention?

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: Slipped under the radar. Will correct it

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,136 @@
...
+		const width = Math.floor((childGameObject.scale.z / parentGameObject.scale.z) * 100)
+		const height = Math.floor((childGameObject.scale.x / parentGameObject.scale.x) * 100)
+		position = { left, top, width, height }
+	}
+	if (name === "root") {
+		position.top = Math.floor(50 - position.height / 2)
+		position.left = 0
+	}
+	return position
+}
```
ce-bo, line 95: Let's add some short explanation why the division by 2 is necessary and why position.top is calculated as it is.
It's hard to remember the reasons for those calculations.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsSchema.json
```diff
@@ -0,0 +1,271 @@
...
+										"examples": [228.5]
+									}
+								},
+								"additionalProperties": true
+							},
+							"scale": {
+								"$id": "#/properties/gameObjectPositions/items/anyOf/0/properties/scale",
+								"type": "object",
+								"title": "The scale schema",
+								"description": "An explanation about the purpose of this instance.",
```
ce-bo, line 167: I am not sure about the purpose of the descriptions. Should we adjust them?

IhsenBouallegue: Hm, it doesn't make a huge difference, but better have a descriptive description. I will adjust it.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: I thought the variable name is descriptive enough. But you are right better have a more detailed description in this case. 
Will add a few comments 👍

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	// get current node name and create cc node structure
+	const [nodeName] = nodeNames
+	let node: CodeMapNode = {
+		name: nodeName,
+		type: isFile(nodeNames) ? NodeType.FILE : NodeType.FOLDER,
+		attributes: {}
+	}
+	if (!isFile(nodeNames)) node.children = []
+	if (isFile(nodeNames)) node = wrapFileInAFolder(nodeName, node, gameObjectPosition)
```
cschild, line 46:
```suggestion
	if (!isFile(nodeNames)) node.children = []
	else node = wrapFileInAFolder(nodeName, node, gameObjectPosition)
```

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+function calculateFixedFolderPosition(node: CodeMapNode, name: string, parentGameObject, childGameObject) {
+	let position: FixedPosition
+	if (node.type === NodeType.FOLDER && name !== "base") {
+		// translate the center point from the middle of the gameObject to the corner as needed for fixedPosition
+		const cornerXofParent = parentGameObject.position.x - parentGameObject.scale.x / 2
+		const cornerZofParent = parentGameObject.position.z - parentGameObject.scale.z / 2
+		const cornerXofChild = childGameObject.position.x - childGameObject.scale.x / 2
+		const cornerZofChild = childGameObject.position.z - childGameObject.scale.z / 2
```
cschild, line 84: Maybe the translation could be move in a function?
```
translatePoint(position, scale) {
     position - scale / 2
}
```

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,137 @@
...
+	const childNode = { ...node }
+	childNode.attributes = { height: gameObjectPosition.scale.y }
+	const parentNode: CodeMapNode = { name: nodeName, type: NodeType.FOLDER, attributes: {}, children: [childNode] }
+	return parentNode
+}
+
+function calculateFixedFolderPosition(node: CodeMapNode, name: string, parentGameObject, childGameObject) {
+	let position: FixedPosition
+	if (node.type === NodeType.FOLDER && name !== "base") {
+		// translate the center point from the middle of the gameObject to the corner as needed for fixedPosition
```
cschild, line 80: I'm not sure if it explains why we calculate top, bottom, width and height

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -178,6 +178,9 @@ function mergeFolderChain(data: CodeMapNode) {
 	// e.g. a /folder which includes anotherFolder that includes other files or folders
 	// will be merged to a node with path /folder/anotherFolder and children are set accordingly
 
+	//TODO: We should discuss if we want this behavior in general for now it's turned of for fixed folders
+	if (data.children?.length === 1 && data.children[0]?.fixedPosition) return
+
```
Hall-Ma, line 183: @ce-bo Do you think we can leave it like that? I would like to know your thoughts on this.

ce-bo: I agree that folders may not be chained, if they have fixed positions.
If I am correct, all folders in the same level must have fixed positions as soon as one folder is fixed.
Thus, it looks good for me.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,137 @@
...
+			x: 0,
+			y: 0,
+			z: 0
+		},
+		scale: {
+			x: longEdge,
+			y: 1,
+			z: longEdge
+		}
+	}
```
ce-bo, line 136: I would define an interface for this structure.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -0,0 +1,137 @@
...
+		const cornerXofChild = childGameObject.position.x - childGameObject.scale.x / 2
+		const cornerZofChild = childGameObject.position.z - childGameObject.scale.z / 2
+		const top = Math.floor(((cornerXofChild - cornerXofParent) / parentGameObject.scale.x) * 100)
+		const left = Math.floor(((cornerZofChild - cornerZofParent) / parentGameObject.scale.z) * 100)
+		const width = Math.floor((childGameObject.scale.z / parentGameObject.scale.z) * 100)
+		const height = Math.floor((childGameObject.scale.x / parentGameObject.scale.x) * 100)
+		position = { left, top, width, height }
+	}
+	if (name === "root") {
+		position.top = Math.floor(50 - position.height / 2)
```
ce-bo, line 92: What is the reason for this calculation? Maybe we can extract and move it to a separate speaking method?

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
ce-bo: I would prefer the inlined (current) version. The calculations might assume complex/confusing and currently you are not forced to analyse further methods to understand the code (which is good in this case).

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
Hall-Ma: We need this function to ensure that the root folder is in the middle of the "base plate".

ce-bo (approved):
Code LGTM. I also tested it locally and it works very nice.
Let's address the minor remarks before the merge.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
IhsenBouallegue: Take a look at our fixed folders' documentation, it will explain what we are doing and why 👍 
https://maibornwolff.github.io/codecharta/how-to/fixate_folders_with_a_custom_cc_json/

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
Hall-Ma: That's not how I see it. In principle you're right but in this case, we don't return anything, so we can't use the advantage of an early return. Additionally, it's not necessary to check if `nodeNames` is a file when we checked the other case before.

visualization/app/codeCharta/util/gameObjectsParser/gameObjectsImporter.ts
```diff
@@ -118,20 +152,19 @@ function createAttributeTypes() {
...
+		name: BASE_NAME,
 		position: {
 			x: 0,
 			y: 0,
 			z: 0
 		},
 		scale: {
 			x: longEdge,
-			y: 1,
+			y: 0,
```
ce-bo, line 166: Was that your intention?

Hall-Ma: We don't need a height. It's a folder and folders don't need heights. In addition, it has no effect if there is a 0, 50, 100, ...

PR #2660, opened 2022-02-10
Refactor/2318/migrate id to building service

shaman-apprentice:
Migrate idToBuilding.service.

Move it out of store, as they are not atomic data, but a combination of node data (which can be derived by idToNodeSelector) and threejs state.

Most changes are adoption of boiler code and adjusting tests.


IhsenBouallegue (approved):
LGTM 😁

Issue #2066, opened 2021-06-24
Remove None option from edge metric chooser


Christian-Eberhard:
# Feature request

## Description

As a user, I want to set edge metrics more intuitive so that i can work more efficiently. For that it
would help if "None" was not an option when im choosing an edge metric option. Currently "None"
is just an option like all other metrics.

![None.JPG](https://images.zenhubusercontent.com/6076b0525fa0ae9470963a22/f5dafe46-e1e4-4ad6-a0d9-3d6a2cb70edd)

Also edge metrics are shown in the node metric lists, they should be filtered.

## Acceptance criteria
- None removed from edge metric options
- Add option to deactivate the edge metric options in it's menu
- ~~No edge metrics in node metrics list~~
- The new disable option is covered by CustomViews.
- The disable option can be reset

## Development notes (optional Task Breakdown)
- [Remove none from the edge metric list] 
- [Create an option in the edge metric options, to disable them] 
- [Filter edge metrics in the node metric lists]

Issue #1542, opened 2020-12-02
Edge metric is only shown if there are activated edges


Gaugi28:
# Feature request

<Please note by far the quickest way to get a new feature is to file a Pull Request.>
<We will consider your request but it may be closed,>
<if it's something we're not actively planning to work on>
<or it does not match our visison of this software.>

## Description

As user, I want that the metric `Edges` disappears if there are no activated edges so that I can't get confused about the meaning of the metric `Edges`.

## Acceptance criteria
- if there are no activated edges the metric `Edges` is hidden

RomanenkoVladimir:
We show a tooltip and the metric appears grey.
I guess the idea was that the interface remains largely unchanged so that the user has a fast recognition factor as well as not being confused by disappering features.
The current implementation provides a feedback that the metric isn't just gone, it is disabled for the current map, with the tooltip that you need to generate edges in analysis.
What do you think if you few it under this aspect?

Gaugi28:
Yes, I know, but for me, as user who don't know, where I can generate edges in analysis, it's difficult to understand the tooltip. I discussed it yesterday with @BridgeAR. The result was: either the tooltip tells me how to generate the edges in analysis, or the metric is hidden from the beginning.

PR #2665, opened 2022-02-15
Feature/2066/Remove 'none' option from edge metric chooser

Hall-Ma:
# Refactor edge metric panel 

Closes: #2066
Closes: #1542

## Description

## Screenshots or gifs
image


BridgeAR:
> What do you think about this? Is that enough? When the edge metric is disabled, the text is just grey, but everything else is still working. This one is the easiest and fast solution to show the user that something has changed here.

Sounds good to me :+1:

PR #2679, opened 2022-02-21
Feature/2474/risk profile calculation for every language

Hall-Ma:
# Risk profile calculation for all occurring programming languages

Closes: #2427

## Description

- WIP

### DIFF
--- a/visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
+++ b/visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
@@ -40,6 +40,17 @@ interface MetricSuggestionParameters {
 	isOutlier?: boolean
 }
 
+interface RiskProfile {
+	lowRisk: number
+	moderateRisk: number
+	highRisk: number
+	veryHighRisk: number
+	totalRloc?: number
+}
+
+const HEIGHT_METRIC = "mcc"
+const AREA_METRIC = "rloc"
+
 export class ArtificialIntelligenceController
 	implements FilesSelectionSubscriber, BlacklistSubscriber, ExperimentalFeaturesEnabledSubscriber
 {
@@ -47,12 +58,7 @@ export class ArtificialIntelligenceController
 		analyzedProgrammingLanguage: string
 		suspiciousMetricSuggestionLinks: MetricSuggestionParameters[]
 		unsuspiciousMetrics: string[]
-		riskProfile: {
-			lowRisk: number
-			moderateRisk: number
-			highRisk: number
-			veryHighRisk: number
-		}
+		riskProfile: RiskProfile
 	} = {
 		analyzedProgrammingLanguage: undefined,
 		suspiciousMetricSuggestionLinks: [],
@@ -65,9 +71,17 @@ export class ArtificialIntelligenceController
 		}
 	}
 
-	private debounceCalculation: () => void
+	private rlocRisk = {
+		lowRisk: 0,
+		moderateRisk: 0,
+		highRisk: 0,
+		veryHighRisk: 0,
+		totalRloc: 0
+	}
+
 	private fileState: FileState
 	private blacklist: BlacklistItem[] = []
+	private debounceCalculation: () => void
 
 	constructor(private $rootScope: IRootScopeService, private storeService: StoreService) {
 		"ngInject"
@@ -145,7 +159,7 @@ export class ArtificialIntelligenceController
 		this.clearRiskProfile()
 
 		if (mainProgrammingLanguage !== undefined) {
-			this.calculateRiskProfile(this.fileState, mainProgrammingLanguage, "mcc")
+			this.calculateRiskProfile(this.fileState, HEIGHT_METRIC)
 			this.calculateSuspiciousMetrics(this.fileState, mainProgrammingLanguage)
 		}
 	}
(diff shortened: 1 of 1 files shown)


visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -40,34 +40,37 @@ interface MetricSuggestionParameters {
...
+	lowRisk: number
+	moderateRisk: number
+	highRisk: number
+	veryHighRisk: number
+	totalRloc?: number
+}
+
+const HEIGHT_METRIC = "mcc"
+const AREA_METRIC = "rloc"
+const EXCLUDED_FILES = new Set(["html", "sass", "scss", "scss", ".txt", "md", "json", undefined])
```
NearW, line 53:
```suggestion
const EXCLUDED_FILES = new Set(["html", "sass", "scss", ".txt", "md", "json", undefined])
```

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -139,80 +142,80 @@ export class ArtificialIntelligenceController
 			return
 		}
 
-		const mainProgrammingLanguage = this.getMostFrequentLanguage(this.fileState.file.map)
-		this._viewModel.analyzedProgrammingLanguage = mainProgrammingLanguage
+		this.setDefaultRiskProfile()
```
NearW, line 145: Instead of setting the default risk profile here, you can also set defaults within the viewModel-Construction.
Currently a cast is required, because you attempt to set the risk profile to a value, that should not be possible (empty object)

```ts
		riskProfile: {} as RiskProfile
```

Instead, set it straight to 
```ts
{ lowRisk: 0, moderateRisk: 0, highRisk: 0, veryHighRisk: 0 }
```

You won't need the cast then.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -40,34 +40,37 @@ interface MetricSuggestionParameters {
...
-			lowRisk: 0,
-			moderateRisk: 0,
-			highRisk: 0,
-			veryHighRisk: 0
-		}
+		riskProfile: {} as RiskProfile
 	}
 
-	private debounceCalculation: () => void
+	private rlocRisk = {} as RiskProfile
```
NearW, line 70: I don't understand yet why we have two RiskProfiles. We have one in the viewModel, which represents what we render at the end and we also have this class variable here.
I assume, that this class variable `rlocRisk` is just required for some internal calculations before transferring it to the viewModel. (Correct me if I'm wrong).

If that's the case, we should try to only work with the viewModel here.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -139,80 +142,80 @@ export class ArtificialIntelligenceController
...
+
+	private isFileInvalid(node: CodeMapNode, metricName: string) {
+		return (
+			node.type === NodeType.FILE &&
+			!isPathBlacklisted(node.path, this.blacklist, BlacklistType.exclude) &&
+			node.attributes[metricName] !== undefined &&
+			node.attributes[AREA_METRIC] !== undefined &&
+			!EXCLUDED_FILES.has(this.getFileExtension(node.name))
+		)
+	}
```
NearW, line 203: I can't really follow that expression. I assume, that a file is invalid, if it's a file and one of the following things are true:
- the file is excluded
- missing a specific metric
- missing the area metric
- the file extension is on our EXCLUDED_FILES blacklist

```ts
	private isFileInvalid(node: CodeMapNode, metricName: string) {
		return node.type === NodeType.FILE && (
			isPathBlacklisted(node.path, this.blacklist, BlacklistType.exclude) ||
			node.attributes[metricName] === undefined ||
			node.attributes[AREA_METRIC] === undefined ||
			EXCLUDED_FILES.has(this.getFileExtension(node.name))
		)
	}
```

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
NearW: Also I'd rename this to `EXCLUDED_FILE_EXTENSIONS`

Hall-Ma: We don't need the point in `.fileExtension` because it's sliced in getFileExtension(fileName: string) method and I don't the any point to change this method.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
```diff
@@ -139,80 +142,80 @@ export class ArtificialIntelligenceController
...
+
+	private calculateRlocRisk(nodeMetricValue: number, thresholds, nodeRlocValue: number) {
```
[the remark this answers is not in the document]
Hall-Ma: This type is `any`. Is that necessary to declare this type as parameter?

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
Hall-Ma: It was a renaming mistake. Actually this method should call `isFileValid`.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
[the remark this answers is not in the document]
NearW: `any` should be avoided as a type. If it depends depending on the input and you don't know what you'll expect, use `unknown` instead. However, you access `thresholds.percentile70` and other properties here. So it must have a deterministic type. :)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Added 🚀
```
ce-bo, line 10: Hmm, I would say that it is a Changed behavior.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.ts
NearW: I'm confused. :D 
a few things here, I'd change:
- Rename to `EXCLUDED_FILE_EXTENSIONS` since this is a set of file extensions. It's more expressive than just calling it `EXCLUDED_FILES`. A string is not a file.
- The element `scss` is listed twice. It can be removed once.
- The dot in `.txt` can be removed I guess.

Hall-Ma: You're right. I was a bit sloppy.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligence.component.spec.ts
```diff
@@ -287,52 +276,53 @@ describe("ArtificialIntelligenceController", () => {
...
-		})
-	})
-
-	it("should calculate risk profile only for valid file extensions", () => {
-		const FILE_STATES_OTHER = klona(FILE_STATES_JAVA)
-		for (const codeMapNode of FILE_STATES_OTHER[0].file.map.children[1].children) {
-			codeMapNode.name = codeMapNode.name.replace(/\.java/, ".html")
-		}
+			const fileStateJava = createFileStateWithNodes([{ name: "file1.java", type: NodeType.FILE, attributes: { rloc: 50, mcc: 1 } }])
+			const totalRocJava = fileStateJava.file.map.children.reduce((sum, child) => child.attributes.rloc + sum, 0)
```
ce-bo, line 282: totalRlocJava vs. totalRocJava

ce-bo (approved):
LGTM :-)

PR #2686, opened 2022-02-25
Refactor/2318/migrate search bar component

shaman-apprentice:


visualization/app/codeCharta/state/store.service.spec.ts
```diff
@@ -118,24 +116,5 @@ describe("StoreService", () => {
 
 			expect(storeService.getState().appSettings.isLoadingMap).toBeFalsy()
 		})
-
-		describe("connection to action subject", () => {
```
shaman-apprentice, line 122: Moved the logic and test into plain store, so that effects works also in tests, which don't inject AngularJS' store.service

shaman-apprentice:
the code smells are all from simple moving stuff without touching it. As I don't want to waste too much time into this internal code I ignored them.

PR #2698, opened 2022-03-01
Refactor/add eslint brace style rule

Hall-Ma:
# eslint brace style rule

## Description

- To further unify the code style of CodeCharta, the bracket rule is implemented.

## Examples
This is not possible anymore:
```
if (foo) foo++;
```

Now, we have to add braces to the conditions:
```
if (foo) {
    foo++;
}
```


ce-bo (commented):
Are those warnings new?
image

Hall-Ma:
@BridgeAR @ce-bo @shaman-apprentice 
I don't know what to do. There is a conflict between prettier, especially with `printwidth` and eslint `brace style`.  Eslint wants to have the braces in the same line but when the print width is exceeded then prettier puts the brace in a new line and there we have the problem. So far I couldn't find a solution. I tried to use this rule 
```
"max-len": [2, { code: 140, tabWidth: 4, ignoreStrings: true, ignoreTrailingComments: true, ignoreComments: true }],
```
to have the same condition in prettier and eslint but it doesn't bring the desired solution. Any idea?

NearW (approved):

Issue #2692, opened 2022-02-26
Close ribbon bar things on outside clicks


shaman-apprentice:
# Feature request

As an user I expect the search bar and dropdowns of the ribbon bar to close, when I click outside of any of them, to have more map vision.

Right now search bar doesn't close when for example clicking on the legend panel or on a chose metric selection box.

Opening a chose a metric select box doesn't close a settings-panel.

![image](https://user-images.githubusercontent.com/3596742/155852061-09896722-82e7-4936-902f-ec50803c12fa.png)

## Acceptance criteria
- When opening any kind of dropdown in the ribbon bar and clicking outside of the opened dropdown, it closes.

## Development notes (optional Task Breakdown)
- We can add a general document click lister when opening, which closes the dropdown on outside click and removes itself, when dropdown is closed again. Or facilitate mat-select which has that out of the box.
- Then we can remove `searchPanelMode` from store including `searchPanelMode.service`.
- Then we can remove `panelSelection` from store including `panelSelection.service`.
- Then we can remove `codeCharta.mouseEvent.service.ts`.

This makes the code leaner and facilitates migration to Angular, as we don't have to migrate, what we have removed.

PR #2700, opened 2022-03-03
Refactor/2318/migrate search panel mode selector

shaman-apprentice:
Early merge regarding migration of search-panel-mode-selector-component (#2318) and always close search panel on outside click (#2692)


shaman-apprentice:
supercede by https://github.com/MaibornWolff/codecharta/pull/2702 which contains same changes and one additional bug fix.

Issue #95, opened 2018-03-02
Project / Directory summary


ukinimod:
It would be Great to have some summary in the visualization concerning the number of Buildings in a package. The number of changed, added, deleted buildings in Delta mode. This would have helped with debugging. What do you think?

mrpatpat:
I think this would be a nice feature with low implementation costs. The new algorithms used for #91  already have distinct branches for added/deleted/modified files. Adding flags to a node and counting them should be a rather simple implementation. 
You can expect it after the delta issues #91 #92 #93 are resolved.

BridgeAR:
The number of total files is already included.

The delta mode does not yet show the number of deleted or added files. That would be a nice addtive.

It could roughly look like that: 
![image](https://user-images.githubusercontent.com/8822573/141840877-6cd00fbc-523e-4477-bcf3-cfa7cbfdea2f.png)

PR #2701, opened 2022-03-03
Implement counting removed or added files in delta mode

Hall-Ma:
# Counting removed or added files in delta mode

Closes: #95

## Description

- In delta mode added and removed files are displayed in the attribute sidebar when a folder is selected

## Screenshots or gifs
image


visualization/app/codeCharta/util/deltaGenerator.ts
```diff
@@ -12,15 +12,27 @@ export class DeltaGenerator {
...
 			} else {
-				const parentNode = getParent(hashMapWithAllNodes, path)
-				parentNode.children.push(node)
+				const childNode = node
+				let parentNode = getParent(hashMapWithAllNodes, path)
+				const originalParent = parentNode
+				const added = childNode.deltas["addedFiles"] ?? 0
+				const removed = childNode.deltas["removedFiles"] ?? 0
+
+				while (parentNode) {
```
BridgeAR, line 21: This will likely work, but it's not ideal performance wise. It should be possible to count together all files in a linear fashion when doing this in a depth first for loop.

Hall-Ma: Any explicit suggestions are welcomed. I cannot find a better solution at this time.

BridgeAR:
This is already looking great!

We should include the number of changed files as well (this could be done in a follow-up PR by adding a blue colored delta for the number of files that have attributes that changed. The information would be inside `getDeltaAttributeList()`: if any attribute attribute is non zero, it's a changed file).

visualization/app/codeCharta/util/deltaGenerator.ts
```diff
@@ -89,7 +109,8 @@ export class DeltaGenerator {
 				deltaAttribute[key] = -referenceAttribute[key]
 			}
 		}
-		return deltaAttribute
+
+		return { ...deltaAttribute, addedFiles: 0, removedFiles: 0 }
```
BridgeAR, line 113: Instead of copying the deltaAttributes here, it's possible to set these up front in [line 98](https://github.com/MaibornWolff/codecharta/pull/2701/files#diff-5a4760ef17d74c30e6235c456ecbf0bf5754ff461c369b3506cd5fb1e294f04fR98).

Please be aware that adding values like these mean that it's not possible to use the same names as metric name and we do not control what metric names are loaded into CodeCharta. It's rather unlikely that anyone uses these names and therefore it's fine to implement it like this for now.

Christian-Eberhard (approved):

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/fileCountDescription.selector.spec.ts
```diff
@@ -1,23 +1,18 @@
-import { node2fileCountDescription } from "./fileCountDescription.selector"
```
ce-bo, line 1: Please rename the file to match the text file name.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBarHeaderSection/nodePath/nodePath.component.ts
```diff
@@ -1,19 +1,21 @@
...
 })
 export class NodePathComponent {
 	@Input() node?: Pick<CodeMapNode, "path" | "children">
-	fileCountDescription$: Observable<string | undefined>
+	fileCountDescription$: Observable<FileCounter | undefined>
+	isDeltaMode$: Observable<boolean>
 
 	constructor(@Inject(Store) store: Store) {
-		this.fileCountDescription$ = store.select(fileCountDescriptionSelector)
+		this.fileCountDescription$ = store.select(fileCountSelector)
```
ce-bo, line 18: The word `Description` is obsolete.

visualization/app/codeCharta/util/nodeDecorator.ts
```diff
@@ -99,7 +100,9 @@ export const NodeDecorator = {
 					setNodeMediansToParent(medians, selector, parentSelector, data, name, isDeltaState)
 					collectNodeMediansOnParent(medians, parentSelector, data, name, isDeltaState)
 				} else {
-					parent.data.attributes[name] += data.attributes[name]
+					if (parent.data.attributes[name] !== undefined) {
+						parent.data.attributes[name] += data.attributes[name]
+					}
```
ce-bo, line 105: I am not sure why this is needed. I would expect that both (the parent and children) have the same set of attributes.
If this case occurs in delta mode, then propably some attributes of a (parent) folder would be missing.

ce-bo (commented):
Generally LGTM!

I think there is a bug in the UI (Windows latest Chrome):
image
And the neon green color is hard to read. Would you be so kind and change it a bit?

visualization/app/codeCharta/util/nodeDecorator.ts
ce-bo: Now I understand. Ignore my comment. I was wrong. :-)

ce-bo (approved):

PR #2709, opened 2022-03-09
WIP Tech/1609/margin issues

RomanenkoVladimir:
# Fix margin issue

_**Please read the [CONTRIBUTING.md](https://github.com/MaibornWolff/codecharta/blob/main/CONTRIBUTING.md) before opening a PR.**_

Issue: #1609

## Description

- Fixed treemap algorithm
- Fixed height issues with labels when changing map height
- Fixed height issues with new algorithm on startup

Missing:
- Tests
- Issue with Floor-Labels
- Issue with Big maps not rendering correctly due to culling distance
- Merge main, last commit was reverted due to errors a few weeks ago, hence the amount of changed files

Issue #1609, opened 2020-12-18
Bug collections concerning map rendering and margins


ce-bo:
# Bug

- When changing the margins, the height of the building should not change.
- Refactor how we zoom the map. Currently we move the "Kegel" to zoom in and out but threejs can do that by its own (pass the zoom to threejs).
- treeMapGenerator.js: Do not estimate the map size, calculate it: read the view port size substract some pixels to have space between the view port and the drawn map and consider the current zoom factor. The resulting numbers should be passed to threejs, so that the map fits perfectly into the view port.
- By default all buildings should be visible (this can be achived by modifying the zoom level)

@RomanenkoVladimir @BridgeAR 

## Open TODOs before merge:
- [x] Re-enable Floor Labels
- [x] Add/Expand Unit Tests for area calculation
- [ ] #3146

## Bugs (currently in main):
- Fix "floor labels next to the maps" after changing height scaling #2611 .
- [ ] #3145

Med-H:
I think the margin problem could be solved when using the d3 hierarchy padding property 
https://github.com/d3/d3-hierarchy#treemap_padding

in this page you can check it in action
https://observablehq.com/@d3/animated-treemap

ce-bo:
I will check the open tasks after the next release. For now, I decrease the priority because the urgent tasks have been addressed.

ce-bo:
I close this issue since the main (urgent) parts have been solved. 🥳 
Open issues are tracked by #3145 and #3146 .

ce-bo:
Thank you @BridgeAR. I just added two new issues #3148 and #3147 with different priorities to track point 2 and 4.

PR #2720, opened 2022-03-15
fix: apply changes from PR 2709 to main

shaman-apprentice:
I skimmed the changes from #2709 and copied those into latest main, which seemed relevant.

No warranty that I got everything correctly. It is not clear to me, if those changes were already finished. On a first glance I see the following todos; tests needs to be adjusted, due to changes of calculated `y` values and how often `addLeafLabel` is called:

- [ ] test everything manual
- [ ] adjust codeMap.label.service.spec.ts
- [ ] adjust codeMap.render.service.spec.ts
- [ ] adjust threeSceneService.spec.ts
- [ ] do we want to keep added testMap.cc.json?
- [ ] add a changelog entry

ref #1609
ref #2709

### DIFF
--- a/visualization/app/codeCharta/assets/testMap.cc.json
+++ b/visualization/app/codeCharta/assets/testMap.cc.json
@@ -1,0 +1,114 @@
+{
+	"projectName": "Sample Project with Edges",
+	"apiVersion": "1.2",
+	"fileChecksum": "invalid-md5-sample",
+	"nodes": [
+		{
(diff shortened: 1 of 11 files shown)


BridgeAR (changes requested):
I believe important parts are missing from #2709. Especially the part from `visualization/app/codeCharta/util/algorithm/treeMapLayout/nodeAreaCalculator.ts? .

Issue #2710, opened 2022-03-10
Expand calculations for suspicious metrics and risk profile in multiple (now: standard) mode


Hall-Ma:
# Feature request

## Description

As a user, I select multiple files in the standard mode so that I get a valid risk profile and suspicious metrics of the shown map.
Currently, suspicious metrics and risk profile are only calculated for the first selected file in the list. 

## Acceptance criteria
- When multiple files are selected suspicious metrics and risk profile are calculated for shown map.

## Development notes (optional Task Breakdown)
- [x] filter file state for all currently selected files
- [x] find most frequently used language across all selected files
- [x] calculate suspicious metrics and risk profile across all selected files

Issue #2589, opened 2022-01-03
Suspicious Metrics - Broken menu


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN
loading firefox.cc.json with kind of exotic metrics

THEN
Suspicious Metrics menu should be grayed out (it should not be clickable) and show a message on hover (a regular title behavior, showing the message after a short delay) that it's not available, if no metrics have been calculated due to not matching any known metric types. Show a short message that no suspicious metrics have been found if that's the case (only inconspicuous metrics were found).

## Actual Behavior
Suspicious Metrics menu is broken

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/147918662-5c42568f-d5d7-4f03-896a-40f7e6598244.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.86.0
- OS: all
- Browser: all

BridgeAR:
I updated the above to behave a bit different depending on two different scenarios.

Hall-Ma:
Solved in #2721

PR #2721, opened 2022-03-16
Feature/2710/calculation of suspicious metrics in standard mode

Hall-Ma:
# Expand calculations for suspicious metrics and risk profile in standard mode

closes: #2710, #2589

## Description

- all selected files/maps will be included in calculations of risk profile and suspicious metrics
- additionally, artificial intelligence component is decluttered and calculations are splitted in different files
- when no programming language was found the user gets a message when clicking on suspicious metrics button


ce-bo (commented):
Concerning a possible performance optimization: I think we could run the two loops (fileStates, hierachy(map)) only once, if we introduce a kind of preprocessing. It reminds me a bit of the visitor pattern.
Idea: The loops are defined in one place, looping over every node. For the calculations of Risk Profile, getSortedMetricValues, getMostFrequentLanguage you could introduce three new "Processor" classes and pass the `data` variable as a function parameter to a corresponding function (e.g. `processNode`). Every Processor will be called for every node and can apply their own processing rules. They must hold their own state, so that you can access the final calculations at the end.

BridgeAR:
@ce-bo I personally would just change the internal multiple model. Let's just combine the maps and just handle what maps are considered to be part of the new internal map in one place. That way all other logic just keeps on working properly, no matter if multiple mode is used or not..

Currently:

```
// input = map1, map2, map3, map4. map4 is not active
const maps: HierarchyNodes = [map1, map2, map3]

for (const map of maps) {
  // logic for internal maps
}
```

Future:

```
class HierarchyNodes {
  constructor(...children) {
    this.newRootNode = {
      children,
      /* other properties */,
    }
  }
}

const maps: HierarchyNodes[] = [map1, map2, map3, map4] // map4 is not active
const internalMap: HierarchyNodes = new HierarchyNodes(map1, map2, map3)

// logic for internalMap
```

Hall-Ma:
Correct if I'm wrong, but this already happens in aggregationGenerator?

BridgeAR:
@ce-bo AFAIC we do not have to copy any data at all. It should just be processed at the right time. This is an optimization and we can not stop cloning right away.

We have to keep the original input data around somehow. We might want to check how much memory overhead we have by just loading (not processing) multiple big maps. We could limit the memory usage to just parse the required part and keep the original input compressed around. That would require more processing while lowering the memory footprint. This is therefore a trade-off.
As soon as the map is loaded, transform it into the required internal structure by iterating over the input tree and create the new structure on the go without cloning the input.

visualization/app/codeCharta/ui/artificialIntelligence/ArtificialIntelligenceHelper/riskProfileHelper.ts
```diff
@@ -0,0 +1,48 @@
...
+	if (nodeMetricValue <= thresholds.percentile70) {
+		return (rlocRisk.lowRisk += nodeRlocValue)
+	}
+	if (nodeMetricValue <= thresholds.percentile80) {
+		return (rlocRisk.moderateRisk += nodeRlocValue)
+	}
+	if (nodeMetricValue <= thresholds.percentile90) {
+		return (rlocRisk.highRisk += nodeRlocValue)
+	}
+	return (rlocRisk.veryHighRisk += nodeRlocValue)
```
BridgeAR, line 36: I know this code is only moved, it would however be good to update the code to signal that the metrics are not meant to be returned and instead manipulate the input object.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/selectors/artificialIntelligence.selector.ts
```diff
@@ -0,0 +1,103 @@
...
+	const metricValues: MetricValues = {}
+	const metricValuesByLanguage: MetricValuesByLanguage = {}
+
+	for (const { data } of hierarchy(accumulatedData.unifiedMapNode)) {
+		const fileExtension = getFileExtension(data.name)
+		if (data.type === NodeType.FILE && fileExtension !== undefined && !isPathBlacklisted(data.path, blacklist, BlacklistType.exclude)) {
+			const filesPerLanguage = numberOfFilesByLanguage.get(fileExtension) ?? 0
+			numberOfFilesByLanguage.set(fileExtension, filesPerLanguage + 1)
+
+			setMetricValues(data, metricValues)
```
ce-bo, line 62: `setMetricValues` accumulates the metric values of a node. The accumulation does not differentiate the programming language. In the next line the `metricValues` are indexed by fileExtension but `metricValues` does include values of other languages too, isn't it?

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/selectors/artificialIntelligence.selector.ts
```diff
@@ -0,0 +1,103 @@
...
+			metricValuesByLanguage[fileExtension] = metricValues
+
+			if (isFileValid(data, fileExtension)) {
+				totalRloc += data.attributes[AREA_METRIC]
+				calculateRiskProfile(data, rlocRisk, fileExtension)
```
ce-bo, line 67: The function is more like an `update` or `aggregation` of the risk profile.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/selectors/artificialIntelligence.selector.ts
```diff
@@ -0,0 +1,103 @@
...
+			}
+		}
+
+		if (totalRloc > 0) {
+			artificialIntelligenceViewModel.riskProfile = getPercentagesOfRiskProfile(rlocRisk)
```
ce-bo, line 72: Is `getPercentagesOfRiskProfile` used elsewhere? If not, lets do not export it and call it inside calculateRiskProfile().

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/riskProfileHelper.ts
```diff
@@ -0,0 +1,47 @@
+import { CodeMapNode } from "../../../codeCharta.model"
```
ce-bo, line 1: Nice, this class is much more smaller and cleaner.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/selectors/artificialIntelligence.selector.ts
Hall-Ma: Thanks for finding this bug. I will fix it.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/selectors/artificialIntelligence.selector.ts
Hall-Ma: We spoke about it. We leave it like that.

visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/suspiciousMetricsHelper.ts
```diff
@@ -41,18 +41,13 @@ export function calculateSuspiciousMetrics(metricAssessmentResults: MetricAssess
...
 	for (const [metricName, value] of Object.entries(node.attributes)) {
 		if (value === 0) {
 			continue
 		}
 
-		if (metricValuesByLanguage[fileExtension][metricName] === undefined) {
-			metricValuesByLanguage[fileExtension][metricName] = []
-		}
+		metricValuesByLanguage[fileExtension] ??= {}
+		metricValuesByLanguage[fileExtension][metricName] ??= []
```
ce-bo, line 50: Sophisticated :-)

ce-bo (approved):

Issue #2731, opened 2022-03-20
attribute type not correctly set in attribute side bar


shaman-apprentice:
# Bug

## Expected Behavior

Given a map with edge metrics, when opening the attribute side bar, I expect the attribute type of the edge metric to be selected correctly

## Actual Behavior

The initial active value is always sum

## Screenshots
![image](https://user-images.githubusercontent.com/3596742/159163861-75378723-5e40-4b41-9baf-ff404fb0a7bf.png)

## Steps to Reproduce the Problem
1. Open attribute side bar by clicking on a folder
2. set edge metric to median
3. close attribute side bar
4. open attribute side bar again

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.93

PR #2732, opened 2022-03-20
Refactor/2318/migrate metric value hovered component

shaman-apprentice:
- migrate metricType.component
- store hovered node id in store instead of hovered node path, as hovered node path can be easily derived by id but not vice versa
- while refactoring a little I noticed a bug and created #2731. When fixing it, I will refactor a little more

ref #2318


shaman-apprentice:
The 2 code smells are the typical "redux pattern smells" and actual fine

Christian-Eberhard (approved):
Should we keep empty files like "visualization/app/codeCharta/ui/metricType/metricType.component.scss" ?

shaman-apprentice:
> Should we keep empty files like "visualization/app/codeCharta/ui/metricType/metricType.component.scss" ?

We shouldn't. And I have removed it, haven't I?

Christian-Eberhard:
I found one file thats new with empty content, -> visualization/app/codeCharta/ui/metricType/metricType.component.scss

Issue #2341, opened 2021-08-05
Improve parser usability


BridgeAR:
# Feature request

## Description

As a user, I want to be asked for what parser I want to execute and for necessary and optional input by our parsers so that I do not have to remember all of the different options.

## Acceptance criteria
- Keep the current behavior that allows to execute a single parser.
- Keep the current behavior that allows to pass through different options to our parsers
- If no parser is specified, the user is going to be asked for parser to be executed.
- If no arguments are passed to the parser (running `ccsh scmlogparserv2`), the user is going to be asked about the necessary input, if any is required (some parsers have might have no options that are required, e.g., `scmlogparserv2`).
- Default arguments are used for almost all situations where they seem useful (e.g., `silent` and using a format that contains the most detailed output).
- The flow would look for example like this for the scmlogparserv:

```
> ccsh
Which parser do you want to execute?
1) SonarImporter - Import Metrics from Sonar Web API
2) SourceCodeParser - Parse Java Source Code and generate some standard metrics
3) ScmLogParser - Generate metrics from SVN/Git Logs
> executing scmlogparser
Do you want to add authors to the output file?
> yes/no
The `scmgitlog.cc.json` file is being created. This might take a bit.
# Show the ongoing steps.
Creating git log file.
Parsing git log.
scmgitlog.cc.json created.
```

## Development notes (optional Task Breakdown)
- [x] Start with a single parser
- [x] Check if the implementation is possible to be abstracted for all parsers
- [ ] Implement the behavior for all other parsers
- [x] If some parsers are more complicated than others, focus on the simple ones and open a PR for each parser

## @IhsenBouallegue Edit: Interactive dialogs to implement
- [x] check
- [x] merge
- [x] edgefilter
- [x] modify
- [x] csvimport
- [x] sonarimport
- [x] gitlogparser
- [x] svnlogparser
- [x] sourcemonitorimport @luelista 
- [x] csvexport
- [x] sourcecodeparser  @luelista 
- [x] codemaatimport @Chrisp485 
- [x] tokeiimporter  @luelista 
- [x] rawtextparser @IhsenBouallegue  @Chrisp485

PR #2737, opened 2022-03-21
Feature/2341/improve parser usability

IhsenBouallegue:
# Improve parser usability

Close: #2341

## Description


analysis/tools/InteractiveCLI/src/main/kotlin/de/maibornwolff/codecharta/tools/interactivecli/ParserDialogInterface.kt
```diff
@@ -0,0 +1,18 @@
+package de.maibornwolff.codecharta.tools.interactivecli
+
+import picocli.CommandLine
+
+interface ParserDialogInterface {
+    fun generateParserArgs(args: Array<String>, commandLine: CommandLine): Array<String>
```
NearW, line 6: I'd recommend to use `List` instead of `Array`

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
```diff
@@ -70,7 +71,25 @@ class Ccsh : Callable<Void?> {
...
+            if (args.isEmpty() && !isParserUnknown(args, commandLine)) {
+                val selectedParser = ParserService.selectParser(commandLine)
+                println("Executing $selectedParser")
+                ParserService.executeSelectedParser(selectedParser)
+            }
+        }
+
+        private fun isParserUnknown(args: Array<String>, commandLine: CommandLine): Boolean {
+            if (args.isNotEmpty()) {
+                val firstArg = args[0]
```
NearW, line 88:
```suggestion
                val firstArg = args.first()
```

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/parser/ParserService.kt
```diff
@@ -0,0 +1,60 @@
...
+            val subCommands = commandLine.subcommands.values
+            val parsersList = mutableListOf<String>()
+
+            for (subCommand in subCommands) {
+                val parserName: String = subCommand.commandName
+                val parserDescriptions = subCommand.commandSpec.usageMessage().description()
+                val parserDescription = parserDescriptions[0]
+                parsersList.add("$parserName - $parserDescription")
+            }
+            return parsersList
```
NearW, line 53:
```suggestion
            return subCommands.map {
                val parserName = subCommand.commandName
                val parserDescriptions = subCommand.commandSpec.usageMessage().description()
                val parserDescription = parserDescriptions[0]
                "$parserName - $parserDescription"           
            }
```

analysis/tools/InteractiveCLI/src/main/kotlin/de/maibornwolff/codecharta/tools/interactivecli/ParserDialogInterface.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    fun isValidFileName(fileName: String): Boolean
+
+    fun getOutputFileName(fullFileName: String): String {
+        return fullFileName.substringAfterLast("/").substringAfterLast("\\").substringBefore(".")
+    }
+
+    fun checkExtension(fileName: String, expectedFileExtension: String): Boolean {
```
NearW, line 14: rename to `hasMatchingExtension` or `hasExpectedExtension`. This implies that the function returns something AND that this something is a boolean. `checkExtension` sounds like a function, that does not return anything and throws errors in case, something does not match.

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
```diff
@@ -116,16 +119,16 @@ class EdgeProjectBuilder(private val project: Project, private val pathSeparator
...
+            filteredEdges: List<Edge>
     ): MutableMap<String, Any> {
         val aggregatedAttributes: MutableMap<String, Any> = mutableMapOf()
 
         listOfAttributes.forEach { key: String ->
             val attributeType = getAttributeTypeByKey(key)
             val filteredAttribute = filteredEdges.filter { edge: Edge -> edge.attributes.containsKey(key) }
             var aggregatedAttributeValue =
-                filteredAttribute.sumBy { edge: Edge -> edge.attributes[key].toString().toFloat().toInt() }
+                    filteredAttribute.sumBy { edge: Edge -> edge.attributes[key].toString().toFloat().toInt() }
```
NearW, line 131: What on earth is this conversion? :D I know this is old code, but we should work on that.
Is that not working?
```suggestion
                    filteredAttribute.sumBy { edge: Edge -> edge.attributes[key].toInt() }
```

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
```diff
@@ -0,0 +1,38 @@
...
+            val defaultOutputFileName = getOutputFileName(inputFileName)
+            val outputFileName: String = KInquirer.promptInput(
+                message = "What is the name of the output file?",
+                hint = defaultOutputFileName,
+                default = defaultOutputFileName
+            )
+
+            val defaultPathSeparator = "/"
+            val pathSeparator: String = KInquirer.promptInput(
+                message = "What is the path separator?",
```
NearW, line 27: Can't we detect this through the operating system?

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class ParserDialog {
+    companion object : ParserDialogInterface {
+        private const val EXTENSION = "cc.json"
+        override fun generateParserArgs(args: Array<String>, commandLine: CommandLine): Array<String> {
+
+            var inputFileName = KInquirer.promptInput(message = "What is the $EXTENSION file that has to be parsed?")
+            if (inputFileName.substringAfter(".") != EXTENSION)
+                inputFileName += ".$EXTENSION"
+            println("File path: $inputFileName")
```
NearW, line 16: Not sure, if we should use println when using picocli. If I remember correctly, we use a logger to log to fixed logstreams like info or error.

analysis/tools/InteractiveCLI/src/main/kotlin/de/maibornwolff/codecharta/tools/interactivecli/ParserDialogInterface.kt
IhsenBouallegue: Fixed in the next commits.
But is there a specific reason? Because I would still have to cast the result back to an array later.

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
IhsenBouallegue: Fixed in the next commits

analysis/tools/InteractiveCLI/src/main/kotlin/de/maibornwolff/codecharta/tools/interactivecli/ParserDialogInterface.kt
IhsenBouallegue: I agree! Fixed in the next commits

analysis/tools/InteractiveCLI/src/main/kotlin/de/maibornwolff/codecharta/tools/interactivecli/ParserDialogInterface.kt
NearW: https://stackoverflow.com/a/36263748

The main function is derived from java and uses an Array as input. We can't change that. However, we can decide what we use in our code and we should prefer lists there (see link above).
In that case it's fine to simply convert from array to list and list to array when dealing with the main function.

IhsenBouallegue: Aha got you! thanks for the explanation 👍

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
IhsenBouallegue: Thank you for the tip. Indeed I found Kotlin Logger being used, and I changed the println to logger.info.

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/parser/ParserService.kt
```diff
@@ -0,0 +1,60 @@
...
+            val selectedParser: String = KInquirer.promptList(
+                message = "Which parser do you want to execute?",
+                choices = getParserNamesWithDescription(commandLine)
+            )
+
+            return extractParserName(selectedParser)
+        }
+
+        fun executeSelectedParser(selectedParser: String) {
+            val args = emptyArray<String>()
```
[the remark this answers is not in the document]
IhsenBouallegue: Yes, it is being used on line 26. The parsers main method needs to be called with an empty array, so instead of passing an empty array to each, we declare it once. For now, there is only the edgefilter interactive dialog implemented. But the goal is to implement it in all parser and get rid of the "print("Not implemented yet")".

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
IhsenBouallegue: Thanks for the note @NearW 😄
I finally got around to investigating that. Here is what I discovered:
- *toString()* is needed because `edge.attributes[key]` is of type `any`. So `toFloat()` and `toInt()` can't cast it directly.
- *toFloat()* is needed because the result at that point would be "90.0" and to `toInt()` throws an error when trying to cast it. I assume because of the ".".
So casting to a string avoids the problem with `any` and casting to a float avoids the problem with `.`.

However, I fixed the deprecation problem because of `sumBy` and converted to `sumOf`. So that's an improvement at least 👍

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
```diff
@@ -60,9 +59,13 @@ class EdgeProjectBuilder(private val project: Project, private val pathSeparator
     private fun insertEdgeAsNode(nodeEdgeName: String) {
         val nodeFilename = nodeEdgeName.split(pathSeparator).reversed().first()
         val nodePath = nodeEdgeName.split(pathSeparator)
-        val nodeParentPath = nodePath.subList(2, max(2, nodePath.size - 1))
-        val node = Node(nodeFilename, NodeType.File)
-        insertNodeInProjectBuilder(node, nodeParentPath)
+        if (nodePath.size <= 1) {
+            println("No parent node found")
```
ce-bo, line 63: Im not sure if we should display this message, because it is printed at the command line without any information which edge is affected. So maybe the specified message is a bit confusing, isn't it? What do you think?

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
```diff
@@ -116,16 +119,16 @@ class EdgeProjectBuilder(private val project: Project, private val pathSeparator
     }
 
     private fun getAggregatedAttributes(
-        listOfAttributes: MutableList<String>,
-        filteredEdges: List<Edge>
+            listOfAttributes: MutableList<String>,
+            filteredEdges: List<Edge>
```
ce-bo, line 123:
```suggestion
        listOfAttributes: MutableList<String>,
        filteredEdges: List<Edge>
```

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
```diff
@@ -0,0 +1,37 @@
...
+
+private val logger = KotlinLogging.logger {}
+
+class ParserDialog {
+    companion object : ParserDialogInterface {
+        private const val EXTENSION = "cc.json"
+        override fun collectParserArgs(): List<String> {
+
+            var inputFileName = KInquirer.promptInput(message = "What is the $EXTENSION file that has to be parsed?")
+            if (inputFileName.substringAfter(".") != EXTENSION)
```
ce-bo, line 16: I think this will not work for filenames that have multiple dots included, for instance: my.file.cc.json
I suggest to use substringAfterLast instead.

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/parser/ParserService.kt
```diff
@@ -0,0 +1,57 @@
...
+
+            when (selectedParser) {
+                "check" -> print("Not implemented yet")
```
ce-bo, line 23: Maybe we should add a new private helper function in this ParserService that just prints the "Not implemented yet" message => private fun printNotSupported(parserName: string) { ... }

And I would prefer a message like "The interactive usage of parser <Foo> is not supported yet. Please specify the full command to run the parser."

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
IhsenBouallegue: After last will pick up only the json and not cc.json. Maybe we do a regex match and make sure cc.json is at the end?

analysis/tools/ccsh/src/test/kotlin/de/maibornwolff/codecharta/ccsh/parser/ParserServiceTest.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        Assertions.assertThat(outContent.toString()).contains("Not implemented yet")
+    }
+
+    @Test
+    fun `should execute sourcemonitorimport parser`() {
+        ParserService.executeSelectedParser("sourcemonitorimport")
+
+        Assertions.assertThat(outContent.toString()).contains("Not implemented yet")
+    }
```
ce-bo, line 94: Idea for future refactoring of the tests: Use a DataProvider to test the execution of all parsers by only writing one test.
Described here: https://www.baeldung.com/parameterized-tests-junit-5#6-method

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
IhsenBouallegue: I am not sure either as I didn't add this change. And honestly I don't know why it's there. It has nothing to do with the CLI.

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
ce-bo: Ah ok. I think you can do substringAfterLast(".cc.json")

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
BridgeAR: This is something @Hall-Ma and @polinasch ran into during the initial development in tests. If I recall correct it's about a faulty cc structure / path unrelated to each other. In that case, the bug would be hidden. It should probably already throw an error before reaching this point by doing some sanity checks. To at least let the user know something is wrong, this has been added. The message itself should be improved and we might even throw an error here (I am not certain how the tool is used, it might be caught upper layers).

analysis/build.gradle
```diff
@@ -80,7 +81,7 @@ subprojects {
         implementation group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-jdk8', version: kotlin_version
         implementation group: 'org.jetbrains.kotlin', name: 'kotlin-reflect', version: kotlin_version
         implementation group: 'org.jetbrains.kotlinx', name: 'kotlinx-coroutines-core', version: kotlinx_coroutines_version
-
+        implementation 'com.github.kotlin-inquirer:kotlin-inquirer:0.1.0'
```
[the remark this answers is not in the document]
IhsenBouallegue: You are right. I fixed it and moved it to the packages where it was used instead of all of analysis.

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
IhsenBouallegue: Fixed in the next commits 👍

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/parser/ParserService.kt
IhsenBouallegue: I like this approach 👍  I implemented it in this way and slightly adjusted the tests.

analysis/tools/ccsh/src/test/kotlin/de/maibornwolff/codecharta/ccsh/parser/ParserServiceTest.kt
IhsenBouallegue: Awesome! The "not implemented yet" tests are there temporarily, but as soon as we implement more interactive dialogs we should look into parameterized tests 😄

analysis/tools/ccsh/build.gradle
```diff
@@ -55,11 +55,14 @@ dependencies {
...
 
     implementation group: 'info.picocli', name: 'picocli', version: picocli_version
+    implementation group: 'io.github.microutils', name: 'kotlin-logging', version: kotlin_logging_version
 
     testImplementation group: 'junit', name: 'junit', version: junit4_version
     testImplementation group: 'org.assertj', name: 'assertj-core', version: assertj_version
+    testImplementation group: 'io.mockk', name: 'mockk', version: mockk_version
 
     testImplementation("org.junit.platform:junit-platform-runner:$junit_platform_version")
+    implementation "org.jetbrains.kotlin:kotlin-script-runtime:1.6.10"
```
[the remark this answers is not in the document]
IhsenBouallegue: I am not sure why this dependency was added. When I removed it, everything worked as intended and no errors showed up anywhere. So I took it out completely.

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
IhsenBouallegue: Maybe 🤔 I went with the regex approach as it was clearer to me. Now it will make sure .cc.json is the extension of the file, unrelated to how many `.` the file has 👍

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
ce-bo: I removed the println and log an error instead. Furthermore, I improved the error message a bit.

ce-bo (commented):
LGTM! I had some issues when executing the ccsh. I get the error that the ccsh Main class could not be found, so I could not test it locally. Can someone else confirm this?

NearW:
@ce-bo I'm pretty sure, that this can be tested automatically as well. There are no happy-path e2e-tests right now, but it might make sense to add one for each parser. We'd only check for output on stderr and in that case fail the pipeline.

However, this requires a refactoring, since we mostly use println. A kotlin logger should be used instead. You could also document some logging-guidelines and implement best-practices here.

TL;DR
- 2 issues
- add kotlin logging and use dedicated log levels
- add automatic happy-path e2e test for each parser

IhsenBouallegue:
> @ce-bo I'm pretty sure, that this can be tested automatically as well. There are no happy-path e2e-tests right now, but it might make sense to add one for each parser. We'd only check for output on stderr and in that case fail the pipeline.
> 
> However, this requires a refactoring, since we mostly use println. A kotlin logger should be used instead. You could also document some logging-guidelines and implement best-practices here.
> 
> TL;DR
> 
>     * 2 issues
> 
>     * add kotlin logging and use dedicated log levels
> 
>     * add automatic happy-path e2e test for each parser

I totally agree. Proper logging should be used everywhere. When going over the parsers to implement the interactive dialogs, I will try to refactor as much as possible 😄

IhsenBouallegue:
> LGTM! I had some issues when executing the ccsh. I get the error that the ccsh Main class could not be found, so I could not test it locally. Can someone else confirm this?

After building, you will find the ccsh script under bin in the build folder. In my testing, I navigated to it and executed it simply with `./ccsh`.

analysis/build.gradle
```diff
@@ -32,7 +33,7 @@ buildscript {
 
         mockito_core_version = '3.12.4'
         mockito_kotlin_version = '2.2.0'
-        mokk_version = '1.12.3'
+        mockk_version = '1.12.3'
```
shaman-apprentice, line 36: Is that on purpose?

analysis/build.gradle
```diff
@@ -68,7 +70,7 @@ allprojects {
     apply plugin: "org.jlleitschuh.gradle.ktlint"
 
     jacoco {
-        toolVersion = jacoco_version
+      toolVersion = jacoco_version
```
shaman-apprentice, line 73: Why have you changed the spaces here?

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
```diff
@@ -9,15 +9,14 @@ import de.maibornwolff.codecharta.model.Path
...
-import kotlin.math.max
 
 class EdgeProjectBuilder(private val project: Project, private val pathSeparator: Char) {
     private val logger = KotlinLogging.logger {}
 
     private val projectBuilder = ProjectBuilder(
-        listOf(MutableNode("root", NodeType.Folder)),
-        mutableListOf(),
-        getAttributeTypes()
+            listOf(MutableNode("root", NodeType.Folder)),
```
shaman-apprentice, line 17: This seems like one identation to much?

shaman-apprentice (approved):
early approve. See comments

analysis/build.gradle
IhsenBouallegue: Yes, it is on purpose. It's for the Mock Kotlin dependency https://mockk.io/

analysis/build.gradle
IhsenBouallegue: That's weird. Will fix it in the next commit

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeProjectBuilder.kt
IhsenBouallegue: It seems like the Intellij formatter is a bit confused. Fixed in the next commit 👍

Issue #2738, opened 2022-03-21
Add more variants of the color gradient


Christian-Eberhard:
# Feature request

## Description

As a user, I want to choose a gradient which clearly shows which files are close to the set color ranges.

## Acceptance criteria
- Implementation of a color gradient that only applies its color to the buildings in between the set range

ce-bo:
Closed by #2750

PR #2750, opened 2022-03-25
Adds focused gradient

Christian-Eberhard:
# Adds the focused gradient option

Issue: #2738

## Description

Adds a gradient that only fades between the "from" and "to" range options. Keeps values below the from or above the to value at its respective color.

## Screenshots or gifs

grafik


visualization/app/codeCharta/util/algorithm/treeMapLayout/treeMapHelper.ts
```diff
@@ -243,6 +247,33 @@ function calculateTrueGradient(mapColors: MapColors, colorRange: ColorRange, met
 	return ColorConverter.convertColorToHex(new Color().lerpColors(neutralColorRGB, negativeColorRGB, negativeFactor))
 }
 
+function calculateFocusedGradient(mapColors: MapColors, colorRange: ColorRange, metricValue: number) {
```
ce-bo, line 250: We could extract the 3 or 4 calculation methods to separate calculation classes. This would reduce the cyclomatic complexity of treeMapHelper.ts.

ce-bo (approved):

ce-bo:
LGTM. We should think about displaying the descriptions of the single gradient options like we do for metrics.

ce-bo (approved):

Issue #2611, opened 2022-01-17
Floor labels drawn next to the map


BridgeAR:
# Bug

## Expected Behavior

GIVEN a map in multiple mode

WHEN using the map as always

THEN the map should contain the labels on the floor of the folders.

## Actual Behavior

The labels are drawn next to the map.

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/149800414-bd599740-294d-4966-944d-549ae03f3d3d.png)

## Steps to Reproduce the Problem

- Change margin
- Scale height of buildings
- some other (unknown) actions

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.87.1
- OS: Unix
- Browser: Chrome

PR #2763, opened 2022-03-30
Fixes floor labels being drawn next to map

Christian-Eberhard:
# Fixes floor labels being drawn next to map
Issue: #2611 

## Description

Fixes the issue of the floor labels being drawn next to map, after changing the margin when the height was not equal to 1.

### DIFF
--- a/visualization/app/codeCharta/ui/codeMap/threeViewer/floorLabels/floorLabelDrawer.ts
+++ b/visualization/app/codeCharta/ui/codeMap/threeViewer/floorLabels/floorLabelDrawer.ts
@@ -1,7 +1,7 @@
 "use strict"
 
 import { Node } from "../../../../codeCharta.model"
-import { CanvasTexture, BackSide, Mesh, MeshBasicMaterial, PlaneGeometry, RepeatWrapping, Vector3 } from "three"
+import { BackSide, CanvasTexture, Mesh, MeshBasicMaterial, PlaneGeometry, RepeatWrapping, Vector3 } from "three"
 import { FloorLabelHelper } from "./floorLabelHelper"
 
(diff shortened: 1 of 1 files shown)


visualization/app/codeCharta/ui/codeMap/threeViewer/floorLabels/floorLabelDrawer.ts
```diff
@@ -111,32 +111,37 @@ export class FloorLabelDrawer {
 		planeMesh.rotateX((90 * Math.PI) / 180)
 
 		// Position plane over the map
-		const liftToPreventZFighting = 10
-		plane.translate(scaledMapWidth / 2, scaledMapHeight / 2, -2.01 * (floorLevel + 1) - liftToPreventZFighting)
+		const liftToPreventZFighting = 2
```
ce-bo, line 114: The lift is not ideal, but what is the reason why you changed this?
The labels are disappearing more quickly now.

ce-bo (commented):
It works! Would you check my comment before approval, please?

visualization/app/codeCharta/ui/codeMap/threeViewer/floorLabels/floorLabelDrawer.ts
Christian-Eberhard: Only when changing the height though, which is proven to be a bug as mentioned in [#2763 2764](https://github.com/MaibornWolff/codecharta/issues/2764). Changing the lift to 2 is just a visual change, that is apparent when having maps with high scales and floor labels on the correct position. The floating with the value set to 10 is very apparent

ce-bo (approved):

Issue #2764, opened 2022-03-30
Floor labels inside of folder geometry when changing height scaling


Christian-Eberhard:
# Bug

## Expected Behavior

GIVEN a map with nodes that are folders

WHEN changing the height scaling

THEN the floor labels should be visible

## Actual Behavior

The floor labels are inside of the folder geometry

## Screenshots

![missingLabels.JPG](https://images.zenhubusercontent.com/6076b0525fa0ae9470963a22/e69ecd8d-da8c-45b9-8b33-d66da5b2ac41)

## Steps to Reproduce the Problem
1. Change height scale to something high

## Workaround after [#2763](https://github.com/MaibornWolff/codecharta/pull/2763)
Change the margin

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.95.0
- OS: Windows
- Browser: Firefox, Chrome

Christian-Eberhard:
The problem is as follows:

When changing the height of the map in the height panel, only the scale of the whole map is getting changed but the whole map won't be redrawn, for performance reasons. The floor labels stay on the same coordinates but the folder geometry is beeing scaled accordingly, resulting in the floor labels beeing "invisible" since they're now stuck inside the folder geometry.

The floor labels can't be scaled in the y axis, since they're planes.

PR #2766, opened 2022-03-30
Fixes floor labels being stuck inside folder geometry when scaling

Christian-Eberhard:
# Fixes floor labels being stuck inside folder geometry when scaling

Closes: #2764

## Description

Moves the canvases according to the folder level and changed scale


ce-bo (approved):
LGTM. I tested it on my machine. Would you please fix the tests?


I found another edge case: Set Height Scaling to minimum and Margins to Maximum.
Then, some floor labels were lost. Would you have a quick look on that, please? Maybe we should fix that in a follow up issue.

Issue #2671, opened 2022-02-18
Display Project Name if not empty or blank


Richargh:
# Feature request

## Description

When I load cc.json files via url (https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=https:%2F%2Fraw.githubusercontent.com%2FMaibornWolff%2Fcodecharta%2Fmain%2Fgh-pages%2Fassets%2Fccjson%2Fshowcase%2Fjunit4%2Fjunit4_2016-10-28.cc.json), then the file name is no longer properly readable:

<img width="499" alt="grafik" src="https://user-images.githubusercontent.com/28268662/154696003-bd2ec265-34ab-45a1-96e0-ff7b4ff1aa56.png">
<img width="343" alt="grafik" src="https://user-images.githubusercontent.com/28268662/154696029-ea9844c3-5a58-4795-a0a4-ed6ea2e099f3.png">

Can CodeCharta instead display the "projectName" if it's not empty or blank?

## Acceptance criteria
- CodeCharta displays the "projectName" if it's not empty or blank? 
- if "projectName" is empty or blank and the URL is very long, then CodeCharta shows the last characters of the URL, not the first.

PR #2767, opened 2022-03-31
Feature/2671/display project name if not empty or blank

cschild:
# Display project name if not empty or blank

Close: #2671

## Description

- Displays the file projectName if it's not empty or blank
- Otherwise, displays the last part of the URL


visualization/app/codeCharta/util/urlExtractor.ts
```diff
@@ -46,8 +46,22 @@ export class UrlExtractor {
...
+	getFileName(oldFileName: string, projectName: string) {
+		let fileName: string
+
+		if (projectName?.trim()) {
+			fileName = projectName
+		} else {
+			const fileNameParts = oldFileName.split("/")
+			fileName = fileNameParts[fileNameParts.length - 1]
+		}
+		return fileName
```
ce-bo, line 65: I suggest to return early (in line 60) and omit the else block. This can then be optimized like so:
```suggestion
		return projectName?.trim().length >0 ? projectName : oldFileName.split("/").pop();
```

cschild: That's nice, thanks! I didn't think of the pop method

BridgeAR:
```js
return projectName?.trim() || oldFileName.split("/").pop();
```

ce-bo (approved):

ce-bo (approved):
Let's add a changelog entry.

Issue #2762, opened 2022-03-30
Using custom view of the sample map after upload is not possible in the demo


Hall-Ma:
# Bug

## Expected Behavior

<GIVEN>
Sample maps (codecharta.cc.json, codecharta_analysis.cc.json)

When custom views are created, saved, and downloaded from the preselected sample file, they should be clickable again after uploading.

## Actual Behavior

When custom views are created, saved, and downloaded from the preselected sample file, they are not clickable again after uploading.
### Note: 
This only happens when you use the preselected map. (After each reselection of the files, they will be set to multiple mode.)
For more information see https://github.com/MaibornWolff/codecharta/issues/2761#issuecomment-1082766278

## Steps to Reproduce the Problem
1. Save, and download a new custom view from our sample maps (chose the map that is already selected and don't select it again)
2. Delete custom views in the demo
4. Upload saved custom view (it won't be clickable now)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.95.0
- OS: Windows
- Browser: Chrome

PR #2769, opened 2022-04-01
Fix/2762/Set loaded files via url to multiple mode by default

Hall-Ma:
# Set the loaded files via URL to the multiple mode

closes: #2762

## Description

- loaded files via URL will now be set to multiple mode by default or delta mode (if corresponding query parameter is given)


visualization/app/codeCharta/e2e/url.e2e.ts
```diff
@@ -56,7 +56,6 @@ describe("codecharta", () => {
 	it("should load data when file parameters in url are valid", async () => {
 		await mockResponses()
 		await goto(`${CC_URL}?file=fileOne.json&file=fileTwo.json`)
-		await checkSelectedFileName("Sample Project with Edges")
```
ce-bo, line 59: Ideally there would be a test to check the behavior if the mode parameter is present and another one if not.

ce-bo (approved):

Issue #2734, opened 2022-03-20
Migrate Custom Views to Angular


Hall-Ma:
# Feature request

## Description

To continue the improvement of Custom Views (see #2489) , the entire component should be migrated to Angular.

## The component should be migrated into new smaler ones  (suggestions):
- [x] Add button
- [x] Download button
- [x] Upload button 
- [x] List & Sub list View

## Note
When starting the migration of a component, please mentioned this also in the migration issue #2318 (component + your name)

PR #2771, opened 2022-04-04
Refactor/2734/migrate and extract upload custom configs button

### LINKED ISSUE #2761: What requirements exist for uploading downloaded custom configs
# Question

I am not sure if this is a bug or designed behavior.

**As a user I want to know why an upload failed.**

# Observed behavior:

1. Save a custom config and modify the file name ending when save dialog pop ups
2. Upload the same custom config
3. It fails due to invalid or missing API version

image

image

## Question

I would expect that the third point would be a successfull upload. So my question is:
- Is this a bug?
- Is it designed behavior? Can we improve the error message in that case, that it becomes clear, that the file name must match a specific file format?

shaman-apprentice:
ref #2734 
ref #2318
close #2761


shaman-apprentice:
superceded by #2773

PR #2784, opened 2022-04-19
refactor: introduce custom angular material theme

shaman-apprentice:
introduce custom theme, so that we don't have to overwrite our main green color everywhere

ref #2318


Hall-Ma (approved):
LGTM :)

Issue #2329, opened 2021-07-30
Limit maximum legend width


BridgeAR:
# Bug

## Expected Behavior

GIVEN a map with nodes that have long folder / node names

WHEN opening the legend panel

THEN the legend panel should have a maximum width of ~15-20% of the view and a minimum size so that "Outgoing Edge" is always fully visible. 

## Actual Behavior

The legend panel may increase it's size so that a lot of the screen is not readable anymore.

## Screenshots

![image](https://user-images.githubusercontent.com/8822573/127677894-f3e25218-5886-4dfc-9b35-7b273e248670.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.77.0

luelista:
What are we going to do with the long labels? 

Cut them off?
<img width="541" alt="grafik" src="https://user-images.githubusercontent.com/388142/168556351-0af3e698-eade-4fdb-a6bf-a9e7e5611ff6.png">

Or word wrap?
<img width="544" alt="grafik" src="https://user-images.githubusercontent.com/388142/168556476-bb1ec1d3-3289-4955-a72a-6ca940c8ac9d.png">

I think first looks better, but second is more usable...

BridgeAR:
I agree that the first looks better and I also tend to use that with a tooltip showing the full label with one difference: cut off the start of the label, not the end. The end is often the most important part.

luelista:
direction: rtl; is only recommended to use on text which is actually in a RTL language. See https://developer.mozilla.org/en-US/docs/Web/CSS/direction#examples

If I'd set our file paths to RTL, it's displayed incorrectly: the first forward slash is moved to the end (same behaviour with the full stop in the example above).

luelista:
<img width="596" alt="grafik" src="https://user-images.githubusercontent.com/388142/168795500-41690729-f9a0-43c3-aef7-14bfb42ee7bc.png">

luelista:
Okay, I found a kind-of hacky workaround for that, not sure if we should really do this, but it seems to work...

https://github.com/MaibornWolff/codecharta/pull/2804/files#diff-50667783b76530e0fe16fe24035dfe4c6c004b0e5aea0c190c3e53becdb9d845R86-R93

BridgeAR:
Interesting!

[...]

PR #2804, opened 2022-05-16
fix: limit length of labels in legend panel (#2329)

luelista:
# fix: limit length of labels in legend panel

Issue: #2329

## Description

Cut off overlong labels in the legend panel, so it doesn't take the width of the whole screen. The full label is visible as a tooltip.

## Screenshots or gifs

grafik


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Update slider of area metric options correctly on changes of related input field [#2787](https://github.com/MaibornWolff/codecharta/pull/2787)
+-   Limit length of labels in legend panel #2329
```
shaman-apprentice, line 13: This GitHub linking doesn't work with the changes information we display to the user within CodeCharta, which are generated by this changelog. Although we might add this parsing to the generation script.

We link the PR and not the issue. Feel very free to address that. I find it more natural to link the issue as well :)

```suggestion
-   Limit length of labels in legend panel [#2804](https://github.com/MaibornWolff/codecharta/pull/2804)
```

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
```diff
@@ -70,19 +72,25 @@ cc-legend-panel {
...
 		.bold {
 			font-weight: 600;
 		}
+
+		.cc-labelled-color-picker-container > .cc-labelled-color-picker-label {
+			white-space: nowrap;
+			text-overflow: ellipsis;
+			overflow: hidden;
+
+			/* This tricks the browser into cutting off the beginning instead of the end of the label. However,
```
shaman-apprentice, line 86: I am not too sure about the comments. They describe standard css. Maybe rename the css selector to `cc-cut-text-overflow-at-beginning` instead? On the other side there is no harm by the comments. I am fine with leaving it as it is as well.

shaman-apprentice (approved):
Early approve. Just the changelog entry and the optional comment about the comment (commentception :P).

Random information: I like to use [GitHub's closing words](https://docs.devart.com/studio-for-sql-server/source-controlling-databases/associating-commits-with-github-issues.html) in PR description.

visualization/app/codeCharta/ui/legendPanel/legendPanel.component.scss
luelista: IMO this is still a misuse, or at least exploiting an edge case of standard CSS. The `direction: rtl;` property is not intended to switch text-overflow cutoff, but for right-to-left languages, which this is not. So as this a hack with some weird side effects, I want to make sure there is some warning at least.

PR #2810, opened 2022-05-18
Implement gitlogparser Interactive CLI

IhsenBouallegue:
# Implement gitlogparser Interactive CLI

Issue: #2341

## Description

- Add the interactive dialog for the gitlogparser

## Screenshots or gifs
grafik


analysis/import/GitLogParser/src/main/kotlin/de/maibornwolff/codecharta/importer/gitlogparser/GitLogParser.kt
```diff
@@ -52,6 +54,7 @@ class GitLogParser(
     @CommandLine.Option(names = ["-o", "--output-file"], description = ["output File (or empty for stdout)"])
     private var outputFile = ""
 
+    // confusing name: it should be --compress because when this flag is set to true it will compress the file
```
BridgeAR, line 57: I totally agree. We used to have `--compressed` and it was changed. The goal was to set the default behavior to compressed output instead of uncompressed output. I do not remember why the flag was changed to achieve that. The old name would however we better.

luelista: this is really confusing, due to some tricky behavior of the cli parser. if used as documented, this flag works correctly: not passing `--non-compressed` causes the file to be compressed, passing `--non-compressed` causes it to be uncompressed. However, if you specify `--non-compressed=<bool>`, it get's wrong:

| Argument | File compressed |
| - | - |
| none | yes |
| `--non-compressed` | no |
| `--non-compressed=true` | yes |
| `--non-compressed=false` | no |

luelista: As a quick fix I would specify `arity = "0"` in the Option annotation, so the confusing `--non-compressed=<bool>` version is rejected.

luelista (approved):

PR #2815, opened 2022-05-22
Refactor/2318/migrate global settings button with dialog

shaman-apprentice:
migrate:
- globalSettingsButton.component
- dialog.globalSettings.component
- layoutSelection.component
- sharpnessModeSelector.component
- maxTreeMapFiles.component
- maxTreeMapFiles.service
- hideFlatBuildings.service
- resetCameraIfNewFileIsLoaded.service


shaman-apprentice:
> LGTM! I only have two aspects that should change if you feel the same way. What do you think about to remove the first grey line in these fields? (I think I darkly remember that it's not that easy to delete that, is it?) image Secondly, what do you think about putting both line on the same level? image

I thought about both and find both reasonable as well. But I like it this way more, so I will leave it for now.

PR #2820, opened 2022-05-24
Fix/2580/make sure all command line tools understand gz files

### LINKED ISSUE #2580: Make sure all command line tools understand .gz files
# Bug

## Expected Behavior

GIVEN a folder of cc.json.gz files

WHEN trying to merge them (and probably more commands)

THEN all files are combined

## Actual Behavior

The .gz files are not possible to be read by the merge filter.

Use `ccsh merge`

## Specifications
- CodeCharta Version: 1.85.0
- OS: all
- Browser: not applicable

phanlezz:
# Make sure all command line tools understand .gz files

Issue: #2580 

## Description

Enables .gz support for all analysis cli tools

Closes #2580


analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/CompressedStreamHandler.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    fun handleInput(input: InputStream): InputStream {
+        var content = input
+        if (!input.markSupported()) content = BufferedInputStream(input)
+        content.mark(2)
+        val byteArray = content.readNBytes(2)
+        content.reset()
+        if (isGzipByteHeader(byteArray)) return GZIPInputStream(content)
+        return content
+    }
```
ce-bo, line 17: - Maybe find a better name of this function that expresses what the function is actually doing.
- Please always use `{}` for control / if / loop statements (even for one line if-blocks).

ce-bo (commented):
It is looking good to me. It would be nice if anybody else could test this locally.

Another thing: We are able to "pipe in" a stream when executing a ccsh command (e.g. https://github.com/MaibornWolff/codecharta/blob/main/analysis/filter/StructureModifier/README.md). Would you please check, if it is possible to pipe in a gziped file?

Thank you.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Added 🚀
 
 -   Added missing documentation for ccsh filters and csv exporter [#2831](https://github.com/MaibornWolff/codecharta/pull/2831)
+-   Added support for compressed files to all analysis filters and to `check` validation
```
IhsenBouallegue, line 13: Please add the PR link

IhsenBouallegue:
We are trying to get away from using Spek for testing. For new tests, please use the JUnit library. If refactoring an old test that uses Spek please convert it to JUnit. It should be a simple syntax update.

CHANGELOG.md
phanlezz: Done ✅ Good catch ty

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/CompressedStreamHandler.kt
phanlezz: `handleInput` => `wrapInput`

ce-bo:
> > It is looking good to me. It would be nice if anybody else could test this locally.
> > Another thing: We are able to "pipe in" a stream when executing a ccsh command (e.g. https://github.com/MaibornWolff/codecharta/blob/main/analysis/filter/StructureModifier/README.md). Would you please check, if it is possible to pipe in a gziped file?
> > Thank you.
> 
> It is possible to pipe in compressed files ✅ I just tested it.
> 
> Bildschirmfoto 2022-06-07 um 09 57 49

Thank you.
Does it work for other parsers/filters/tools as well?

IhsenBouallegue (approved):
LGTM 🥳

phanlezz:
> [...]
> Thank you. Does it work for other parsers/filters/tools as well?

Works with:
- [ ]  Modify
- [ ]  RawTextParser


The other importers, exporters, and tools like merge, edgefilter, and check do not support to pipe into, or don't handle `cc.json` files

ce-bo:
> > [...]
> > Thank you. Does it work for other parsers/filters/tools as well?
> 
> Works with:
> 
> * [x]   Modify
> * [x]   RawTextParser
> 
> The other importers, exporters, and tools like merge, edgefilter, and check do not support to pipe into, or don't handle `cc.json` files
> 
> @ce-bo Or was the implication behind the question and part of the issue, that **all other** importers/exporters should also support compressed files, e.g. `.xml.gz` ? Then I would need to change some more stuff 😄

No, I just wanted to be sure that to pipe in a file is still possible. Thank you :)

PR #2822, opened 2022-05-24
use interface for interactive parsers + implement in ValidatorTool

luelista:
# use interface for interactive parsers + implement in ValidatorTool

Issue: #2341

## Description

- Use interface for interactive parsers (this allows us to automatically detect them from the subcommands, and makes a separate map unnecessary)
- Add the interactive dialog for the ValidatorTool 
- Add the tests for the dialog using Spek framework

### DIFF
--- a/analysis/tools/InteractiveParser/src/main/kotlin/de/maibornwolff/codecharta/tools/interactiveparser/InteractiveParser.kt
+++ b/analysis/tools/InteractiveParser/src/main/kotlin/de/maibornwolff/codecharta/tools/interactiveparser/InteractiveParser.kt
@@ -1,0 +1,5 @@
+package de.maibornwolff.codecharta.tools.interactiveparser
+
+interface InteractiveParser {
+    fun callInteractive()
+}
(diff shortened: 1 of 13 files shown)


analysis/tools/InteractiveParser/src/main/kotlin/de/maibornwolff/codecharta/tools/interactiveparser/InteractiveParser.kt
```diff
@@ -0,0 +1,5 @@
+package de.maibornwolff.codecharta.tools.interactiveparser
+
+interface InteractiveParser {
+    fun callInteractive()
```
IhsenBouallegue, line 4: The name `callInteractive` can create a bit of confusion. It can seem as if it's an alternative to the method `call` that's needed for the CLI which includes the logic of the parsers. This method however has nothing to do with the logic. I suggest `mainWithInteractiveDialog` or `executeWithInteractiveDialog`

IhsenBouallegue (approved):

IhsenBouallegue:
Can't merge because of test coverage is under 80%

IhsenBouallegue:
Please add a changelog entry

IhsenBouallegue (approved):

PR #2824, opened 2022-05-26
Refactor/2318/migrate distribution metric chooser

shaman-apprentice:
migrate distribution chooser and distributionMetric.service

ref #2318


IhsenBouallegue (approved):
LGTM 👍

IhsenBouallegue:
Quick question. Why is the old AngularJS metricChooser deprecated but still present? Are we unable to remove it completely?

shaman-apprentice:
> Quick question. Why is the old AngularJS metricChooser deprecated but still present? Are we unable to remove it completely?

the metricChooserController was host of the distribution metric chooser and is still host of the area, height and color metric chooser

PR #2833, opened 2022-06-03
Implement modify Interactive CLI

IhsenBouallegue:
# Implement modify Interactive CLI

Issue: #2341

## Description

- Add the interactive dialog for the modify 


analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/ParserDialog.kt
```diff
@@ -0,0 +1,46 @@
...
+            )
+            return listOf(
+                inputFolderName,
+                "--output-file=$outputFileName",
+                "--print-levels=$printLevels",
+                "--set-root=$setRoot",
+                "--move-to=$moveTo",
+                "--move-from=$moveFrom",
+                "--remove=$remove",
+            )
```
luelista, line 43: We need the possibility to only specify some of the options, so either  
a) ask before each step if we want to even do this (e.g. "Do you want to extract a path within the project?" -> only if yes ask for root and add the --set-root option), or   
b) check whether the user left the option empty and then skip the option (then we should tell the user that everything is optional, e.g. "If you want to extract a path within the project, specify the new root path:").

In case b), we also shouldn't ask follow-up questions if the first of a group was left empty (don't ask where to move, if nothing to move was specified)

luelista (changes requested):

luelista (approved):

PR #2836, opened 2022-06-07
Implement csvexporter Interactive CLI

IhsenBouallegue:
# Implement csvexporter Interactive CLI

Issue: #2341

## Description

- Add the interactive dialog for the csvexporter 


analysis/export/CSVExporter/src/test/kotlin/de/maibornwolff/codecharta/exporter/csvexporter/ParserDialogTest.kt
```diff
@@ -0,0 +1,45 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class ParserDialogTest {
+
+    @AfterAll
+    fun afterTest() {
+        unmockkAll()
+    }
+
+    @Test
+    fun `should output arguments that are parsed correctly`() {
```
ce-bo, line 28: This is a test for the happy path. Do we have another one for the "negative" path. For example if the user specifies a string instead of a number for the third (compressing) question.

analysis/export/CSVExporter/src/main/kotlin/de/maibornwolff/codecharta/exporter/csv/ParserDialog.kt
```diff
@@ -0,0 +1,33 @@
...
+
+            val defaultOutputFileName = getOutputFileName(inputFileName)
+            val outputFileName: String = KInquirer.promptInput(
+                message = "What is the name of the output file?",
+                hint = defaultOutputFileName,
+                default = defaultOutputFileName
+            )
+
+            val maxHierarchy: BigDecimal =
+                KInquirer.promptInputNumber(message = "Do you want to compress the file?", hint = "10", default = "10")
```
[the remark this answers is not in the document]
IhsenBouallegue: The question is actually wrong! It should be what is the maximum hierarchy?

analysis/export/CSVExporter/src/main/kotlin/de/maibornwolff/codecharta/exporter/csv/ParserDialog.kt
```diff
@@ -0,0 +1,33 @@
...
+
+            return listOfNotNull(
```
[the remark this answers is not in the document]
IhsenBouallegue: The interactive parser dialog is not responsible for that. It only collects arguments for the CLI then executes the CLI with those arguments. If the file is not found, the CLI will output an error.

analysis/export/CSVExporter/src/test/kotlin/de/maibornwolff/codecharta/exporter/csvexporter/ParserDialogTest.kt
IhsenBouallegue: It is not possible to specify a string when prompted for a number. That is handled from the Kotlin Inquirer library when using `promtNumberInput(...)` instead of `promtInput(...)`

ce-bo: What is happening, if you specify a wrong number (e.g. a string)? I would ensure that the parser does not crash in this case and for example that the parsers asks again for the correct number. It this is already the case, then that's fine for me.

Issue #2823, opened 2022-05-24
Standalone app doesn't work


shaman-apprentice:
# Bug

## Expected Behavior
Given a new version is released
When downloading the assets
Then we get executable for Windows, Mac and Linux.

## Actual Behavior
See [ci output of job "Build visualization"](https://github.com/MaibornWolff/codecharta/runs/6468255242?check_suite_focus=true) and available assets of [release 1.96.0](https://github.com/MaibornWolff/codecharta/releases/tag/1.96.0):

- Windows version is not build due to missing `wine` in build image
- Mac x64 is not build due to a bug in [nwjs-builder-phoenix](https://github.com/evshiron/nwjs-builder-phoenix); maybe related to [this bug](https://github.com/evshiron/nwjs-builder-phoenix/issues/106) and [this bug](https://github.com/evshiron/nwjs-builder-phoenix/issues/139)
- When I download Linux x64 version and want to run it in my Ubuntu 20.04 WSL I get the following error: `(codecharta-visualization:293): Pango-ERROR **: 17:54:03.066: Harfbuzz version too old (1.2.1)`

## Open questions
- Do we need any standalone version and if so do we need Windows, Mac and Linux? -> erstmal ja
- For Linux: Is it the user responsibility to have the right version of `Harfbuzz` or out responsibility to use a not outdated version?
- **nwjs-builder-phoenix isn't maintained anymore. Do we want to migrate away from it?**

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.96.0
- OS: Windows, Mac, Linux

IhsenBouallegue:
If I understood correctly, we need to fix the standalone builds.
But in this issue it was mentioned that we need to remove them and keep the web version (which works on all platforms) https://github.com/MaibornWolff/codecharta/issues/2573.
I don't see a lot of benefit from the windows linux and mac versions. It's a lot more maintainability for little benefit.

BridgeAR:
We might have different understandings when it comes to this: #2573 was about simplifying the build process for the standalone version. We used to have two types of builds: one for the web that caused different javascript code to be executed than in the standalone version. Now both builds are identical but we still have to be able to execute the standalone version properly.

So if I understood this issue correct it's only about not being able to properly start the standalone version at the moment and that should definitely be the case.

[...]

PR #2847, opened 2022-06-15
fix packaging for macOS and Windows

luelista:
issue #2823

## Description

nw for mac packaging needs exact nwVersion

for Windows it needs Wine installed in the build environment

### DIFF
--- a/visualization/package.json
+++ b/visualization/package.json
@@ -39,6 +39,7 @@
 	},
 	"main": "dist/webpack/index.html",
 	"build": {
+		"nwVersion": "0.65.1",
 		"output": "./dist/packages/",
 		"outputPattern": "codecharta-visualization-${VERSION}-${PLATFORM}-${ARCH}",
 		"files": "./dist/webpack/**/*",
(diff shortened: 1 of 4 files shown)


visualization/package.json
```diff
@@ -39,6 +39,7 @@
 	},
 	"main": "dist/webpack/index.html",
 	"build": {
+		"nwVersion": "0.65.1",
```
IhsenBouallegue, line 42: Note: I think at some point we removed it because it was causing some problem with updates. Renovate can't detect this line, so please make sure every time nw is updated to update this line manually!

IhsenBouallegue (approved):

visualization/package.json
ce-bo: I assume that renovate is also able to take care of this fixed version. We should have a look on that later, because most likely this will be forgotten to be updated in the future.

ce-bo:
Does this fix solve #2823?

visualization/package.json
luelista: I added a regex to renovate.json to automatically check this for updates. I don't know how to test this, however. Any ideas?

luelista:
Now it should also fix windows packaging

visualization/package.json
IhsenBouallegue: I have no idea how to test it, so maybe just wait and see what happens in the next update.

PR #2854, opened 2022-06-18
fix: creating .cc.json files not possible anymore 2850

### LINKED ISSUE #2850: Creating .cc.json Files not possible anymore
# Bug

## Expected Behavior

WHEN
`npm i -g codecharta-analysis`
`ccsh sonarimport https://sonarcloud.io/ maibornwolff-gmbh_codecharta_visualization -o cc-viz-2`

THEN
the output file should be created.

## Actual Behavior
the process is hanging up and no zip file is created.

## Notes:
Version 1.97.0 is working properly:
`npm i -g codecharta-analysis@1.97.0`
`ccsh sonarimport https://sonarcloud.io/ maibornwolff-gmbh_codecharta_visualization -o cc-viz-2`

## TODO:
Please check other ccsh commands, so that they are working properly.

phanlezz:
# Creating .cc.json files not possible anymore

Issue: #2850

## Description

This PR will fix the issue that all analysis parts, that were able to merge piped projects, were stuck on the final part of the processing. This was happening  in the final processing step of each modul, that was able to merge a piped-in project. When there was no project given, the issue occurred.

Closes #2850


phanlezz:
When a project is (or is not) piped into an analysis part, the input stream will be handled by the `ProjectDeserializer` even when there is actually no project on `System.in` . The method simply returns `null` in that case. Before that the `CompressedStreamHandler` would try to determine if the input stream is compressed and read its header. But, because it was really only intended to be called on actual files, the handler would get stuck on empty streams from `System.in`. 

> So it is as @ce-bo  probably suspected, piped projects cause problems. However, not in their presence, but in their absence.

I'm not quite sure if this fix is enough. I was able to produce files again, and pipe compressed and uncompressed files into sourcecodeparser, but I am not certain that there might not be a rare case where the parallelism of the pipe operator may cause problems.

ce-bo:
Would you please add a changelog entry?

phanlezz:
This is a workaround fix. We need to change the behaviour of the our Parsers. Details:
https://github.com/MaibornWolff/codecharta/issues/2858

Issue #1532, opened 2020-12-01
Default metric is preselected


Gaugi28:
# Feature request

<Please note by far the quickest way to get a new feature is to file a Pull Request.>
<We will consider your request but it may be closed,>
<if it's something we're not actively planning to work on>
<or it does not match our visison of this software.>

## Description

As user who don’t know the metrics and their meaning, I want a default metric which fits to the file so that I don’t have to select the right metric by myself.

## Acceptance criteria
- if a user loads his/her data the right metric is pre-selected.

## Open questions
- How do the system know which metric is the right/default one?

BridgeAR:
@ce-bo and me discussed this a few times recently and we thought about implementing a heuristic for known metrics based upon frequently used metrics in audits. We also thought about using machine learning. To do so we might need to ask users to transfer some statistics anonymously about their used metrics, the map size and a few more things.

BridgeAR:
Specified requirements:

1. In our cc.json we add a new property that includes the information what three (or four with the edge metrics) metrics should be chosen as default metrics while loading the map.

![image.png](https://images.zenhubusercontent.com/5e995c73efe923788cbf2771/b8230248-4bcb-48a2-951e-6996043e195d)

1. Check our parsers/importers what metrics they generate and try to chose a logical frequently used combination that seems useful to most people.

1. Update our parsers/importers to always include the default metrics in the generated file.

1. Create a fallback mode in the frontend for older cc.json files that do not yet include the default metrics in. This may be done by using the well defined combinations already known from our parsers/importers and to check if any of those apply. If neither applies, use the first entry in the metric list.

1. Update our cc.json file merge tool to use the "best" default set from the provided ones. This may be done by prioritizing some metrics above others. If neither applies or a file has no defaults, just use the first entry in the metric list.

BridgeAR:
I updated the requirements to simplify the issue:

1. Check our parsers/importers what metrics they generate and try to chose a logical frequently used combination that seems useful to most people. 

2. Create a default mode in the frontend for cc.json files. This may be done by using the well defined combinations already known from our parsers/importers and to check if any of those apply (as full metric combination). If neither applies, use the first entry in the metric list.

ce-bo:
Example:

**Known combination 1:**
Size / Complexity / Complexity

Category Size: rloc, loc, number_of_statements, etc.
Categroy Complexity: mcc, complexity, sonar_cognitive_complexity, etc.

jannikr:
[...]

PR #2862, opened 2022-06-22
Select matching default metrics

IhsenBouallegue:
# Select matching default metrics

Issue: #1532

## Description

Select a combination of default metrics based on available metrics. When no matching predefined combination has been found, metrics will be chosen randomly.
(good combinations of default metrics are predefined as scenarios in `scenarios.json`)


MW-Friedrich (approved):
LGTM!

PR #2863, opened 2022-06-22
fix: set default value for compressed in ParserDialogs to true

luelista:
issue #2341

# set default value for compressed in ParserDialogs to true

The default value on command line is true (files are not compressed only if the -nc param is specified), so it should be the same in interactive dialogs.


ce-bo (approved):
LGTM, Thanks!

Issue #2619, opened 2022-01-21
Import Metrics of MetricGardener


ce-bo:
# Feature request

Before this can be done, the release of MetricGardener must be planed (e.g. publish on npm.js).

## Description

As a User, I want to import [MetricGardener ](https://github.com/MaibornWolff/metric-gardener/)imports, so that I am able to visualize standard metrics (e.g. complexity, size, etc.) and dependency metrics (e.g. outgoing/incoming coupling) for a lot of languages.

## Acceptance criteria
- [x] A new importer in CodeCharta Analysis can be called from the User to run and import MetricGardener files (execute MetricGardener as a terminal command via npm).
- [x] MetricGardener needs to be published to the npm repository. [Try me](https://www.npmjs.com/package/metric-gardener)
- [x] The importer provides interactive CLI usage.
- [x] The outcomming structure of MetricGardener is transformed to our .cc.json structure
- [x] The importer exports .cc.json files in zipped format per default.

## Development notes (optional Task Breakdown)
- Use the npm package https://github.com/MaibornWolff/metric-gardener/ to parse source code in lots of programming languages
- MetricGardener files are volatile, the structure is not stable and will most likely change
- The file pathes must be transormed appropriately to match .cc.json structure

MW-Friedrich:
We still need to find a way to elegantly execute npm cli commands from within our Kotlin code. So far we have looked into:

- Java's Runtime.exec
- Turtle (https://github.com/lordcodes/turtle)
- Running a shell script via Runtime.exec or Turtle
- Gradle Node Plugin (https://github.com/node-gradle/gradle-node-plugin)
- Kotlin Multiplatform (https://kotlinlang.org/docs/multiplatform.html)

For release 1.101 our focus was more on the parsing from a metric-gardener json to our cc.json, so we only tried out these solutions experimentally.

BridgeAR:
@MW-Friedrich please do not use something like the Gradle Node Plugin. It is specific to Node.js but what, if we want to use something else? Then we need another individual solution. The command should just be executed as any other shell command. Using turtle for it is probably fine.

ce-bo:
I had a look on this. I ended up running metric-gardener npm binary using Turtle. This worked for me. But I am not sure if it is a good idea. @BridgeAR what do you think?
```    @Test
    fun `should run metric gardener help`() {

        val userDir = System.getProperty("user.dir")
        println(userDir)

        val os = System.getProperty("os.name")

[...]

PR #2875, opened 2022-06-30
Feature/2619/import from metric gardener

Chrisp485:
# Import metrics via Metric Gardener

Issue: #2619

## Description

This PR is the first part of importing metrics from metric-gardener. The parser receives a m-g,json and parse it into the cc.json format. 

- [x] Write logic to parse metric.json to cc.json
- [x]  Add Interactive Dialog
- [ ]  Write Tests


jannikr:
In order to find the pipeline error I decoupled the MetricGardenerImporter from the ccsh file. This change leads to another error in the pipeline. I could not reproduce these errors locally with the commands either:

```
./gradlew
./gradlew build integrationTest
./gradlew jacocoTestReport
```

charlotteKehm (approved):

PR #2880, opened 2022-07-01
Refactor/2318/rework custom dispatch

shaman-apprentice:
This PR does:
- remove need for custom dispatch logic
- remove camera.service
- remove cameraTarget.service


visualization/app/codeCharta/ui/codeMap/threeViewer/threeCameraService.ts
```diff
@@ -1,49 +1,21 @@
...
-	onCameraChanged() {
-		this.throttledCameraChange()
-	}
 
 	init(containerWidth: number, containerHeight: number) {
 		const aspect = containerWidth / containerHeight
 		this.camera = new PerspectiveCamera(ThreeCameraService.VIEW_ANGLE, aspect, ThreeCameraService.NEAR, ThreeCameraService.FAR)
-		this.setPosition()
-		ThreeOrbitControlsService.subscribe(this.$rootScope, this)
+		this.setPosition(new Vector3(0, 300, 1000))
```
ce-bo, line 14: I am just wondering where do you got these values here.

visualization/app/codeCharta/util/customConfigBuilder.ts
```diff
@@ -22,7 +23,8 @@ export function buildCustomConfigFromState(configName: string, state: State): Cu
 			dynamicSettings: undefined,
 			fileSettings: undefined,
 			treeMap: undefined
-		}
+		},
+		camera: cameraPosition
```
ce-bo, line 27: What's happening if we have stored an old custom config without the camera position property?
I think there could be an error, if we try to resolve the old custom config structure against the new one.

visualization/app/codeCharta/util/customConfigBuilder.ts
```diff
@@ -51,7 +53,6 @@ function initializeAppSettings(target: CustomConfig) {
 		},
 		amountOfEdgePreviews: 0,
 		amountOfTopLabels: 0,
-		camera: undefined,
```
ce-bo, line 54: Same goes here. We should check that this won't produce any errors.

visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -155,7 +154,8 @@ export class CustomConfigHelper {
 				customConfigVersion: exportedConfig.customConfigVersion,
 				mapChecksum: exportedConfig.mapChecksum,
 				mapSelectionMode: exportedConfig.mapSelectionMode,
-				stateSettings: exportedConfig.stateSettings
+				stateSettings: exportedConfig.stateSettings,
+				camera: exportedConfig.camera
```
ce-bo, line 158: exportedConfig.camera is not available in old configs that has been created based on current main.

ce-bo (commented):
First review: The removal of dispatching the camera is LGTM. I think now some breaking changes has been introduced unintentionally.

visualization/app/codeCharta/ui/codeMap/threeViewer/threeCameraService.ts
shaman-apprentice: It was the default value, which got applied initially and whenever an action with undefined was fired from [here](https://github.com/MaibornWolff/codecharta/blob/main/visualization/app/codeCharta/state/store/appSettings/camera/camera.actions.ts#L22)

visualization/app/codeCharta/util/customConfigBuilder.ts
shaman-apprentice: Good point. I haven't thought about backwards compatibility. You are right, it will throw an error.

Custom configs are still experimentally. Do you think, we should support both formats? My first thought is to keep it simple and don't transform old format into new one. I wouldn't even increment major version, as experimental features are allowed to introduce breaking changes.

But being backwards compatibly would be a simple check for old field and applying it to new field. So we could easily support it. 

What do you think?

ce-bo: Good suggestion. Let's add the check and a TODO to remove it in 3 month. "TODO [2022-10-11] ...".
I think we can ignore such old and deprecated configs without the camera properties from October on.

shaman-apprentice: done. But I noticed that we need to adjust metric templates as well. Therefore we would need a major update, once we remove the wrapper.

visualization/app/codeCharta/util/customConfigBuilder.ts
shaman-apprentice: it passes my legacy tests :)

visualization/app/codeCharta/util/customConfigHelper.ts
shaman-apprentice: same as [this comment](https://github.com/MaibornWolff/codecharta/pull/2880#discussion_r919175014)

shaman-apprentice:
TODO: Add changelog

visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -262,15 +262,31 @@ export class CustomConfigHelper {
...
-		}, 100)
+		if (customConfig.camera) {
+			setTimeout(() => {
+				threeOrbitControlsService.setControlTarget(customConfig.camera.cameraTarget)
+				threeCameraService.setPosition(customConfig.camera.camera)
+			}, 100)
+		}
+	}
+
+	// Todo test
```
ce-bo: Remove todo test

visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -267,12 +267,24 @@ export class CustomConfigHelper {
...
+				threeCameraService.setPosition(customConfig.camera.camera)
+			}, 100)
+		}
+	}
 
-			store.dispatch(setCamera(customConfig.stateSettings.appSettings.camera))
-			store.dispatch(setCameraTarget(customConfig.stateSettings.appSettings.cameraTarget))
-		}, 100)
+	// TODO [2023-01-01] remove support
+	private static transformLegacyCustomConfig(customConfig: any) {
```
ce-bo, line 279: I would rename this method to express that the camera settings are transformed.

visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -155,7 +153,8 @@ export class CustomConfigHelper {
 				customConfigVersion: exportedConfig.customConfigVersion,
 				mapChecksum: exportedConfig.mapChecksum,
 				mapSelectionMode: exportedConfig.mapSelectionMode,
-				stateSettings: exportedConfig.stateSettings
+				stateSettings: exportedConfig.stateSettings,
+				camera: exportedConfig.camera
```
ce-bo, line 157: What if exportedConfig.camera is not defined because it is an old config that is going to be imported.
Does it just assign undefined like this? importedCustomConfig.camera = undefined? Or is an undefined error thrown?

ce-bo (approved):
LGTM!

visualization/app/codeCharta/util/customConfigBuilder.ts
shaman-apprentice: as discussed tested and todo added for later removal

visualization/app/codeCharta/util/customConfigHelper.ts
shaman-apprentice: done with next push

PR #2881, opened 2022-07-05
fix: switch to nw-builder

luelista:
issue #2823

# switch to nw-builder

## Description

As described in issue #2823 we need to switch to a newer nwjs build tool, because nwjs-builder-phoenix references an outdated version of node-rcedit whcih doesn't work on x64-only systems, and also isn't maintained anymore since 2018.
Here we switch to nw-builder. This should fix win32 and win64 standalone builds.


CHANGELOG.md
```diff
@@ -12,15 +12,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 -   Fix resetting of color range within color settings panel [#2877](https://github.com/MaibornWolff/codecharta/pull/2877)
 -   Fix resetting of colors in delta mode within color settings panel [#2873](https://github.com/MaibornWolff/codecharta/pull/2873)
 -   Restore global settings on page load again [#2878](https://github.com/MaibornWolff/codecharta/pull/2878)
+-   Fix windows standalone builds [#2881](https://github.com/MaibornWolff/codecharta/pull/2881)
 
 ### Chore 👨‍💻 👩‍💻
 
 -   Migrate color settings panel to Angular [#2873](https://github.com/MaibornWolff/codecharta/pull/2873)
 
-## [1.99.3] - 2022-06-28
```
ce-bo, line 20: Would you please have a look at our .md files in gh-pages/_posts/release/? I think we should delete them as well.

ce-bo (approved):
Code is LGTM, but I did not tested it locally so far.

ce-bo (changes requested):
The following command is not working (anymore) on my system: `npm run start`
We use it to start the standalone app in dev environment.

Furthermore, I built and started the binaries locally and it worked so far. Thank you!

phanlezz (approved):

visualization/package.json
```diff
@@ -73,7 +53,7 @@
 		"e2e": "cross-env TEST_TYPE=\"e2e\" jest --config ./jest.config.json .*\\.e2e\\.ts --coverage false",
 		"e2e:auto": "npm run e2e -- --watch",
 		"e2e:ci": "npm run e2e -- --runInBand",
-		"package": "node -e 'fs.rmSync(\"dist/packages\", { recursive: true, force: true })' && build --concurrent --tasks win-x86,win-x64,linux-x86,linux-x64,mac-x64 --mirror https://dl.nwjs.io/ . && bestzip ./dist/packages/codecharta-visualization-$npm_package_version-web.zip ./dist/webpack",
```
ce-bo, line 76: Is the removel controlled by the configuration in build-nwjs.js?
I am not sure which property is responsible for that.

ce-bo (commented):
In general it is working for me :-)
One last thing: I ran `npm run package` and the ZIP file does not have the correct version number in it's file name. Is this correct?
image

phanlezz:
> In general it is working for me :-) One last thing: I ran `npm run package` and the ZIP file does not have the correct version number in it's file name. Is this correct? image

I think this is a windows problem. On @luelista 's machine it worked fine. But I can check if I can fix it on windows, too

visualization/package.json
phanlezz: But I will re-add this clean up call, because the implementation inside nw-builder does not reliably remove unused data

phanlezz:
The nw-builder's flag `option.zip` didn't do anything under windows and linux, but it worked under macOS. I could not find the problem or reason for this behavior. Instead, I decided to add manuell compressing (i.e. as we already do for the webpack). It might be reasonable to fix the problem with the nw-builder to be able to use the built-in compression process.

Manuell compression raised the following problems:
- the `osx64` app can't be compressed under windows, because `bestzip` prompts an error while using `fs` to `stat` some files. This didn't occur under linux.
- - without `osx64` you can package the project just fine, if you want to try it temporary
- All zip-files now have a top-level folder. e.g. `linux64/`
- - I changed the top-level folder of the web app from `dist/webpack/` to `webpack/`
- - If you want to compress the files without a top-level folder, the following problems occur:
- - - Native windows-zip is not able to decompress archives that are produced by `bestzip` directly in windows or linux. If `zip` is installed under linux, it will be used by `bestzip` and the produced archives can be opened by windows-zip just fine, even without a top-folder.

Conclusion: All `npm run package` tasks are now handled by a single file `build-nwjs.js` (clean, apps, compression). Packaging should work under osx and linux (with windows only if you remove `osx64` from the platform array, because the compression step at the end fails). All zip-files now have a top level folder, so that they can be opened by windows-zip even when `zip` is not installed. I added the `zip` program as a release dependency to `release_gh_pages.yml`, because it is faster.  File names now have the correct version numbers under windows.

If this is acceptable, please merge this to main and publish the next release, as it closes #2823 .

visualization/script/build-nwjs.js
```diff
@@ -0,0 +1,68 @@
...
+      cwd: path.join(platform.releasePath, "..")
+    }).then(function () {
+      console.log(name + " zipped!")
+    }).catch(function (err) {
+      console.error(err.stack)
+      process.exit(1)
+    });
+  });
+
+  zip({
```
BridgeAR, line 53: You use two run away promises here. Instead, you could just write:

```js
await Promise.all([
  zip({ ... }).
  zip({ ... }),
])
console.log('web and %s zipped', name)
```

Do not use explicit catch clauses. The error is going to be logged anyway.
Only logging the stack is also going to potantially hide important information of an error. Instead, always log the error. The stack is just a string that is a combination of `error.name + error.message + error.frames`. An error could include many other things as well such as a cause.

.github/workflows/release_gh_pages.yml
```diff
@@ -25,10 +25,10 @@ jobs:
...
 
       - name: Install wine
-        run: |
-          sudo dpkg --add-architecture i386
-          sudo apt update
-          sudo apt-get install -y wine-stable wine32 libwine:i386
+        run: sudo apt-get install -y wine-stable
+
+      - name: Install zip
+        run: sudo apt-get install -y zip
```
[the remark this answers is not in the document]
phanlezz: `zip` performed more reliable for me, because it didn't struggle in any situation, and `wine` is used for `.exe` modification.

renovate.json
```diff
@@ -38,14 +38,5 @@
 		}
 	],
 	"timezone": "Europe/Berlin",
-	"schedule": ["on the 2nd and 4th day instance on sunday after 9pm"],
```
ce-bo, line 41: Why this has been removed?

ce-bo: Ah ok it was the nw-phoenix-builder one.

visualization/script/build-nwjs.js
```diff
@@ -0,0 +1,64 @@
...
+      })
+    )
+  });
+  zipPromises.push(zip({
+    source: path.join("..", "webpack"),
+    destination: baseZipPath + "web.zip",
+    cwd: nw.options.buildDir
+  }));
+
+  await Promise.all([zipPromises]);
```
ce-bo, line 62: I think it should be:
```suggestion
  await Promise.all(zipPromises);
```

PR #2891, opened 2022-07-17
Refactor/2318/migrate file panel

### LINKED ISSUE #2879: Displayed selected files differ from rendered map
# Bug
Displayed selected files differ from rendered map

## Expected Behavior
Given I select some files and switch between standard and delta mode then the displayed selected file should be the same as the rendered file.

## Actual Behavior

At least in the following two scenarios displayed selected files differ from rendered map:

### Scenario 1
1. Switch to delta mode
1.1 UI shows reference and comparison file as empty
1.2 In store first file is selected as "Reference" and second file as "None"
image
2. Select second file as "Reference"
2.1 CodeCharta map doesn't fit to selection in file panel
image

### Scenario 2
1. Switch to delta mode
1.1 UI shows reference and comparison file as empty
1.2 In store first file is selected as "Reference" and second file as "None"
image
2. Select first file as "Reference" and second file as "Comparison"
image
3. Switch back to standard mode
4. Switch back to delta mode
4.1 CodeCharta map doesn't fit to selection in file panel
image

### LINKED ISSUE #2892: Restore previous selected files when switching back from delta to standard mode
# Feature request

## Description
When I switch back from delta to standard mode, I want my previous selected Files restored, so don't have to reselect them again.

In case I want to work with all files again, I can simply click "All".

E.G.:
image -> Delta Mode -> Standard Mode -> only "blade" is selected

shaman-apprentice:
- migrate / remove `recentFiles.service` (ref #2318)
- migrate `filePanel.component`
- close #2879
- close #2892


jannikr (changes requested):
- [x] #2879
- [ ] #2892 

There is just one issue with #2892. When I switch back from delta mode to standard mode the state is correct as described in the issue. However, the representation is not correct. I think the representation is taken over by the Delta mode.

visualization/app/codeCharta/state/selectors/allNecessaryRenderDataAvailable/utils/areDynamicSettingsAvailable.spec.ts
```diff
@@ -2,9 +2,9 @@ import { ColorMode, DynamicSettings, SortingOption } from "../../../../codeChart
 import { areDynamicSettingsAvailable, _isDynamicSettingAvailable } from "./areDynamicSettingsAvailable"
 
 describe("areDynamicSettingsAvailable", () => {
-	it("should return false when recent files aren't set", () => {
+	it("should return false when colorRange aren't set", () => {
 		const dynamicSettings = {
-			recentFiles: null
+			colorRange: null
```
MW-Friedrich, line 7: Is the method name of this test still applicable after recent files are deleted? Is the colour range really that critical to the dynamic settings?

shaman-apprentice:
> * [x]  #2879
> * [ ]  #2892
> 
> There is just one issue with #2892. When I switch back from delta mode to standard mode the state is correct as described in the issue. However, the representation is not correct. I think the representation is taken over by the Delta mode.

As discussed this seems to be an existing bug, which wasn't detected so far due to the bug of #2879. I created #2893 for it, which we can tackle separated from this PR after a merge.

visualization/app/codeCharta/state/selectors/allNecessaryRenderDataAvailable/utils/areDynamicSettingsAvailable.spec.ts
shaman-apprentice: good question. This is just an example action and not optimal. But better as it was before, when I introduced it :)

MW-Friedrich: Very fair point, thanks for the explanation.
LGTM! 🚀

Issue #2896, opened 2022-07-21
Big maps cause significant CPU time during mouse movement


BridgeAR:
# Bug

## Expected Behavior

GIVEN a big map with lots of nodes

WHEN hovering the mouse over the buildings fast

THEN the CPU time is low and the movement is smooth

## Actual Behavior

The CPU time is very high and the movement is lagging.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.100.0

Issue #2895, opened 2022-07-21
Significant CPU time during usage tracking


BridgeAR:
trackEventUsageData causes significant overhead.

PR #2897, opened 2022-07-21
Fix 2896 and 2895

BridgeAR:
[fix: remove trackEventUsageData to reduce performance overhead](https://github.com/MaibornWolff/codecharta/pull/2897/commits/56ba230e791565ebd897d59c005ab86d06f78ae0) 

The implementation had a bad time complexity and caused some maps
to take a significant time to load. Remove the tracking as it's
currently not used. We might need that at a later point again. If
that's the case, let's rework the logic.

----------

Fixes: https://github.com/MaibornWolff/codecharta/issues/2895

[fix: use a debounce while hovering buildings](https://github.com/MaibornWolff/codecharta/pull/2897/commits/0df7621297c19ce264932f578e26f439fe5be0eb) 

Hovering a building causes a calculation to draw the name of the
node. Maps with lots of small nodes trigger lots of these
calculations during hovering and that is causing lags. To overcome
the issue, a debounce time is introduced that is pretty much not
noticable by the user while reducing the calculations significantly.

Fixes: https://github.com/MaibornWolff/codecharta/issues/2896


visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -21,6 +24,14 @@ export class CodeMapArrowService
 		CodeMapMouseEventService.subscribeToBuildingUnhovered(this.$rootScope, this)
 		ThreeSceneService.subscribeToBuildingSelectedEvents(this.$rootScope, this)
 		ThreeSceneService.subscribeToBuildingDeselectedEvents(this.$rootScope, this)
+
+		this.debounceCalculation = debounce(hoveredBuilding => {
```
knoffi, line 28: This looks good. But shouldn't we add a test in the spec-file to check whether debounce was used/ the debouncing-effect occurred at all?
Or would you "refer this matter to Q/A"?

knoffi (approved):

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
knoffi: Seems to work more fluently now (on my local machine). :)

BridgeAR: I am fine adding a check before the wait in the test to verify that the debounce is working as expected.

knoffi: For the test, please check (the small) commit [95870f4](https://github.com/MaibornWolff/codecharta/pull/2897/commits/95870f4ba9ea6b67e32a5d56c871eb373f895b30) .

Much of it happened while pair-programming with Cedrik, but I would appreciate it if you gave me your advice too.

MW-Friedrich:
🚀

PR #2906, opened 2022-07-25
Add more default metrics

MW-Friedrich:
# Add more default metrics for Preselection

Issue: #1532

## Description

As a continuation of PR #2862 this pull request adds more metrics that can be preselected when loading a cc.json


MW-Friedrich:
Based the branch not off of main, sorry

Issue #2861, opened 2022-06-21
Interactive usage of SonarImporter is failing


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN
trying to import CodeCharta metrics from SonarCloud by the interacitve SonarImporter Dialog without any authentication credentials

THEN
this should work 

## Actual Behavior
Java Exception occurs

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/174836435-e108c6a6-fa40-44f5-979d-1f180482764c.png)

## Steps to Reproduce the Problem
1. run `ccsh`
1. choose sonarimporter
1. answer the questions to match this request `ccsh sonarimport https://sonarcloud.io/ maibornwolff-gmbh_codecharta_visualization -o cc-viz`

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.99.0
- OS: all
- Browser: -

## Developer Notes
The Backend tries to fetch `https://sonarcloud.io/api/measures/component_tree?component=maibornwolff-gmbh_codecharta_visualization&qualifiers=FIL,UTS&metricKeys=&p=1&ps=500`
The request fails with `{"errors":[{"msg":"The \u0027metricKeys\u0027 parameter must contain at least one metric key"}]}`

phanlezz:
It is also not possible to give a list of metrics via the non-interactive cli. Comma-separated lists produce an error. (My guess would be, that the url escaping does not work at the moment.) The default cli seems to add a list of all metrics, while the interactive parser uses an empty list by default.
Example with two metrics given:
` https://sonarcloud.io/api/measures/component_tree?component=maibornwolff-gmbh_codecharta_visualization&qualifiers=FIL,UTS&metricKeys=statements, open_issues&p=1&ps=500`

PR #2913, opened 2022-07-26
Fix interactive usage of SonarImporter

MW-Friedrich:
# Fix interactive usage of SonarImporter

Issue: #2861 

## Description

When no list of metrics is supplied in the interactive usage of the ccsh, the corresponding command (and therefore the request url) contains "--metrics=" instead of omitting the flag and selecting all metrics. Same with the --user flag.


Chrisp485 (approved):
LGTM, tested it and works again!

MW-Friedrich:
@knoffi and I will merge this after the patch release from #2915 :)

Issue #2800, opened 2022-05-13
Improving CLI output name handling by adding file extensions if necessary


phanlezz:
If given an output file name like `-o fileName` the output file name should be `fileName.json` compressed in `fileName.json.gz`. 
If `fileName.gz` is given, the file name should be `fileName` compressed into `fileName.gz`.

=> If the user gives a file extension, it should be used if it fits (`.json`, `.json.gz`, or `.gz`). If no extension is given it will be added.

MW-Friedrich:
Just as a heads up, json files generated by CodeCharta should still end as .cc.json, not .json

ce-bo:
Moving issues from in progress back to backlog is mostly not ideal. They are likely to be forgotten but we already started work on in.

PR #2914, opened 2022-07-27
Improve CLI output name handling by adding file extensions

charlotteKehm:
# {Meaningful title}

Issue: #2800 Improving CLI output name handling by adding file extensions if necessary 

## Description

  1. if the user enters no outputfile, then a default output file is created
  2. every outputfile is checked if it has the ending .cc.json. if it does not have the ending, it is automatically inserted (has been implemented in a static/object class `OutputFileHandler.kt`
  3. In order to run `golden_test.sh`  and `TokeiImporterTest.kt` properly the variable systemout has been introduced to enable console output
  - Open issues:
  1. `golden_test.sh` needs to be updated. Importers that are being used in production aren't tested there (e.g. metricgardener, gitlog...). Importers that are not used in production are tested there (eg. crococosmo)
  2. A Decision has to be made about the default behavior of the outputfilehandling: If an outputfile shall be generated, then several issues need to be considered: conditions for serializing the project, is there a more elegant way to run the tests than using the variable systemout, 
 3. Raw Text Parser needs to be adjusted 
 4. It would be nice if the logic of `OutputFileHandler` would be transferred to the interface `ParserDialogInterface`
 5. `build.gradle`: task integrationtest: It is not possible to run the integrationtests on a windows notebook without having wsl enabled (via bash). All operating systems need to be considered, hence it is always possible to run the integration tests locally.


analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.Writer
+
+object OutputFileHandler {
+
+    fun checkAndFixFileExtension(outputName: String): String {
+        if (outputName.isEmpty()) return "default.cc.json"
+        val delimiter = extractDelimiter(outputName)
+        if (delimiter.isEmpty()) {
+            return outputName.substringBefore(".") + ".cc.json"
+                }
```
ce-bo, line 17: It seems like that the code style has not been formatted automatically.
I recommend to always add curly braces even for if-statements with one line.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.PrintStream
+import java.io.Writer
+
+object OutputFileHandler {
+
+    fun checkAndFixFileExtension(outputName: String): String {
+        if (outputName.isEmpty()) return "default.cc.json"
+        val delimiter = extractDelimiter(outputName)
+        if (delimiter.isEmpty()) {
+            return outputName.substringBefore(".") + ".cc.json"
```
ce-bo, line 16: This will not work if multiple dots are provided in the fileName like: "my.output.name". This would lead to my.cc.json which would not be correct. Instead you can use substringBeforeLast.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+     fun writer(outputName: String?, test: Boolean, output: PrintStream): Writer {
+         return if (test) {
+             OutputStreamWriter(output)
+         } else {
+             BufferedWriter(FileWriter(File(checkAndFixFileExtension(outputName ?: ""))))
+         }
+     }
+
+    private fun extractPath(outputName: String, delimiter: String): String {
+        return outputName.split(delimiter).dropLast(1).joinToString(delimiter) + delimiter
```
ce-bo, line 31: I would prefere existing Utility functions (examples: https://mkyong.com/java/how-to-get-the-filepath-of-a-file-in-java/).
Thus, different operating system depending pathes are handled automatically.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    private fun extractFileName(outputName: String, delimiter: String): String {
+        val fileName = outputName.split(delimiter).reversed().get(0)
```
ce-bo, line 35: Here I would do the same like in [comment](https://github.com/MaibornWolff/codecharta/pull/2914/files#r934216289)

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+        if (fileName.isEmpty()) {
+            return "default.cc.json"
```
ce-bo, line 37: Let's create a constant for "default.cc.json" string.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+        return fileName.substringBefore(".") + ".cc.json"
```
ce-bo, line 39: Use substringBeforeLast instead.

analysis/model/src/test/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandlerTest.kt
```diff
@@ -0,0 +1,49 @@
+package de.maibornwolff.codecharta.serialization
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+internal class OutputFileHandlerTest {
```
ce-bo, line 6: Please add more strings to be tested like "my.long.map.name.cc.json" or ".hiddeenfile" and so on. We should forget to test all the edge cases.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ProjectGenerator.kt
```diff
@@ -22,7 +22,7 @@ class ProjectGenerator(private val writer: Writer, private val filePath: String,
             project = MergeFilter.mergePipedWithCurrentProject(pipedProject, project)
         }
 
-        if (toCompress && filePath != "notSpecified") {
+        if (toCompress && filePath != "default.cc.json") {
```
ce-bo, line 25: Let's add/reuse the constant for default.cc.json

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
charlotteKehm: @ce-bo We were considering using substringBeforeLast. If you are providing a file name such as "test.cc.json", this would lead to test.cc.cc.json which would also be incorrect. Therefore we decided that by convention fileNames do not involve dots.

ce-bo: Restricting the filename would not be intuitive. So let's not do this. 
In case of "test.cc.json" we can simply check if "cc.json" is included or not. If yes, stop filename processing, if not, continue.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/RawTextParser.kt
```diff
@@ -92,11 +94,11 @@ class RawTextParser(
     }
 
     private fun getWriter(): Writer {
-        return if (outputFile != null) {
-            BufferedWriter(FileWriter(outputFile!!))
-        } else {
-            OutputStreamWriter(output)
+        if (!test) {
```
ce-bo, line 97: Is this condition only for debugging?

analysis/import/JasomeImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/jasome/JasomeImporter.kt
```diff
@@ -1,44 +1,52 @@
...
 
     override fun call(): Void? {
         val jasomeProject = JasomeDeserializer().deserializeJasomeXML(file!!.inputStream())
         val project = JasomeProjectBuilder().add(jasomeProject).build()
-        val filePath = outputFile?.absolutePath ?: "notSpecified"
+        val filePath = outputFile ?: "notSpecified"
 
-        if (compress && filePath != "notSpecified") ProjectSerializer.serializeAsCompressedFile(project, filePath) else ProjectSerializer.serializeProject(project, writer())
+        if (compress && filePath != "notSpecified") ProjectSerializer.serializeAsCompressedFile(project,
+                OutputFileHandler.checkAndFixFileExtension(filePath)) else ProjectSerializer.serializeProject(project, OutputFileHandler.writer(outputFile ?: "", systemout || test, output))
```
ce-bo, line 40: To not copying the `checkAndFixFileExtension` method to all importers/parsers, we should try to call this function within `serializeAsCompressedFile` (in de/maibornwolff/codecharta/serialization/ProjectSerializer.kt).

analysis/model/src/test/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandlerTest.kt
charlotteKehm: how about cases like "my.long.map.name". when using substringBeforeLast, we would get : "my.long.map.cc.json", which would be incorrect. shall we use some function that checks if the last substring is any file-ending or shall we only check for .json or .cc.json since we would not expect any other file ending

ce-bo: Your suggestion sounds good. Let's expect json and cc.json only.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/RawTextParser.kt
charlotteKehm: no this is for executing the parser dialog tests. previously the default condition was to write the output in terminal. now a default file is created. during the build it would not be convenient to generate all the default files (also many tests would need to be corrected) therefore we still write the output of .jsons during the build in terminal.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ProjectGenerator.kt
charlotteKehm: I could not find any file of constants in the project. Ist there one? Otherwise I would suggest to start one for all the importers (hence, in the directory: import).

ce-bo: I think YAGNI. Let's start with just one constant in the ProjectGenerator or in the OutputFileHandler.

charlotteKehm:
@ce-bo I resolved your remarks in my latest commit. 
I happen to have a problem with sonarcloud code coverage, because it highlights code as uncoverd I haven't written. These are in my opinion legacy issues and I am not sure how we shall deal with them.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
charlotteKehm: I am using the gradle target ktlintFormat and have enabled the codestyle.xml in my intellij-settings.
 I have seen in multiple files that curly braces are not used in if - statements with one line, for instance ```TokeiImporter.kt``` and assumed it would be a convention. Do I misunderstand something?
image

ce-bo: Everything is fine :) There are a lot of if statements without curly braces from the past. I think we discussed this for the visualization. We agreed to use curly braces for new code there. I think we should also do it in analysis part. But this is just a minor remark.

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/RawTextParser.kt
ce-bo: I think we should not change code only to be able to run unit tests properly. Why has the default behavior changed?

charlotteKehm: I can see your point, but I thought the requirement was if the user does not enter any input into the interactive terminal we still create a default cc.json file in the working directory.

ce-bo: Ah okay. I think this is not included in the scope of this issue. Now it is a bit mixed. 
I would rather name the default something like this <INPUT-FILE-NAME>-output.cc.json. But ideally we should change this in a separate PR. Currently the default output name is handled in de.maibornwolff.codecharta.tools.interactiveparser.ParserDialogInterface#getOutputFileName. But it seems that not all parsers/importers are using this method.

analysis/import/JasomeImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/jasome/JasomeImporter.kt
```diff
@@ -1,44 +1,51 @@
...
 
         return null
     }
 
+    /**
     private fun writer(): Writer {
         return outputFile?.writer() ?: System.out.writer()
     }
-
+**/
```
ce-bo, line 48: Please remove this.

ce-bo:
Let‘s Not change the default behavior when the User does not provide a output file name. Thus, the new systemout flag is not needed to be added in every parser. In the Parser Dialog Interface, I think it is called InteractiveParser or similar, we can extend  the method for providing the output file, so that the correct file ending is added if necessary. We can suggest a default Output file Name to the user in the interactive usage but the User should still be possible to not set any Output file name, so that the results are going to be dumped to System.out.

Chrisp485:
@ce-bo Sorry I'm still a bit confused. Do we want to achieve the following two Cases?
1. Users doesn't specify a filename -> The output gets written into stdout
2. User enters an filename -> We just check if the fileextension exists/ is correct, else we add or correct it.

So the changes with the  sysout flag need to be removed again and move the functionality of the new OutputFileHandler into the ParserDialogInterface?

ce-bo:
Yes, that is what I ment. Moving the OutputFileHandler can be checked. I did not have a deep look into that.

Hall-Ma (approved):

Hall-Ma (dismissed):
Sorry, wrong PR :(

phanlezz (approved):
Lets get this merged! 🚀

Issue #2538, opened 2021-11-29
Create zipped CodeCharta cc.json files during pipeline build


ce-bo:
# Feature request

In build_gh_pages.sh we create unzipped cc.json files. The default case is, that parsers are creating compressed (.gz) files.
Let's build compressed files during the build pipeline.
Then, to ensure the functionality of CodeCharta (current parsers are working with current visualization), the created and zipped files should be used in our e2e tests to test the behaviour automatically.

## Acceptance criteria
- During Build Pipeline .gz and unzipped files are generated
- When loading CodeCharta the zipped files should be loaded instead of unzipped ones to reflect the default case.
- E2E Test to test uploading compressed files is using the continuously updated compressed sample files.

## Development notes (optional Task Breakdown)
- Modify build_gh_pages.sh to generate unzipped and zipped files.

ce-bo:
@BridgeAR please have a look on this issue and feel free to adjust acceptance criteria.

jannikr:
This issue is not closed, because the merge and patch was just for testing.

Hall-Ma:
Still not working on our Website. We have to figure it out where we have to change the url.

PR #2917, opened 2022-07-28
Tech/2538/Load gz files via url parameter

Hall-Ma:
# Loading gz files via URL parameter

closes: #2538

## Description

- Implement loading of gz files via URL
- From now on, CodeCharta's zipped cc.json files will be used in every deployment.
- Exchange of the library zlib to [pako](https://www.npmjs.com/package/pako)

# Screenshots or gifs
image


visualization/app/codeCharta/util/urlExtractor.ts
```diff
@@ -1,8 +1,9 @@
 "use strict"
-import { ILocationService, IHttpService } from "angular"
+import { IHttpService, ILocationService } from "angular"
 import { NameDataPair } from "../codeCharta.model"
-import { ExportCCFile, ExportWrappedCCFile } from "../codeCharta.api.model"
+import zlib from "zlib"
```
BridgeAR, line 4: I recommend to use another zipping library instead. This is a very old unmaintained module that was intended as a drop in replacement for the Node.js one. By now they strongly deviate and it should probably not be used anymore. I also wonder if this is automatically added by webpack as it's not listed in our package.json as dependency?

visualization/app/codeCharta/util/urlExtractor.ts
```diff
@@ -39,15 +40,44 @@ export class UrlExtractor {
...
+		const options = { responseType: "blob" as const }
+		const response = await this.$http.get<Blob>(fileName, options)
+		if (response.status >= 200 && response.status < 300) {
+			return new Promise(resolve => {
+				const reader = new FileReader()
+				reader.readAsArrayBuffer(response.data)
+				let content: ExportCCFile
+
+				reader.onload = event => {
+					const readerContent = zlib.unzipSync(Buffer.from(<string>event.target.result)).toString()
```
BridgeAR, line 59: Using the blocking variant is not ideal as it'll block until it's fully unzipped. Instead, use the async version as it won't block the program from running in the meanwhile.

Hall-Ma: @BridgeAR I'm not sure what you mean. Please, could you specify it or give an example of how to do it?

BridgeAR: It is already resolved by using `pako` instead of `zlib`. `zlib` has blocking ("sync") and non-blocking operations (async, the regular name without sync). Sync APIs should pretty much never be used besides rare exceptions, if they exist next to an async one.

jannikr (approved):

Chrisp485:
We needed to change the e2e Test. The page.on eventlistener we are using, breaks after using it more then two times. It's a known [Issues](https://github.com/puppeteer/puppeteer/issues/5334)

Issue #2410, opened 2021-09-29
PoC: Switch from puppeteer to Cypress for E2E Testing


ce-bo:
# Feature request

<Please note by far the quickest way to get a new feature is to file a Pull Request.>
<We will consider your request but it may be closed,>
<if it's something we're not actively planning to work on>
<or it does not match our visison of this software.>

## Description

As a Developer, I want to be sure, that the switch to cypress will enable us to collect E2E coverage and that tests run stably.

## Acceptance criteria
- E2E Coverage can be calculated properly

## Development notes (optional Task Breakdown)
- [ ]  Implement a PoC before migrating every puppeteer test.

shaman-apprentice:
It seems downloads can not be tested nicely with current Puppeteer version. When migrating we might add a test for the screenshot-button. (Also see [this comment](https://github.com/MaibornWolff/codecharta/pull/2919#pullrequestreview-1055258939))

PR #2919, opened 2022-07-28
Refactor/2818/migrate screenshot button

shaman-apprentice:


visualization/app/codeCharta/ui/screenshotButton/screenshotButton.component.spec.ts
```diff
@@ -1,95 +1,39 @@
...
+		fixture.componentInstance.makeScreenshotToClipBoard = jest.fn()
 
-			screenshotButtonController.onScreenshotToClipboardEnabledChanged(
-				storeService.getState().appSettings.screenshotToClipboardEnabled
-			)
-
-			expect(screenshotButtonController["_viewModel"].screenshotToClipboardEnabled).toBe(true)
-		})
+		userEvent.click(screen.getByTitle("Copy screenshot to clipboard (Ctrl+Alt+F), export it as a file by (Ctrl+Alt+S)"))
+		expect(fixture.componentInstance.makeScreenshotToClipBoard).toHaveBeenCalled()
```
knoffi, line 28: This test would pass if we replaced
`
makeScreenShotToClipBoard = ( ) => { }`

Is there a way how we can make this test more "demanding" (=stricter assert part)?

visualization/app/codeCharta/ui/screenshotButton/screenshotButton.component.spec.ts
```diff
@@ -1,95 +1,39 @@
...
-			screenshotButtonController.makeScreenshotToFile()
-			expect(screenshotButtonController["buildScreenShotCanvas"]).toBeCalled()
-		})
+	it("should save to file on click, when screenshot to clipboard is not enabled", async () => {
+		Store.store.dispatch(setScreenshotToClipboardEnabled(false))
+		const { fixture } = await render(ScreenshotButtonComponent, { excludeComponentDeclaration: true })
+		fixture.componentInstance.makeScreenshotToFile = jest.fn()
+
+		userEvent.click(screen.getByTitle("Export screenshot as file (Ctrl+Alt+S), copy it to clipboard by (Ctrl+Alt+F)"))
+		expect(fixture.componentInstance.makeScreenshotToFile).toHaveBeenCalled()
```
knoffi, line 37: This test would pass if we would replace
`
makeScreenShotToFile = ( ) => { }`

Is there a way how we can make this test more "demanding" (=stricter assert part)?

knoffi (changes requested):
This is not a "hard" request, merely a proposal.

visualization/app/codeCharta/ui/screenshotButton/screenshotButton.component.spec.ts
shaman-apprentice: I like your way of thinking! 👍 

I applied the boys scout rule here. In an ideal world, the UI shouldn't contain the logic behind it at all, but just use some service, which does all the calculation. Then the UI has only the current test to use the service and the logic should be tested within the service.

Unfortunately the logic was already in the component and it wasn't tested but mocked away at a deeper level. 

So the tests cover the same as before and now we have an Angular component more and an AngularJS component less. Small improvements 🤓

shaman-apprentice: I added a reminder to #2410. Maybe we can add a nice e2e test later. But lets keep with small improvements (Angular Migration) for now. Although 👍  for the spirit!

Issue #2923, opened 2022-08-01
ccsh merge throws NoSuchMethodError


FlorianHoldt:
# Question
When calling ccsh merge in order to add custom metrics, I get a java exception complaining about a NoSuchMethodException:
Exception in thread "main" java.lang.NoSuchMethodError: java.io.InputStream.readNBytes(I)[B
        at de.maibornwolff.codecharta.serialization.CompressedStreamHandler.wrapInput(CompressedStreamHandler.kt:14)
        at de.maibornwolff.codecharta.serialization.ProjectDeserializer.deserializeProject(ProjectDeserializer.kt:33)
        at de.maibornwolff.codecharta.filter.mergefilter.MergeFilter.call(MergeFilter.kt:66)
        at de.maibornwolff.codecharta.filter.mergefilter.MergeFilter.call(MergeFilter.kt:14)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
        at picocli.CommandLine.access$1300(CommandLine.java:145)
        at picocli.CommandLine$RunAll.recursivelyExecuteUserObject(CommandLine.java:2431)
        at picocli.CommandLine$RunAll.recursivelyExecuteUserObject(CommandLine.java:2433)
        at picocli.CommandLine$RunAll.handle(CommandLine.java:2428)
        at picocli.CommandLine$RunAll.handle(CommandLine.java:2389)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
        at picocli.CommandLine$RunAll.execute(CommandLine.java:2391)
        at picocli.CommandLine.execute(CommandLine.java:2078)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh$Companion.main(Ccsh.kt:82)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh.main(Ccsh.kt)

## Preparation
I am using CodeCharta 1.101.1

I have Java 9 installed on my machine, this is the output of java -version:
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

Please check these resources first before posting a question

* [/] I searched for answers in [older issues](https://github.com/MaibornWolff/codecharta/issues?utf8=%E2%9C%93&q=)
* [/] I looked into the [README.md](https://github.com/MaibornWolff/codecharta)
[...]

PR #2930, opened 2022-08-02
ccsh merge throws NoSuchMethodException

phanlezz:
# Change CompressedStreamHandler usage of java11 method to java8 #2923 

Close: #2923

## Description

Fix for the NoSuchMethodException in CompressedStreamHandler through a switch of the java11 method to the java8 equivalent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix SonarImporter requesting no metrics from SonarQube when the list of metrics was left empty [#2913](https://github.com/MaibornWolff/codecharta/pull/2913)
--   Change CompressionStreamHandler usage of Java11 Method to Java9 [#2930](https://github.com/MaibornWolff/codecharta/pull/2930)
+-   Change CompressionStreamHandler's wrapInput usage of method readNBytes to read (java8) [#2930](https://github.com/MaibornWolff/codecharta/pull/2930)
 
 ## [1.101.1] - 2022-07-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix SonarImporter requesting no metrics from SonarQube when the list of metrics was left empty [#2913](https://github.com/MaibornWolff/codecharta/pull/2913)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix SonarImporter requesting no metrics from SonarQube when the list of metrics was left empty [#2913](https://github.com/MaibornWolff/codecharta/pull/2913)
+-   Change CompressionStreamHandler usage of Java11 Method to Java9 [#2930](https://github.com/MaibornWolff/codecharta/pull/2930)
```
ce-bo, line 23: I would name the affected Method in the Changelog, so that it is easier to track this error.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/CompressedStreamHandler.kt
```diff
@@ -11,7 +11,8 @@ object CompressedStreamHandler {
         var content = input
         if (!input.markSupported()) { content = BufferedInputStream(input) }
         content.mark(2)
-        val byteArray = content.readNBytes(2)
+        val byteArray = ByteArray(2)
+        content.readNBytes(byteArray, 0, 2)
```
ce-bo, line 15: Would you please Check, if we can use content.read() instead since it is supported by Java8 also. I think I just saw, that readNBytes is not available in Java8.

phanlezz: Addressed in d35f7cb 👍

CHANGELOG.md
phanlezz: Addressed in d35f7cb 👍

ce-bo (approved):
LGTM, thanks!
Regarding the changelog, I would try to express why the fix was necessary. In this case you fixed a NoSuchElementException due to a call of method `readNBytes()` that is not available in Java 9 and you replaced it with `read()`

ce-bo:
I editet the description from `Issue #2923` to `Close #2923`. By this, GitHub will close the referenced issue automatically.

phanlezz:
> LGTM, thanks! Regarding the changelog, I would try to express why the fix was necessary. In this case you fixed a NoSuchMethodException due to a call of method `readNBytes()` that is not available in Java 9 and you replaced it with `read()`

Ahh, that's better, thanks! I will add that 👍 

> I editet the description from `Issue #2923` to `Close #2923`. By this, GitHub will close the referenced issue automatically.

Nice, I had a close statement in commit ea8d3fb already, but this will work also 👍

PR #2931, opened 2022-08-03
remove conversion from single to multiple mode for custom configs

Hall-Ma:
# Remove conversion from single to multiple mode for custom configs

## Description

From today, support for custom views/custom configs that still have a single mode has been discontinued.


shaman-apprentice (approved):
LGTM

Hall-Ma:
@ce-bo Would you like to update the `CUSTOM_CONFIGS_LOCAL_STORAGE_VERSION` version to `2.0.0` Because configs that contain the single mode are no longer supported?

ce-bo:
@Hall-Ma I think we can keep v1. As @shaman-apprentice said, it is an experimental feature, so it is ok especially behause the User could use an old version of CC if he/she wants to use an old Custom View.

PR #2934, opened 2022-08-03
Fix/2823/Fix Web Package

phanlezz:
# Fix for the zip file of the webpack package being corrupted

Changing the zip path to not contain a directory up order, which causes problems under linux.

Issue: #2823

## Description

Currently the **-web.zip package in the release notes it not usable, because it is corrupted. This zip is packaged under linux during the release process. Because of a wrong path, the zip is unusable.


Chrisp485 (approved):
LGTM

PR #2935, opened 2022-08-04
fix: keep selected metrics when excluding buildings

### LINKED ISSUE #2920: Chosen metrics are going to be resetted on excluding a folder
# Bug

## Expected Behavior

GIVEN
loaded cc.json with changed metrics (different from preselected ones)

WHEN
excluding a folder using the FileExplorer or the context menu (right click on building)

THEN
the metrics should not be resetted

## Actual Behavior
the metrics are resetted to the default ones.

1. load map
1. change metrics
1. exclude folder or file (open File Explorer > ... > Exclude OR right click on building > exclude)

## Specifications
- CodeCharta Version: 1.100.1
- OS: all
- Browser: all

shaman-apprentice:
fix #2920


Hall-Ma (approved):
Everything works fine! Thanks :)

Issue #2744, opened 2022-03-21
Copy filenames to clipboard, according to all metrics


Christian-Eberhard:
# Feature request

## Description

As a CodeCharta, I want to copy the filenames of the highest / widest buildings so that i do not have to type every single filename myself.

## Acceptance criteria

- [x]  Button for copying filenames
- [x] Entries should be copied in plaintext
- [x] rloc: Filenames of Top 10 highest/widest buildings with metric value should be copied
- [x] selected metrics: Filenames of Top 10 highest/widest buildings with metric value should be copied
- [x] all metrics: Filenames of Top 10 highest/widest buildings with metric value should be copied

Clipboard text looks approximately like this:

rloc
  -   File1: value
  -   File2: value
  -   File3: value
  -   File4: value
  -   ...
  -   File10: value

mcc
  ... like rloc ...

## Open questions
Where in the UI should this functionality be?

BridgeAR:
This is actually a pretty important feature for audits. As such I just changed the priority.

LuisGuenther:
The logic for this feature needs to evaluate for every single metric which values are considered problematic for every single metric (high or low).
How should this be handled? There is already the feature of inverting colors in the frontend for this problem.
@knoffi suggested saving the individual metric values with different leading signs (e.g. test coverage with negative values because small values are considered "good") in the JSONs, which automatically will lead to adjustments in the analysis part. Compatibility of older JSONs also needs to be kept in mind.

@BridgeAR @ce-bo

ce-bo:
@LuisGuenther please see https://github.com/MaibornWolff/codecharta/pull/2942#discussion_r939909423

Hall-Ma:
@BridgeAR @ce-bo 
We have an example how we implemented list yet:
```
RLOC
    • sample2LeafMergedIn.kt (600)
    • sample1OnlyLeaf.scss (400)
    • bigLeaf.ts (120)
    • bigLeaf.ts (100)
    • otherSmallLeaf.ts (70)
    • otherSmallLeaf.ts (70)
    • smallLeaf.html (30)
    • smallLeaf.html (30)
FUNCTIONS
    • otherSmallLeaf.ts (1000)
    • smallLeaf.html (101)
    • smallLeaf.html (100)
    • sample2LeafMergedIn.kt (10)
    • otherSmallLeaf.ts (10)
    • bigLeaf.ts (10)
    • bigLeaf.ts (10)
[...]

PR #2942, opened 2022-08-04
Feature/2744/copy filenames to clipboard

LuisGuenther:
# Copy worst metric violators directly to your clipboard

Issue: #2744 

## Description

**Users want to quickly share a summary of files with high metric values (violators).**

**For this, we added a button (copy-to-clipboard) with which you can paste a summary of the worst violators. This summary contains a top10-list for every metric and it lists the names of the files with the highest metric values. We also included the metric value next to every filename.**

## Screenshots or gifs

MicrosoftTeams-image

MicrosoftTeams-image (2)


visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,45 @@
...
+			}
+		}
+	}
+	return new Map<string, string[]>()
+}
+
+type FileNamesByMetrics = Map<string, string[]>
+type FileToValue = { name: string; value: number }
+//sry Ruben
+function maybeAddToMap(key: string, value: number, fileName: string, map: Map<string, FileToValue[]>) {
```
ce-bo, line 23: Please try to find a better name. The method is adding key/value pairs per Metric with highest value to a limit of 10 entries per metric.

visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,45 @@
...
+				maybeAddToMap(key, value, data.name, fileNamesByMetrics)
+			}
+		}
+	}
+	return new Map<string, string[]>()
+}
+
+type FileNamesByMetrics = Map<string, string[]>
+type FileToValue = { name: string; value: number }
+//sry Ruben
```
ce-bo, line 22:
```suggestion
```

visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,45 @@
...
+function maybeAddToMap(key: string, value: number, fileName: string, map: Map<string, FileToValue[]>) {
+	const keyIsNew = !map.has(key)
+	if (keyIsNew) {
+		map.set(key, [{ name: fileName, value }])
+	} else {
+		const previousValues = map.get(key)
+		if (previousValues.length < 10) {
+			map.set(key, [...previousValues, { name: fileName, value }])
+		} else {
+			const minValue = Math.min(...previousValues.map(pair => pair.value), value)
```
ce-bo, line 32: I think this would be more performant: I would suggest to first create a list of all files sorted by metric value. After that I would just slice the 10 files with highest metric values off and use it as the list to be copied to clipboard.
Currently you have to find the MIN value "NUMBER_OF_TOTAL_FILES"-times.

ce-bo (commented):
Let's wirte a pretty changelog entry to inform the user about this new cool feature. We should add a screenshot to the changelog entry as well.

ce-bo (commented):
We should review the performance in big maps and we should do this asynchroniously.

visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
knoffi: At the moment we have for this method a runtime of approximately:

(For cc.json with 2391 and 1.5 MB)
 70 ms

(For cc.json with 13500 files and 20,7 MB)
400 ms

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	const newPair = { name: fileName, value }
+	const previousValues = map.get(key)
+	if (previousValues.length < MAX_ENTRIES) {
+		insertSorted(previousValues, newPair)
+		return
+	}
+
+	insertInFullDescendingList(newPair, previousValues)
```
Hall-Ma, line 34:
```suggestion
	insertSorted(previousValues, newPair)
	map.set(key, previousValues.slice(0, MAX_ENTRIES
```

You also could do this. You have a more maintainable and readable code and only use on insert method. This implementation is approx. 10 to 15 seconds slower than the current version.

Hall-Ma, line 34:
```suggestion
		insertSorted(previousValues, newPair)
        map.set(key, previousValues.slice(0, MAX_ENTRIES)
```
Or you could do this version. We could spare the if condition. Just an idea :slightly_smiling_face:

visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
knoffi: Resolved by commit https://github.com/MaibornWolff/codecharta/issues/2744

visualization/app/codeCharta/ui/copyToClipboardButton/getFilenamesWithHighestMetrics.ts
knoffi: Was applied on the fly.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+	describe("copyToClipboardButtonComponent", () => {
+		it("should write text to clipboard by using text from service", async () => {
+			const clipboard = navigator.clipboard
+			const writeToClipboardSpy = jest.spyOn(clipboard, "writeText")
+			const getClipboardTextSpy = jest.spyOn(component["service"], "getClipboardText")
+
+			await component.copyNamesToClipBoard()
+
+			expect(writeToClipboardSpy).toBeCalledWith("Magic Monday")
+			expect(getClipboardTextSpy).toBeCalled()
```
shaman-apprentice, line 28: Do we care if this was called, as long as the clipboard was written with the correct text?
```suggestion
			expect(writeToClipboardSpy).toBeCalledWith("Magic Monday")
```

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.ts
```diff
@@ -0,0 +1,15 @@
+import "./copyToClipboardButton.component.scss"
+import { Component, Inject } from "@angular/core"
+import { CopyToClipboardService } from "./copyToClipboard.service"
+
+@Component({
+	selector: "cc-copy-to-clipboard-button",
+	template: require("./copyToClipboardButton.component.html")
+})
+export class CopyToClipboardButtonComponent {
+	constructor(@Inject(CopyToClipboardService) private service: CopyToClipboardService) {}
```
shaman-apprentice, line 10: [Optional] rename service -> copyToClipboardService or copyToClipboard

shaman-apprentice (commented):
Only half way through so far

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	it("should return only header line if there is only one attribute without files", () => {
+		const result = buildTextOfFiles(WITH_ONE_FILELESS_ATTRIBUTE)
+
+		expect(result).toBe("FUNCTIONS\n")
+	})
+	it("should return valid string if there two attributes", () => {
+		const result = buildTextOfFiles(WITH_TWO_ATTRIBUTES)
+
+		expect(result).toBe(
+			`RLOC\n` + `\t${String.fromCodePoint(8226)} fileA (12)\n` + `COMMENTS\n` + `\t${String.fromCodePoint(8226)} fileA (14)\n`
```
shaman-apprentice, line 14: as far as I know node supports utf-8. Maybe also utf-16? Can we just copy paste the real char here?

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { buildTextOfFiles } from "./clipboardString"
+import { FileToValue } from "./getFilenamesWithHighestMetrics"
+
+describe("buildTextOfFiles", () => {
+	it("should return only header line if there is only one attribute without files", () => {
+		const result = buildTextOfFiles(WITH_ONE_FILELESS_ATTRIBUTE)
+
+		expect(result).toBe("FUNCTIONS\n")
```
shaman-apprentice, line 8: is this useful, or should we skip a header without entries?

shaman-apprentice: is this possible at all, or is this attribute then not included in cc.json anyway? I am not sure. Anyway, it is an not important edge case, so let's not spend too much time with it.

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
```diff
@@ -0,0 +1,21 @@
+import { FileToValue } from "./getFilenamesWithHighestMetrics"
+
+function getStringLineFromItem(item: FileToValue): string {
```
shaman-apprentice, line 3: [Optional / heavily opinionated] the type makes already sure, that it will return a string, so the string in the function name is a duplicate from my point of view
```suggestion
function getLineFromItem(item: FileToValue): string {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
```diff
@@ -0,0 +1,21 @@
...
+	return `\t${String.fromCodePoint(8226)} ${item.name} (${item.value})` + `\n`
```
shaman-apprentice, line 4: can we use the real code symbol directly?

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function getStringHeaderFromAttribute(title: string): string {
+	return `${title.toUpperCase()}\n`
+}
+
+export function buildTextOfFiles(attributeToFiles: Map<string, FileToValue[]>): string {
+	let clipboardText = ""
```
shaman-apprentice, line 12: [nitpick]
I think `clipboardText` is too specific as the function doesn't know it will be used in clipboard context later. Maybe just `text` or `result`.

[random info to be ignored]
I personally love a reduce here, which would reduce variable scopes. And in theory a reduce is faster. But as far as I know it is not yet in JS - why I think Ruben doesn't like reduce^^

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
+import { NodeType } from "../../../codeCharta.model"
+import { FileToValue, getFilenamesWithHighestMetrics, updateAttributeMap } from "./getFilenamesWithHighestMetrics"
+
+let MAP: Map<string, FileToValue[]>
```
shaman-apprentice, line 4: The parsing logic behaves kindly and doesn't mutate any passed in state, does it? So we can and should make it a const.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+				{ name: "file1", value: 15 },
+				{ name: "file1", value: 10 },
+				{ name: "file1", value: 5 }
+			]
+		]
+	])
+})
+
+describe("getFilenamesWithHighestMetrics", () => {
+	it("should give map which has exactly two keys (mcc, rloc)", () => {
```
shaman-apprentice, line 27:
```suggestion
	it("should return a map with keys for each attribute", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+		const resultMap = getFilenamesWithHighestMetrics(BIG_NODE_WITH_TWO_ATTRIBUTES)
+		const mccResults = resultMap.get("mcc")
+		const rlocResults = resultMap.get("rloc")
+
+		expect(resultMap.size).toBe(2)
+		expect(mccResults).toBeTruthy()
+		expect(rlocResults).toBeTruthy()
+	})
```
shaman-apprentice, line 35: [Optional] add new lines after each `it`-block

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+	it("should restrict to at most 10 files for each attribute", () => {
+		const resultMap = getFilenamesWithHighestMetrics(BIG_NODE_WITH_TWO_ATTRIBUTES)
+
+		for (const files of resultMap) {
+			expect(files.length).toBeLessThanOrEqual(10)
+		}
+	})
+	it("should return correct values for rloc and mcc", () => {
```
shaman-apprentice, line 43: We have some test duplication here: This test asserts everything from first test and more. Can we remove the first test?

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+	it("should ignore folders", () => {
+		const map = getFilenamesWithHighestMetrics(CONTAINS_FOLDER_WITH_ATTRIBUTES)
+
+		expect(map.get("rloc")).toBeUndefined()
+		expect(map.get("mcc")).toEqual([{ name: "file", value: 0 }])
+	})
+})
+
+describe("updateAttributeMap", () => {
+	it("should not add if value is too low", () => {
```
shaman-apprentice, line 71:
```suggestion
	it("should not add if value not one of the 10 highest", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+		updateAttributeMap("mcc", 0, "file with low value", MAP)
+
+		expect(MAP.get("mcc")).not.toContainEqual({ name: "file with low value", value: 0 })
+		expect(MAP.get("mcc").length).toBe(10)
+	})
+	it("should add if value is high", () => {
```
shaman-apprentice, line 77:
```suggestion
	it("should add if value is one of the 10 highest", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.md
```diff
@@ -0,0 +1,58 @@
+## More perfomant alternativ for insertion algorithm
```
shaman-apprentice, line 1: Please copy past this text into the PR or into the issue and delete this file. I am not confident this documentation will be useful in half a year within the codebase.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+	const keyIsNew = !map.has(key)
+	if (keyIsNew) {
+		map.set(key, [{ name: fileName, value }])
+		return
+	}
+
+	const newPair = { name: fileName, value }
+	const previousValues = map.get(key)
+
+	// NOTE: Perfomance can be improved by 10% to 25% by optimizing insertion algorithm for full arrays. See getFilenamesWithHighestMetrics.md-file for details
```
shaman-apprentice, line 30: There is a typo in "Perfomance". Anyway, see comment above. Or feel free to implement and extract the more performant way, if it was already implemented.

shaman-apprentice (changes requested):
A lot of nitpicks and small comments. But good work and cool features! :)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Added 🚀
+
+-   Copy-to-Clipboard-Button to paste the top 10 of files with highest metric values [#2942](https://github.com/MaibornWolff/codecharta/pull/2942)<br/>![image](https://user-images.githubusercontent.com/46388280/184089603-ecfa8e31-8241-42a2-9954-2de554347381.png)<br/>![image](https://user-images.githubusercontent.com/46388280/184089577-5cd2eec0-5293-4083-b629-0e3c5621047c.png)
```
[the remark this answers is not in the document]
knoffi: Yep, definitely shorter and and good to understand.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: "is one of the 10 highest" might be oudated soon.

What about "is highest value" and then adjust the implementation?

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: "not one of the 10 highest" might be outdated soon. How about "is the lowest value"?

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: Yes, thanks! We missed this clean-up when we changed the test input.

knoffi: Resolved with commit [5ea8361](https://github.com/MaibornWolff/codecharta/pull/2942/commits/5ea8361f314b3e7d048f68e80716bdbba4f96041)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
knoffi: Yes! Resolved with commit [89ab875](https://github.com/MaibornWolff/codecharta/pull/2942/commits/89ab875127f1d597acf125a17d0178e19e5096b9)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
knoffi: Totally! I love it!

knoffi: Resolved with commit [89ab875](https://github.com/MaibornWolff/codecharta/pull/2942/commits/89ab875127f1d597acf125a17d0178e19e5096b9)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.ts
knoffi: I agree with you.

knoffi: Resolved with commit [616f8d2](https://github.com/MaibornWolff/codecharta/pull/2942/commits/616f8d29759422faf9bfb7009480c795d902a03a)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
knoffi: Resolved with commit [89ab875](https://github.com/MaibornWolff/codecharta/pull/2942/commits/89ab875127f1d597acf125a17d0178e19e5096b9)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
knoffi: It is good to know that the function can handle this edge case without throwing an exception (fail safe).

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.ts
knoffi: Sry, but I do not get the benefit from this renaming. The name does not need to be the same as the type, this is what your code editor is for.
Of course, if there are several services in this component, then this would be a different situation...

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe("CopyToClipboardButtonComponent", () => {
+	let component: CopyToClipboardButtonComponent
+
+	beforeEach(() => {
+		const test = TestBed.configureTestingModule({
+			imports: [CopyToClipboardButtonModule]
+		})
+		component = test.createComponent<CopyToClipboardButtonComponent>(CopyToClipboardButtonComponent).componentInstance
+		component["service"].getClipboardText = () => "Magic Monday"
```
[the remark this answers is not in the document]
knoffi: Nice, thanks! : )

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboard.service.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { TestBed } from "@angular/core/testing"
+import { CopyToClipboardService } from "./copyToClipboard.service"
+import { buildTextOfFiles } from "./util/clipboardString"
+import { getFilenamesWithHighestMetrics } from "./util/getFilenamesWithHighestMetrics"
+let service: CopyToClipboardService
```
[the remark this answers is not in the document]
knoffi: Normally, the service declaration needs to be accessible from at least two describe blocks and thus, we would expect it right after the import statements.
This is why I went with this common pattern for readability, although it is not necessary for the scope logic.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboard.service.ts
```diff
@@ -0,0 +1,22 @@
...
+		const { unifiedMapNode } = accumulatedDataSelector(this.state.getValue())
+		return unifiedMapNode
+	}
+
+	getClipboardText(): string {
+		const node = this.getUnifiedMapNode()
+		const filesByAttribute = getFilenamesWithHighestMetrics(node)
+
+		return buildTextOfFiles(filesByAttribute)
+	}
```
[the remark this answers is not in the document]
knoffi: I do not think that your brain is small! xD

Personally, I prefer it this way. `getClipboardText` becomes a table of content and the function instantly reveals its structure and intentions:

1. Get a node (unified map node)
2. Get file names by attribute (highest metrics)
3. Build text
4. Return text

So in the end, there is less boiler code in `getClipboardText`, because we get every variable with one function call.
And whoever wants to analyze _how_ we get `unifiedMapNode` can dive into `getUnifiedMapNode`. There, you find the details and nothing but the details.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+		if (data.type === NodeType.FILE && data.attributes) {
+			for (const [key, value] of Object.entries(data.attributes)) {
+				updateAttributeMap(key, value, data.name, fileToValueByAttributes)
+			}
+		}
+	}
+	return fileToValueByAttributes
+}
+
+export function updateAttributeMap(key: string, value: number, fileName: string, map: Map<string, FileToValue[]>) {
```
knoffi, line 20: ## More perfomant alternativ for insertion algorithm

### Abstract

In `getFilenamesWithHighestMetric.ts` , we implemented the function `updateAttributeMap` which adds items to a `Map<string,FileToValue[]>` . This file shows how the perfomance can be improved.

### Alternative Algorithm

Instead of using the slightly costly

```
...
insertSorted(previousValues, newPair)
map.set(key, previousValues.slice(0, MAX_ENTRIES))
...
```

we could go for the following algorithm (pseudo-code):

```
IF ( previousValues.length < MAX_ENTRIES ) :
        insertSorted( previousValues, newPair )
ELSE
        insertInFullDescendingList( previousValues, newPair )
```

whereby `insertInFullDescendingList` could be implemented as

```
function insertInFullDescendingList<T extends { value: number }>(newItem: T, descendingList: T[]): void {
	let index = descendingList.length - 1
	while (index >= 0) {
		const oldItem = descendingList[index]

		if (oldItem.value < newItem.value) {
			const afterFirstLoop = index < descendingList.length - 1
			if (afterFirstLoop) {
				descendingList[index + 1] = oldItem
			}

			const atLastLoop = index === 0
			if (atLastLoop) {
				descendingList[index] = newItem
			}
		} else {
			const newValueTooSmall = index === descendingList.length - 1
			if (!newValueTooSmall) {
				descendingList[index + 1] = newItem
			}

			break
		}
		index--
	}
}
```

The perfomance would improve because we loop ONCE through the list and we IMMEDIATELY stop the iteration after `newItem` has found it's place. Also, we do not add `newItem` at all and we immediately stop the iteration, if the lowest value in the list is already bigger than `newItem.value` .

knoffi (commented):
Comment about perfomance (+ more perfomant alternative)

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.md
knoffi: Applied with commit [4045391](https://github.com/MaibornWolff/codecharta/pull/2942/commits/40453916f006e10ae233021e376bd307576a0f54)

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
knoffi: Resolved with commit [4045391](https://github.com/MaibornWolff/codecharta/pull/2942/commits/40453916f006e10ae233021e376bd307576a0f54)

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: Resolved with commit [448376b](https://github.com/MaibornWolff/codecharta/pull/2942/commits/448376bca7e0eadf17e15ddb771144acb9c59788)

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: Resolved with commit [caf8910](https://github.com/MaibornWolff/codecharta/pull/2942/commits/caf8910d32f50acb77cc810979421c23c6c2b353)

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: Actually, the method `updateAttributeMap` mutates `MAP` in its test suites.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
knoffi: In my opinion: yes. Because the method hast two jobs: 1. Get the clipboard text by calling the service method, 2. Write the result to clipboard.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+		//clipboard does not exist in jest's JSdom, see https://stackoverflow.com/questions/62351935/how-to-mock-navigator-clipboard-writetext-in-jest
```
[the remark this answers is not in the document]
knoffi: When we wrote the test, we did not know that clipboard is undefined in jests's JSdom.

This is why we wrote this comment, to document our findings. For you, it may be trivial, for us, it was something new. The next person who reads this test might be a newcomer to MaibornWolff and we thought that she/he might be thankful for this comment.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.scss
```diff
@@ -0,0 +1,3 @@
+cc-copy-to-clipboard-button {
```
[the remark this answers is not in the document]
knoffi: Good finding, thank you!

Resolved with commit [58c6583](https://github.com/MaibornWolff/codecharta/pull/2942/commits/58c65833ee7277d9d09e1a4c7769ad267cc49ec5)

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboard.service.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+jest.mock("./util/clipboardString", () => {
+	return { buildTextOfFiles: jest.fn().mockReturnValue("Magic Monday") }
+})
+jest.mock("./util/getFilenamesWithHighestMetrics", () => {
+	return { getFilenamesWithHighestMetrics: jest.fn() }
+})
+describe("copyToClipboardService", () => {
+	beforeEach(() => {
+		TestBed.configureTestingModule({ providers: [CopyToClipboardService] })
+		service = TestBed.inject(CopyToClipboardService)
```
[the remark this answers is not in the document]
knoffi: I applied it but everything got more ugly suddenly. This looks like a tradeoff.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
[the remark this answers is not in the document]
knoffi: What is your opinion about https://ng-mocks.sudo.eu/api/MockProvider#:~:text=A%20mock%20provider%20in%20Angular,token's%20value%20will%20be%20undefined%20. ? It looks promising to me and I likee their syntax. 

I would like to apply it if you give your okay.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboard.service.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+		jest.clearAllMocks()
+	})
+	it("should call text function and mapNode function on getClipboardText", () => {
```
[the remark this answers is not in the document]
Hall-Ma: I also agree with this.

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+			{ name: "file2", value: 84 },
+			{ name: "file3", value: 122 }
+		]
+	]
+])
+const WITH_TWO_ATTRIBUTES = new Map<string, FileToValue[]>([
+	["rloc", [{ name: "fileA", value: 12 }]],
+	["comments", [{ name: "fileA", value: 14 }]]
+])
+const WITH_ONE_FILELESS_ATTRIBUTE = new Map<string, FileToValue[]>([["functions", []]])
```
Hall-Ma, line 35: I prefer to have the constants at the beginning of the file. This would also preserve consistency between test files.

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+		const result = buildTextOfFiles(WITH_ONE_FILELESS_ATTRIBUTE)
+
+		expect(result).toBe("FUNCTIONS\n")
+	})
+	it("should return valid string if there two attributes", () => {
+		const result = buildTextOfFiles(WITH_TWO_ATTRIBUTES)
+
+		expect(result).toBe(`RLOC\n` + `\t• fileA (12)\n` + `COMMENTS\n` + `\t• fileA (14)\n`)
+	})
+	it("should return valid string if there is one attribute with many files", () => {
```
Hall-Ma, line 15: [Optional] add new lines after each `it`-block.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+describe("getFilenamesWithHighestMetrics", () => {
+	it("should return a map with keys for each attribute", () => {
+		const resultMap = getFilenamesWithHighestMetrics(BIG_NODE_WITH_TWO_ATTRIBUTES)
+		const mccResults = resultMap.get("mcc")
+		const rlocResults = resultMap.get("rloc")
+
+		expect(resultMap.size).toBe(2)
+		expect(mccResults).toBeTruthy()
+		expect(rlocResults).toBeTruthy()
```
Hall-Ma, line 34: I would also test the value length of the map.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
Hall-Ma: "Should add value if it is greater than the lowest value"

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+	})
+
+	it("should add if value is high", () => {
+		updateAttributeMap("mcc", 9001, "file with highest value", MAP)
+
+		expect(MAP.get("mcc")).toContainEqual({ name: "file with highest value", value: 9001 })
+		expect(MAP.get("mcc").length).toBe(10)
+	})
+
+	it("should add if map is empty", () => {
```
Hall-Ma, line 80:
```suggestion
	it("should add key-value pair if map is empty", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		const EMPTY_MAP = new Map<string, FileToValue[]>()
+
+		updateAttributeMap("mcc", 0, "first file", EMPTY_MAP)
+
+		expect(EMPTY_MAP.get("mcc")).toContainEqual({ name: "first file", value: 0 })
+		expect(EMPTY_MAP.get("mcc").length).toBe(1)
+	})
+
+	it("should add if key has less than 10 values", () => {
```
Hall-Ma, line 89:
```suggestion
	it("should add value if key has less than 10 values", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		const TINY_MAP = new Map<string, FileToValue[]>([["mcc", [{ name: "first file", value: 424_242 }]]])
+
+		updateAttributeMap("mcc", 0, "second file", TINY_MAP)
+
+		expect(TINY_MAP.get("mcc")).toContainEqual({ name: "second file", value: 0 })
+		expect(TINY_MAP.get("mcc").length).toBe(2)
+	})
+
+	it("should add if key is different", () => {
```
Hall-Ma, line 98:
```suggestion
	it("should add key-value pair if key is different", () => {
```

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		{ name: "leaf1", attributes: { ["mcc"]: 10, ["rloc"]: 5 }, type: NodeType.FILE }
+	]
+}
+
+const CONTAINS_FOLDER_WITH_ATTRIBUTES = {
+	name: "root_folder",
+	type: NodeType.FOLDER,
+	attributes: { mcc: 111, rloc: 15 },
+	children: [{ name: "file", type: NodeType.FILE, attributes: { mcc: 0 } }]
+}
```
Hall-Ma, line 136: Same opinion as already described above.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,40 @@
...
+	const keyIsNew = !map.has(key)
+	if (keyIsNew) {
```
Hall-Ma, line 22:
```suggestion
	if (!map.has(key)) {
```
I don't see any advantages to putting the boolean in an extra variable.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetrics.ts
```diff
@@ -0,0 +1,40 @@
...
+		map.set(key, [{ name: fileName, value }])
+		return
+	}
+
+	const newPair = { name: fileName, value }
```
Hall-Ma, line 27: Please put this line at the beginning of the function. You also can reuse this in line 23.

Hall-Ma (dismissed):

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboard.service.spec.ts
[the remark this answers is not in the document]
knoffi: For me, it sounds like both of you are fine with deleting this test, but you are not against keeping this test, did I get this right?

If we do not apply some kind of "convert data" and "build text from it", then there is certainly a bug or we are refactoring the code.

As a TDD user, I would benefit from these tests if I ever need to refactor this. I might also benefit from these tests as soon as we advance with the follow-up tickets.

As a non-TDD user, the only case I see where you would get bothered by this test is the case where we start to construct the string directly from the codeMapNode.
And honestly, this does not give us a relevant performance advantage, but it would certainly make the code less readable. And based on your previous opinions, you would advice against this anyway.

So from my point of view, I only see potential benefits and no relevant harm.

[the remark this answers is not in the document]
Hall-Ma: Whatever makes you happy :grin:

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.spec.ts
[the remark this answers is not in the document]
knoffi: There seems to a problem with some dependencies and thus I am very hesitant.

visualization/app/codeCharta/ui/copyToClipboardButton/copyToClipboardButton.component.html
```diff
@@ -0,0 +1,2 @@
+<cc-action-icon icon="fa fa-clipboard" title="Copy filenames with worst metric values to clipboard" (click)="copyNamesToClipBoard()">
```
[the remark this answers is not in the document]
knoffi: Resolved with commit [cc5c9ae](https://github.com/MaibornWolff/codecharta/pull/2942/commits/cc5c9aecf905573614ba4125fff7f602c3d78ee8)

visualization/app/codeCharta/ui/copyToClipboardButton/util/clipboardString.spec.ts
knoffi: Actually, it is different in some test files. For example, deltaGenerator.spec.ts .

In deltaGenerator, the reason was the following: When you enter a test file, then you do not want your eyes to immediately focus on the details of your mock objects. You do not care about these objects before you have seen the basic structure of the relevant test.

You want to see the tests. Otherwise, people start hiding their mock objects in mockData.ts again which Cedric wanted to prevent. 

### Mock objects vs config objects

But we all agree that constant config-objects have their place at the top of the file. This is, because we often search for them explicitly! Sometimes, a whole ticket basically says "change config property value". But this does not apply to test objects.

visualization/app/codeCharta/ui/copyToClipboardButton/util/getFilenamesWithHighestMetric.spec.ts
knoffi: Same opinion as already described above.

shaman-apprentice (approved):

Issue #2966, opened 2022-08-09
File extension bar doesn't highlight buildings on hover


shaman-apprentice:
# Bug

## Expected Behavior
Given any map
When I hover over a file type in the file extension bar
Then all buildings of this type become highlighted

## Actual Behavior
Nothing becomes highlighted

## Steps to Reproduce the Problem
1. Hover over a file type in the file extension bar

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes
- CodeCharta Version: 102.0

Refs: https://github.com/MaibornWolff/codecharta/pull/1806#pullrequestreview-596741316

PR #2968, opened 2022-08-09
Refactor/2318/migrate file extension bar

shaman-apprentice:
please note that highlighting is already broken before this migration. I will fix it with an own PR within #2966


Chrisp485 (approved):
LGTM 🎉

Issue #2978, opened 2022-08-11
Upload Release Files


jannikr:
# Bug

## Expected Behavior

GIVEN

Release pipeline script `release_gh_pages.yml

WHEN

Releasing

THEN

Release assets are being uploaded

## Actual Behavior

Socket hang up

## Screenshots
<img width="1252" alt="Screenshot 2022-08-11 at 09 56 44" src="https://user-images.githubusercontent.com/31436472/184088976-478dbb92-c6f6-4d41-81b6-4894baa2c931.png">

Possible solution: https://github.com/alexellis/upload-assets/issues/5

phanlezz:
Closed, reopen if there are still problems after the next release.

phanlezz:
Permissions unknown in github action

PR #2979, opened 2022-08-11
Add permission to upload-assets action #2978

jannikr:
# Add permission to upload-assets action

Issue: #2978


shaman-apprentice (dismissed):
Do we really need https://github.com/alexellis/upload-assets? I am not too happy about using a lib with only 31 stars. There must be a more standard way to do it. I would be fine with approving for now, when we have a follow up ticket.

ce-bo (approved):
LGTM

ce-bo:
> Do we really need https://github.com/alexellis/upload-assets? I am not too happy about using a lib with only 31 stars. There must be a more standard way to do it. I would be fine with approving for now, when we have a follow up ticket.

@shaman-apprentice We use this action already since a while. But it is not following SemVer. I like your suggestion to check, if we can use a stable library. But for now it is working and I would not invest that much time to change it.

PR #2980, opened 2022-08-11
Refactor/2318/migrate pre render service 2

### LINKED ISSUE #2977: loading indicator isn't working
# Bug

## Expected Behavior
Given a huge map
When I load it
Then a loading indicator is displayed on the map while it is loading

## Actual Behavior
The page becomes unresponsive until the map is loading

1. Load a huge map, which needs several seconds to load
2. Wait until it gets loaded and see that there was no loading indicator

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes 
- CodeCharta Version: 1.103.1

## Developer notes
Without checking I assume that:
- Parsing of file is probably done on main thread, which would block animation of loading give anyway.
- Rendering is probably done on main thread, which would block animation of loading give anyway.

## TODOs
- [ ] Check if we still need `isLoadingMap` and `isLoadingFile` within store/appSettings

shaman-apprentice:
- fix #2977 
- ref #2318 (migrate codeMap.preRender.service)
  - explicit state actions on which to render
  - throttle rendering (reducing total initial calls to render of default dev map from 58 to 1)
  - a very rough performance analysis through chrome's performance profile shows no real difference (uploading a 1.704kb and a 28.596kb file)
- delete code for re-selecting building after map change as this didn't work and wasn't missed


BridgeAR:
@shaman-apprentice is it intentional that this is still a draft? I guess so as there are still some open ToDos?

visualization/app/codeCharta/codeCharta.service.spec.ts
```diff
@@ -343,7 +351,9 @@ describe("codeChartaService", () => {
 				}
 			]
 
-			await codeChartaService.loadFiles([{ fileName, content: validFileContent, fileSize: 42 }])
+			expect(async () => codeChartaService.loadFiles([{ fileName, content: validFileContent, fileSize: 42 }])).rejects.toThrow(
+				"No files could be uploaded"
+			)
```
Hall-Ma, line 356: Why can't this file be loaded although it has a valid content?

visualization/app/codeCharta/state/effects/renderCodeMapEffect/renderCodeMap.effect.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	private actionsRequiringRender$ = this.actions$.pipe(
+		filter(action => actionsRequiringRerender.some(actions => isActionOfType(action.type, actions)))
+	)
+
+	renderCodeMap$ = createEffect(
+		() =>
+			combineLatest([this.store.select(accumulatedDataSelector), this.actionsRequiringRender$]).pipe(
+				filter(([accumulatedData]) => Boolean(accumulatedData.unifiedMapNode)),
+				throttleTime(maxFPS, asyncScheduler, { leading: false, trailing: true }),
```
Hall-Ma, line 35: Would you be so kind to explain how `throttleTime` really works? I read the documentation, but it's not so easy to fully understand it. How long does this function wait before continuing?

Hall-Ma (approved):
I haven't noticed any different behaviour than with the non-migrated components. So I give you a go! :smiley:

visualization/app/codeCharta/state/effects/renderCodeMapEffect/renderCodeMap.effect.ts
shaman-apprentice: throttleTime  ensures that the function is not called more often than given time, but when called it will be called after given time once.

In contrast debounce does wait given time after no further calls appeared. Feel free to ask on Friday again in our session :)

visualization/app/codeCharta/codeCharta.service.spec.ts
shaman-apprentice: see the test description:

> it("should show a validation error if two files in a folder have the same name"

which is set up above:
```ts
validFileContent.nodes[0].children[1].children[0].name = "duplicate"
validFileContent.nodes[0].children[1].children[1].name = "duplicate"
```

Issue #2953, opened 2022-08-05
Parsing fails on apple-silicon/arm64, wind and windows


MariusMaibornWolff:
# Bug

gitlogparser, rawtextparser and sourcecodeparser fail on Apple Silicon.

The command gets stuck on apple silicon with no further output.
No output file is written after executing the command.
e.g.:

ccsh gitlogparser "git_output" "--output-file=gitlogparserresult.json" "--file-name-list=git_filename_list_output" "--not-compressed" "--silent=false" "--add-author=true"

output:

 Assumed encoding UTF-8
  57% [=========================================================>                                           ]  9868725/17018092 Byte, ETA: 00:00:00
Unhandled Edge Case in MergeCommit: deleted: false, mutated: true, modification type: RENAME, initalAdd: false, file: apps/community-service/.idea/runConfigurations.xml
  58% [==========================================================>                                          ]  9881689/17018092 Byte, ETA: 00:00:00
Unhandled Edge Case in MergeCommit: deleted: false, mutated: true, modification type: RENAME, initalAdd: false, file: apps/community-service/.idea/runConfigurations.xml
 100% [====================================================================================================>] 17018092/17018092 Byte, ETA: 00:00:00
Created Project with 3325 leaves.

## Expected Behavior

Process finishes and output json file has been created.

## Actual Behavior

Process gets stuck. No further output/error is given.

## Steps to Reproduce the Problem
1. User a mac with apple-silicon / arm64
2. Run any parser on any project
3. Observe that the process is stuck and no further output is given.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes | no
- CodeCharta Version: latest
- OS: Mac OS 
- Browser: Chrome

MariusMaibornWolff:
For solution: Eventually check if parsers run in a docker container successfully.

ce-bo:
@Hall-Ma found out, that this applies also for Windows (tested on Windows with ccsh version 1.103.3).

ce-bo:
Let's extend `golden_test.sh` to have an automated regression test in case this bug reoccurs sometime.

Hall-Ma:
When you are trying to use Sonar Importer on Windows (Git Bash) you get no output file because the script doesn't end:
[...]

Issue #2858, opened 2022-06-21
Parser always waits 1000ms for STDIN


luelista:
# Bug / Technical Debt

Currently, there is a weird behaviour implemented in many parsers: Even if files are specified on the command line, the parser still waits 1000ms for potential data on STDIN. 

We should use the common practise in UNIX applications (e.g. [cat](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/cat.html)):

* If no files are specified, read from STDIN
* If files are specified, don't read from STDIN
   * except if one of the file names is "-", then read from STDIN in the place of this file name

## Expected Behavior

GIVEN a Parser is called with file(s) as arguments  
THEN no data from STDIN is expected, and we don't wait any amount of time whether some might appear.

GIVEN a Parser is called with no file arguments  
THEN we read data from STDIN, without any timeout

GIVEN a Parser is called with file arguments, one of which being "-" (a single minus sign)  
THEN we read data from the files as well as from STDIN

## Actual Behavior

Given a Parser is called with files as argument

WHEN no data is provided on STDIN

THEN it waits 1000ms if some data might still come

## Related

This problem was found when creating a workaround for issue #2850 in PR #2854

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes | no
- CodeCharta Version:
- OS:
- Browser:

PR #2987, opened 2022-08-15
Fix stuck merge process by correcting input stream validation Issue #2953 

phanlezz:
# Fix for importers who were stuck on merge process

This fixes a typo that caused the parsing to fail on nearly every OS, on different terminals, depending on, if they send an EOF token or not. 
The typo was introduced during the compatibility conversion to java8: #2930
This available check is required, because of a workaround described here #2854 

Root of the problem is that we always wait for stdin: #2858 

Issue: #2953

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Fixed 🐞
+
+-   Fix bug in CompressionStreamHandler's input stream validation, that caused endless blocking reads from System.In [#2987](https://github.com/MaibornWolff/codecharta/pull/2987)
+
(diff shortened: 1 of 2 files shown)


ce-bo:
This Bug should have broken out Pipeline. Would you please Check golden_test.sh and extend the script to have a regression test? Thank you!

phanlezz:
> This Bug should have broken out Pipeline. Would you please Check golden_test.sh and extend the script to have a regression test? Thank you!

I think we should merge this for now as-is. It is difficult to build tests that simulate an open input stream on a command line. 
We need to fix our behavior for input files. An issue is already open for this.

Hall-Ma (approved):

BridgeAR:
> I think we should merge this for now as-is. It is difficult to build tests that simulate an open input stream on a command line.
We need to fix our behavior for input files. An issue is already open for this.

Merging this as is means the test is going to be deprioritized and that seems quite urgent.
We can execute the command and check that it's successful and the output file is generated. That would be a sufficient test. The file would have to be created in a temp file with a random name and should be deleted after the test.

PR #2988, opened 2022-08-18
Update the golden_test.sh to cover all used modules #2921

### LINKED ISSUE #2921: golden_test.sh needs to be updated
# Bug
importers tested in golden_test.sh that aren't used in production mode:
- crocosomo
- jasome

importers not tested in golden_test.sh that are used in production mode:
- edgefilter
- csvimport
- sonarimport
- gitlog
- csvexport
- codemaat
- rawtext
- metric gardener

## Expected Behavior
all importers that are used in production should be tested in golden_test.sh.
importers that aren't used right now shouldn't be tested
<GIVEN> 

WHEN

THEN

1.
1.
1.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes | no
- CodeCharta Version:
- OS:
- Browser:

phanlezz:
# Update the golden_test.sh to cover all used modules 

`golden_test.sh` is part of the analysis gradle task `integrationTest` 

This PR will update existing, add new, and refactor, tests in `golden_test.sh` 
If a parser/importer is no longer used in production, but still in the code base, the test might get updated or at least verified as working as intended.

This will also cover the test case, in which a ccsh process would not finish and  would be stuck on the last step before a file output (see #2987).
This problem occurred because we (at this moment) always try to merge piped in projects from System.In, and in the situation when there is nothing incoming, the read would block endlessly. The test case will control, if System.In gets validated before reading.
This behavior of _always reading from System.In_ might change in the future (see #2858) .

Closes #2921


phanlezz:
Some tests might need adjustment before we merge this, if we change the output file handling. 
Currently the tests also check if extension get corrected 
- (.json to .cc.json) 
and if they are added/ignored correctly 
- (name -> name.cc.json.gz or name.cc.json ; name.cc.json -> name.cc.json ; name.cc.json.gz -> name.cc.json.gz)

Further updates in #2989 will be considered in new commits.

phanlezz:
In theory this is ready for review, but:
There is still a bug in OutputNameHandling which causes compressed files to not be named correctly, and this behavior is controlled by a test in this PR. So this PR can't be merged otherwise our build pipeline would fail everywhere.

luelista (approved):

PR #2989, opened 2022-08-19
Fix OutputFileHandler not ignoring a correctly named gz filename

MW-Friedrich:
# Fix OutputFileHandler not ignoring a correctly named gz filename

Issue: #2800 

## Description

Addendum to PR #2914 to fix a bug that was discovered after the merge :(

### DIFF
--- a/analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
+++ b/analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
@@ -19,16 +19,20 @@ object OutputFileHandler {
     }
 
     fun checkAndFixFileExtension(outputName: String): String {
-        if (outputName.endsWith("cc.json")) {
+        if (outputName.endsWith("cc.json") || outputName.endsWith(".cc.json.gz")) {
             return outputName
         }
         val sb = StringBuilder()
+        appendFilePathIfSpecified(sb, outputName)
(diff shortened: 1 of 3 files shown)


knoffi (approved):
LGTM! 🚀

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/serialization/OutputFileHandler.kt
```diff
@@ -19,16 +19,20 @@ object OutputFileHandler {
     }
 
     fun checkAndFixFileExtension(outputName: String): String {
-        if (outputName.endsWith("cc.json")) {
+        if (outputName.endsWith("cc.json") || outputName.endsWith(".cc.json.gz")) {
```
BridgeAR, line 22: I would just use a regular expression for cases like these. That way it's a single check with  optional parts. I would also handle the `.cc` optional (and our logic here is somewhat flawed as it would match files like `abcc.json`. I would accept it as it's a json file but it's just inconsistent right now.

MW-Friedrich: I think I remember @ce-bo saying that we should only consider .cc.json as valid, but thanks a lot for the hint with the regEx!
@phanlezz also noticed that we append .gz again, even if the user already specified it (so test.cc.json.gz becomes test.cc.json.gz.gz)
Should we/I open a new ticket for this or continue this here?

Issue #2992, opened 2022-08-22
Display path name in clipboard top 10


knoffi:
### Reason for hot fix

Ruben requested a different text display in #2744, but because of workflow issues we overlooked it. The branch of #2744 got merged now.

To satisfy our users and to support their work during audits, I will adjust the form of the clipboard text asap during this ticket.

### Request text form

See example:

RLOC
    • path/sample2LeafMergedIn.kt (600)
    • path/sample1OnlyLeaf.scss (400)
    • path/one/bigLeaf.ts (120)
    • path/two/bigLeaf.ts (100)
    • path/one/otherSmallLeaf.ts (70)
    • path/two/otherSmallLeaf.ts (70)
    • path/one/smallLeaf.html (30)
    • path/two/smallLeaf.html (30)

PR #2995, opened 2022-08-24
change clipboard to show file path + no name #2992

knoffi:
# Display file path instead of file name in clipboard text

Issue: #2992 

## Description

For the audits, Ruben prefers the complete path name in the clipboard text instead of the file name. (See comment in Zenhub.)


jannikr (approved):
LGTM 👍

Issue #2962, opened 2022-08-08
Suspicious Metrics Overlay is empty for maps without known metrics


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN
loading a map with a set of metrics that are unknown/not listed in artificialIntelligence.metricThresholds.ts

THEN
the Suspicious Metric Menu should show the message "The given metrics are not covered by this feature, so no finding could be made".

## Actual Behavior
A small and empty Menu is loaded.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes
- CodeCharta Version: 1.101.0
- OS: all
- Browser: all

ce-bo:
Closed by #2996.

PR #2996, opened 2022-08-25
Add untracked metrics to suspiciousMetricsHelper.ts #2962

jannikr:
# Suspicious Metrics Overlay is empty for maps without known metrics

Issue: #2962 

Additionally I suggest adding untracked metrics to the list even though the overlay appears right. This could be useful to inform the user about metrics that could not be tested. For example every metric might be unsuspicious but what about the other metrics that have not been tested. This feature allows the user to be warned about "untracked" metrics.

## Next steps:

- [x] Fix material tooltip (hovering does not work and alignment is not right yet)
- [x] Improve code readability
- [x] Write tests
- [x] Add CHANGELOG.md entry


visualization/app/codeCharta/ui/artificialIntelligence/artificialIntelligenceHelper/suspiciousMetricsHelper.ts
```diff
@@ -105,5 +113,8 @@ export function findGoodAndBadMetrics(
 		}
 	}
 
+	const untrackedMetrics = availableMapMetrics.filter(metric => !assessmentMetricList.includes(metric))
+	Array.prototype.push.apply(metricAssessmentResults.untrackedMetrics, untrackedMetrics)
```
Hall-Ma, line 117:
```suggestion
	metricAssessmentResults.untrackedMetrics.push(...untrackedMetrics)
```
Just a suggestion :)

Hall-Ma (approved):
I have another idea to this issue. All you need is this:
```
	for (const key in metricValuesByLanguages[mainProgrammingLanguage]) {
		const keys = Object.keys(languageSpecificMetricThresholds)
		if (!keys.includes(key) && !metricAssessmentResults.untrackedMetrics.includes(key)) {
			metricAssessmentResults.untrackedMetrics.push(key)
		}
	}
```
So we wouldn't need that array in array thing, that is really confusing me :D
But your approch is also good!

jannikr:
> I have another idea to this issue. All you need is this:
> 
> ```
> 	for (const key in metricValuesByLanguages[mainProgrammingLanguage]) {
> 		const keys = Object.keys(languageSpecificMetricThresholds)
> 		if (!keys.includes(key) && !metricAssessmentResults.untrackedMetrics.includes(key)) {
> 			metricAssessmentResults.untrackedMetrics.push(key)
> 		}
> 	}
> ```
> 
> So we wouldn't need that array in array thing, that is really confusing me :D But your approch is also good!

I agree with you and find your solution much more feasible!

Issue #3011, opened 2022-08-31
Images broken in How-to page for analysis with SonarQube


Hall-Ma:
# Bug

Some images are not embedded correctly on the GitHub (Docs) page 'Detailed instruction of how to use SonarQube 9.4'.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.104.1
- OS: Windows
- Browser: Chrome

PR #3012, opened 2022-08-31
Fix/3011/broken images in docs

Hall-Ma:
# Fix broken images on How-to article: 'Detailed instruction of how to use SonarQube 9.4'

Closes: #3011


MW-Friedrich (approved):
LGTM!
I have a feeling though that this is one of thise changes we can only see, once deployed, yes?

Hall-Ma:
> LGTM! I have a feeling though that this is one of thise changes we can only see, once deployed, yes?

Correct. But this should work. I copy and paste the file name :D

PR #3015, opened 2022-08-31
Feature/2619/run metric gardener automatically

MW-Friedrich:
# Run metric-gardener automatically, if the user does not have a mg.json

Issue: #2619 

## Description
After some questions came up, during PRs #2875 and #2909 develepment needed to be halted on this feature.
Now Metric-Gardener is to be shipped with CodeCharta and executed, when needed.


luelista:
Current version works apart from the fact we don't have metric-gardener installed for the tests, therefore the pipeline fails.

luelista (approved):

Issue #2789, opened 2022-04-28
Delta Mode not working for codecharta.cc.json


ce-bo:
# Bug

## Expected Behavior

GIVEN
- current codecharta.cc.json
- old map: Unzip and upload: [codecharta_2022-04-28_08-53.cc.zip](https://github.com/MaibornWolff/codecharta/files/8581208/codecharta_2022-04-28_08-53.cc.zip)

WHEN
comparing them in the delta mode

THEN
I can see the actual deltas.

## Actual Behavior
the two maps are shown side by side and no deltas are shown

## Screenshots
![image](https://user-images.githubusercontent.com/26900540/165716677-68a2fe25-6032-47b7-a350-22b8b909dd3c.png)

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes
- CodeCharta Version: 1.95.2
- OS: all
- Browser: all

knoffi:
That's not a bug, that's a feature! Or at least it is a feature which we can improve.

The problem seems to be the following:

The built-in `codecharta.json` has a root folder with name "root". And "root" has one child with name "app".

The uploaded cc.json-file has a root folder with a __different__ name: "root/app".

Our frontend compares the name of both root files and only sees `"root" !== "root/app"`.

I see two potential fixes here:

### The user-friendly one

When we compare two file with different root names, then we should inform the user about this. A small pop-up message containing
"The root folders ("root" vs "root/app") are different. These projects are not comparable!" would do the job.

Users would also benefit from it if they accidentally compare two projects with different root folders. This would spare them the hazzle when they overlook the obvious. Thus it would boost the UX of our project. 

Also, in the future, we could give the pop-up message a small button "Adjust root" which makes the frontend act like both root folders had the same name.


### The hacky one

We can also extend our logic so that the frontend handles this __specific__ case:
Instead of comparing the names of the root folders, we simply compare their "collapsed" names.

I.e., If "root" just contains the folder "app" which just contains the folder "src", while "src" has more than one child, then the "collapsed" root name would be:
"root/app/src"

knoffi:
## UPDATE

I tried the above appraoch "The hacky one". And it seems to work so far.

Next, I will try if it works with two small example files which contain edge metrics.

I am a little bit worried that if we enable DeltaMode for the above scenario, that another part of the logic will run into an unexspected case and thus break.

[...]

PR #3016, opened 2022-08-31
Fix/2789/delta mode with codecharta

Hall-Ma:
# Activate delta mode when roots are the same but have different paths

Closes: #2789


MW-Friedrich (commented):
I've had the chance to talk with @knoffi about the changes! The new code looks great and is probably more in line with what we need (YAGNI and all), but we still had a few, small pointers.

- Rename compareCollapsedRoot to better fit the function name
- Rename haveSameRoots, so it's clearer that only the name is compared
- Think of what should happen if two files can not be compared (doe to a malformed/weird cc.json) Maybe we can/should display some sort of toast or notification for the user?

But it's already a definite LGTM from us both!

Hall-Ma:
I'm not really happy with this solution. It's working but I don't like that I also have to handle this root thing in codecharta service (`unsubscribeReferenceFileSubscription`). Unfurtanetly this service isn't migrated yet, perhaps we can deal with it later in a separate selector.
Another solution I can see is that we could create a root folder (first slash) and push everything as a child (after the second slash) so that we could gain the 'original' or rather same root sequence as the compared file. But is it worth the effort?

ce-bo:
This issue will not be fixed. This bug is produced by comparing maps with different structures. So it is fine, that the two compared maps are not shown in delta mode, even if they are very similar. It would take to much effort to refactor the code parts. So I close this for now.

Issue #2749, opened 2022-03-25
SCMLogParserV2/GitLogParser - Bug during Git parsing


ce-bo:
# Bug

An exception occurs during parsing:
![image](https://user-images.githubusercontent.com/26900540/183442940-b5bb8cbc-cc63-44cf-a809-6367b6d85cb4.png)

Please contact me to get access to the used  git logs.

MW-Friedrich:
@ce-bo The error was in the anongit script, not in the parser itself. In your specific case the gitlog would need to be regenerated unfortunately (in case this is still relevant)

PR #3030, opened 2022-09-07
refactor anongit-script to avoid unintended side-effects

MW-Friedrich:
# GitLog-Parser does not work, due to a possible error in the anongit-script

Issue: #2749

## Description

When crating a git.log file with our anongit script, the wrong type of whitespace (space instead of tabs) seem to be used

## Screenshots or gifs

log-problem


phanlezz (approved):
LGTM 🚀

PR #3049, opened 2022-09-24
Refactor/2318/migrate code map component

shaman-apprentice:
- migrate codeMap.component to Angular
- init map initially only once and not twice as before
- remove window event listener on threeViewerService.destroy, so that they don't stack up on sharpness mode changes


Hall-Ma (approved):
LGTM!

PR #3050, opened 2022-09-24
Refactor/2318/remove three update cycle service

shaman-apprentice:
**Note that this PR is build upon #3049. So please review #3049 one first**

remove ThreeUpdateCycleService and call render directly instead

ref #2318


Hall-Ma (approved):
Keep it up!

PR #3051, opened 2022-09-24
Refactor/2318/clean up obsolete angularjs services

shaman-apprentice:
- migrate metricData.service
- migrate edgeMetric.service
- migrate nodeMetricData.service
- migrate edgeMetricData.service

Todo: 
- [x] add changelog
- [x] merge #3050 first


Hall-Ma (changes requested):
As already discussed, the selected edge metric of a map is not reset after selecting a new map (which has no edge metrics) and returning to the first selected map.

shaman-apprentice:
> As already discussed, the selected edge metric of a map is not reset after selecting a new map (which has no edge metrics) and returning to the first selected map. Apart from that, everything is great!

thx, very good catch! FIxed with https://github.com/MaibornWolff/codecharta/pull/3051/commits/c8d952ff149ce5c04f6ba6808772245eab1cbbe2

Hall-Ma (approved):
Super :D

PR #3055, opened 2022-09-27
Fix race condition due to missing await in a test

MW-Friedrich:
# Fix race condition due to missing await in a test

A user event did not have an await, causing a test to fail *sometimes* due to a race condition.


Hall-Ma (approved):
👍

Issue #2946, opened 2022-08-05
Create CodaCharta Docker container


jannikr:
# Feature request

Within a Docker container, both the frontend and the backend are to be loaded with all the necessary dependencies to run the project on different platforms. Furthermore, it would be practical to start SonarQube server with the necessary properties to cover as many programming languages as possible.

## Description

As a user, I want to quickly import and parse source code on different platforms so that I can upload it on the CodeCharta website without any dependency issues.

## Acceptance criteria
- Create docker container
- Add task to install npm, java, ...
- Add task to install CodeCharta Analysis and Visualization
- Add task for SonarQube Server with specific properties like default user and password.
- Add Documentation for a basic workflow in order to parse source code and upload it on CodeCharta

## Task Breakdown

- [x] Fix Visualisation container
   - [x] CC-Analysis does not get installed?
   - [x] Tag "latest" does not get correctly applied (https://user-images.githubusercontent.com/108457762/192304553-dd74b80e-de7b-4860-9f84-e742696c02d6.png)
- [x] Create new analysis container
   - [x] Add Git
   - [x] Add Tokei
   - [x] Add Sonar (Via docker-compose)
   - [x] Add Code Maat
   - [x] Add Metric Gardener
- [x] Write cool docker-compose for the new containers
- [x] Adjust build/release process
- [ ] Update or write new docs if necessary

MW-Friedrich:
@ce-bo SourceMonitor seems to be only available on Windows. (https://www.derpaul.net/SourceMonitor) 
While Windows emulation on Linux is possible with wine and Proton, it would be quite a huge undertaking. Is it okay if we don't include it in the container?

ce-bo:
@MW-Friedrich that‘s fine. The most important thing is to have SonarQube running. By that, we already can analyse a lot of programming languages. Additionally, the User will be able to import CSV files from SourceMonitor by our importers.

PR #3057, opened 2022-09-27
Dockerize CodeCharta

MW-Friedrich:
# Dockerize CodeCharta

Issue: #2946

## Description

To be as portable and convenient as possible, CodeCharta should ship as a Docker container too. To achieve this, this PR handles the creation of a Docker container for the analysis part of the software, where all necessary programs we can import data from are already installed, as well as creating a docker-compose file to easily orchestrate a full CodeCharta installation.


docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+        ports:
+            - "9001:80"
+
+    analysis:
+        image: codecharta/codecharta-analysis
+        volumes:
+            - shared-volume:/mnt/data
+
+volumes:
+    shared-volume: {}
```
ce-bo, line 21: What is the path that is mounted by the volume `shared-volume`?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Added 🚀
+
+-   Add new Analysis docker container, together with a docker-compose file to quickly get started [#3057](https://github.com/MaibornWolff/codecharta/pull/3057)
```
ce-bo, line 12: Would you please mention that the docker-compose is also providing a SonarQube instance?

docker-compose.yml
MW-Friedrich: shared-volume is an internal docker volume, currently without being mounted "from the outside". You can always copy files to and from it using docker cp. The volume is there so you can easily move files between the containers, like code files to and from sonar.

Issue #2507, opened 2021-11-15
Add "code quality default metric behavior"


BridgeAR:
# Feature request

## Description

As a user, I want to change as little as possible so that I have to do less work to achieve my goal.

It is often difficult to understand a map if all three metrics show different things. Mostly people want to change the color and height together and keep the size metric fixed as rloc (with a fallback to loc).

## Acceptance criteria

- Add a button (checkbox) that indicates the "program code" default behavior. In case it's set "bind" the height and color metric together and always use the preselected size metric like `rloc` or `loc` (done by #1532) for the size metric. 

## Open questions

- Define where to place the checkbox. Any suggestions are appreciated.

Hall-Ma:
@BridgeAR Which behavior should be the standard now? The linked or unliked behavior? 
Also, would you like (when the coupling is enabled) the color metric chooser to be disabled in the panel so that the user cannot select a color metric until it is unlocked again?

BridgeAR:
> Which behavior should be the standard now? The linked or unliked behavior?

I suggest we start by implementing the feature and keeping it disabled for now. Let's get some user feedback about it after a while to check if we want to enable it by default. It should also be disabled for cases where `rloc` and `loc` can not be found.

> would you like (when the coupling is enabled) the color metric chooser to be disabled in the panel so that the user cannot select a color metric until it is unlocked again?

Yes, that would be ideal.

PR #3058, opened 2022-09-28
Feature/2507/Link Height And Color Metric As New Default

Hall-Ma:
# Link Height And Color Metric As New Default

Closes: #2507

## Description

- adds a new button that allows to couple height and color metric. 
- when this option is enabled then you can only change the height metric and the color metric changes automatically to the value of the height metric

### By default this feature is not enabled:
image
### This is what it looks like when height and color metrics are linked:
image


visualization/app/codeCharta/state/effects/coupleHeightAndColorMetric/coupleHeightAndColorMetric.effect.ts
```diff
@@ -0,0 +1,26 @@
...
+		combineLatest([
+			this.store.select(heightMetricSelector),
+			this.store.select(isHeightAndColorMetricCoupledSelector),
+			this.store.select(colorMetricSelector)
+		]).pipe(
+			map(([heightMetric, isCouplingEnabled, colorMetric]) => {
+				return isCouplingEnabled ? setColorMetric(heightMetric) : setColorMetric(colorMetric)
+			})
+		)
+	)
```
Hall-Ma, line 25: @shaman-apprentice Do you have a suggestion on how can I approve this effect? It also dispatches the color metric when the coupling is false, but it's not necessary (I don't know any better). It should only dispatch the color metric when the coupling is enabled and the height metric is also changed. So maybe I need a kind of filter when I combine these selectors?

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.ts
```diff
@@ -14,14 +15,19 @@ type PanelSelection = "NONE" | "AREA_PANEL_OPEN" | "HEIGHT_PANEL_OPEN" | "COLOR_
 })
 export class RibbonBarComponent implements OnInit, OnDestroy {
 	panelSelection: PanelSelection = "NONE"
+	RLOC_METRIC = "rloc"
+	LOC_METRIC = "loc"
 	experimentalFeaturesEnabled$ = this.store.select(experimentalFeaturesEnabledSelector)
 	isDeltaState$ = this.store.select(isDeltaStateSelector)
 	hasEdgeMetric$ = this.store.select(edgeMetricDataSelector).pipe(map(edgeMetricData => edgeMetricData.length > 0))
-
+	isMetricCouplingHidden = false
```
BridgeAR: I personally would not hide it, it should just not be active by default. The user might still want to chain the metrics together, even though the used metrics are different from the ones we normally use.

visualization/app/codeCharta/util/customConfigBuilder.ts
```diff
@@ -72,6 +72,7 @@ function initializeAppSettings(target: CustomConfig) {
 		layoutAlgorithm: undefined,
 		maxTreeMapFiles: 0,
 		sharpnessMode: undefined,
```
ce-bo, line 74: Please make sure, that old custom configs without the new property can still be loaded and importet.

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.html
```diff
@@ -3,6 +3,7 @@
 	[value]="selectedMetricName"
 	(selectionChange)="handleMetricChanged($event.value)"
 	(openedChange)="handleOpenedChanged($event)"
+	[disabled]="disabled"
```
ce-bo, line 6: Are all three metric chooser disabled by this?

ce-bo (commented):
I really like the icon you have chosen for this! I will Test the new feature tomorrow. :-)

visualization/app/codeCharta/util/customConfigBuilder.ts
Hall-Ma: I will test it!

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.html
Hall-Ma: No! Only the metric chooser can be disabled because it get's an input property. The other choosers don't get this specific property and the default is false. Maybe I call this property 'isDisabled'.

visualization/app/codeCharta/util/customConfigBuilder.ts
Hall-Ma: It works :)

ce-bo:
Here are some remarks:
- If I link the metrics, then (on hovering the link button) the title should show "unlink metrics" instead of "link metrics".
- I would like to add a title attribute for the color metric chooser, so that the user is going to be informed that he/she should rather select the height metric instead because it is coupled/linked to that.
- The UI looks a bit random, especially if the edge metrics are shown. So maybe we can improve it to make it more intuitive. Do you have a suggestion for that?
image.

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.html
ce-bo: Ah I got it. I was counfused because the variable "disabled" has the same name as the native HTML Attribute "disabled". So I thought this is a static value, that disables all metricChoosers by default.

ce-bo:
I tryed something. Maybe this is one direction to express that height and color metric might be coupled together. But I am not happy with brush icon next to the chain icon:
image

ce-bo:
Another iteration: Please notice that the border is also underneath the height/color icon.
image

ce-bo:
> > I tryed something. Maybe this is one direction to express that height and color metric might be coupled together. But I am not happy with brush icon next to the chain icon: image
> 
> I like this one even more. I tried to adapt it, but with CSS it's quite difficult. 😥 But what do you think? image image

That looks good (although the "Link Height and Color metric" text is wrong)! I would like to extend the "border-bottom" from the select box to cover the brush / height icon as well. What do you think? For that I would remove the border-bottom from the select box and add it one level higher to the section or to the wrapping HTML element.

visualization/app/codeCharta/ui/ribbonBar/colorMetricChooser/colorMetricChooser.component.spec.ts
```diff
@@ -24,14 +25,24 @@ describe("colorMetricChooserComponent", () => {
...
 		await render(ColorMetricChooserComponent, { excludeComponentDeclaration: true })
 
+		expect(screen.getByRole("combobox").getAttribute("aria-disabled")).toBe("false")
+
 		await userEvent.click(await screen.findByText("aMetric (1)"))
 		expect(screen.getByText("Color Metric (highest value)")).not.toBe(null)
 		const options = screen.queryAllByRole("option")
-		expect(options[0].textContent).toMatch("aMetric (1)")
-		expect(options[1].textContent).toMatch("bMetric (2)")
+		await expect(options[0].textContent).toMatch("aMetric (1)")
```
shaman-apprentice, line 33: this `await` is not necessary, as [Jest's `toMatch`](https://jestjs.io/docs/26.x/expect#tomatchregexp--string) expectation returns not a Promise. 

The IDE gets confused, as `expect` is set globally and therefore it doesn't know if the `expect` comes from jest (our unit tests) or jest/expect-puppeteer (our e2e tests). 
image

Do you mind to create a new bug ticket for this falsely IDE hints?

visualization/app/codeCharta/ui/ribbonBar/linkHeightAndColorMetricButton/linkHeightAndColorMetricButton.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+		TestBed.configureTestingModule({
+			imports: [CommonModule]
+		})
+	})
+
+	it("should toggle height and color metric link on click and show associated icon", async () => {
+		const { container } = await render(LinkHeightAndColorMetricButtonComponent)
+		const isHeightAndColorMetricLinked = isHeightAndColorMetricLinkedSelector(Store.store.getState())
+
+		expect(container.getElementsByClassName("fa fa-chain-broken")[0].className).toEqual("fa fa-chain-broken")
```
shaman-apprentice, line 21: I find `expect(container.querySelector(".fa.fa-chain-broken")).not.toBe(null)` more intuitive. In case the element is not there `container.getElementsByClassName("fa fa-chain-broken")` would be an empty array and the test would break with a  null pointer exception and not fail with an assertion.

visualization/app/codeCharta/ui/ribbonBar/linkHeightAndColorMetricButton/linkHeightAndColorMetricButton.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+		expect(container.getElementsByClassName("fa fa-link")[0]).toBeUndefined()
+		expect(screen.queryByText("Unlink Height And Color Metric")).toBe(null)
+
+		await userEvent.click(screen.getByTitle("Link Height And Color Metric"))
+
+		expect(container.getElementsByClassName("fa fa-link")[0].className).toEqual("fa fa-link")
+		expect(container.getElementsByClassName("fa fa-chain-broken")[0]).toBeUndefined()
+		expect(screen.getByTitle("Unlink Height And Color Metric")).not.toBe(null)
+		expect(screen.queryByText("Link Height And Color Metric")).toBe(null)
+		expect(isHeightAndColorMetricLinkedSelector(Store.store.getState())).toBe(!isHeightAndColorMetricLinked)
```
shaman-apprentice, line 31: [Optional / opinionated] I see the store as an implementation detail and would therefore skip this assertion

visualization/app/codeCharta/state/store/appSettings/isHeightAndColorMetricLinked/isHeightAndColorMetricLinked.reducer.ts
```diff
@@ -0,0 +1,14 @@
+import {
+	defaultIsHeightAndColorMetricLinked,
+	IsHeightAndColorMetricLinkedAction,
+	IsHeightAndColorMetricLinkedActions
+} from "./isHeightAndColorMetricLinked.actions"
+
+export function isHeightAndColorMetricLinked(state = defaultIsHeightAndColorMetricLinked, action: IsHeightAndColorMetricLinkedAction) {
+	switch (action.type) {
+		case IsHeightAndColorMetricLinkedActions.TOGGLE_IS_HEIGHT_AND_COLOR_METRIC_LINKED:
```
shaman-apprentice, line 9: [Optional] although this is just a simple "if else", it would be easy to write a quick unit test for it :)

visualization/app/codeCharta/ui/ribbonBar/linkHeightAndColorMetricButton/linkHeightAndColorMetricButton.component.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Component, Inject } from "@angular/core"
+import { Store } from "../../../state/angular-redux/store"
+import { isHeightAndColorMetricLinkedSelector } from "../../../state/store/appSettings/isHeightAndColorMetricLinked/isHeightAndColorMetricLinked.selector"
+import { toggleHeightAndColorMetricLink } from "../../../state/store/appSettings/isHeightAndColorMetricLinked/isHeightAndColorMetricLinked.actions"
+
+@Component({
+	selector: "cc-link-height-and-color-metric-button",
+	template: require("./linkHeightAndColorMetricButton.component.html")
+})
+export class LinkHeightAndColorMetricButtonComponent {
```
shaman-apprentice, line 11: same with effect, I prefer `LinkColorMetricToHeightMetricButtonComponent`

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.html
```diff
@@ -1,106 +1,115 @@
-<cc-search-panel></cc-search-panel>
+<div class="ribbon-bar">
```
shaman-apprentice, line 1: do we need this extra wrapper? We could apply all styles of `.ribbon-bar` to the element `cc-ribbon-bar`, can we not?

visualization/app/codeCharta/ui/ribbonBar/ribbonBar.component.scss
```diff
@@ -147,6 +175,13 @@ cc-ribbon-bar {
 		min-width: 120px;
 	}
 
+	mat-card.link-metrics-card {
+		display: flex;
+		justify-content: center;
+		max-width: 50px;
+		min-width: 50px;
```
shaman-apprentice, line 182:
```suggestion
		flex-grow: 0;
```

shaman-apprentice (commented):
What do you think about making the brush of the color chooser grey as well, when it is disabled?

Besides that I mainly did a technical review as requested. I really like your effect approach and the way you test the new component - great job! :)

Hall-Ma:
> 

Thank you so much for your feedback!

ce-bo:
I really like the idea of @shaman-apprentice to disable the brush icon, when the metrics are linked. Let's do that :-)

visualization/app/codeCharta/ui/ribbonBar/colorMetricChooser/colorMetricChooser.component.spec.ts
Hall-Ma: Thanks for finding it out! I will change it and Luis opens a new ticket.

visualization/app/codeCharta/ui/ribbonBar/colorMetricChooser/colorMetricChooser.component.html
```diff
@@ -1,9 +1,10 @@
-<i class="fa fa-paint-brush"></i>
+<i [style.color]="(isHeightAndColorMetricLinked$ | async) ? disabledColor : nonDisabledColor" class="fa fa-paint-brush"></i>
```
Hall-Ma, line 1: @shaman-apprentice Would do you think about my implementation? Is there a better way to change the icon color?

visualization/app/codeCharta/ui/ribbonBar/colorMetricChooser/colorMetricChooser.component.spec.ts
```diff
@@ -22,16 +23,29 @@ describe("colorMetricChooserComponent", () => {
...
+		await expect(options[1].textContent).toMatch("bMetric (2)")
 
 		await userEvent.click(options[1])
 		expect(screen.queryByText("aMetric (1)")).toBe(null)
 		expect(screen.queryByText("bMetric (2)")).not.toBe(null)
+		expect(container.querySelector(".fa.fa-paint-brush").getAttribute("style")).toEqual(nonDisabledColor)
```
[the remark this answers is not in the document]
shaman-apprentice: the async pipe in `[style.color]="(isHeightAndColorMetricLinked$ | async) ? disabledColor : nonDisabledColor" ` does some magic under the hood in general (it subscribes to the observable, and sets the real proprety). I am not sure how it works under the hood with native style proprety. A workaround could be to subscribe in the component and use a plain value in the template. I will maybe have Monday afternoon some more time to dig into it.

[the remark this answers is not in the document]
Hall-Ma: I found the mistake. The rgba color of the black color was not set correctly. The opacity value was missing but now it works :)

visualization/app/codeCharta/ui/ribbonBar/colorMetricChooser/colorMetricChooser.component.html
shaman-apprentice: I personally would not set those values in the component but inline the values here in the template. But this is fine as well.

jannikr (approved):
lgtm :)

PR #3062, opened 2022-09-29
Fix Analysis Container stopping when detached

MW-Friedrich:
# Fix Analysis Container stopping when detached

Issue: #2946

## Description

When the CodeCharta Analysis container is started in detached mode (like with docker compose), it exits and stops almost immediately, because the chosen entrypoint does not keep the system busy. This replaces the entrypoint of the Dockerfile with a better one


jannikr (approved):
LGTM

PR #3064, opened 2022-09-29
Refactor/Clean up custom config builder

Hall-Ma:
# Clean up custom config builder

## Description

I have often been annoyed with the Custom Config Builder, especially when implementing a new state property for CodeCharta. This meant that you had to add the new property manually each time. But it is much easier if you directly map the state to the custom config. This saves duplications and is less error-prone. 
And I do not see any case where state values should not be complete. Besides, I didn't have to change a single test!
Also, we can still load old configs into CodeCharta.

### DIFF
--- a/visualization/app/codeCharta/util/customConfigBuilder.ts
+++ b/visualization/app/codeCharta/util/customConfigBuilder.ts
@@ -18,126 +18,14 @@ export function buildCustomConfigFromState(configName: string, state: State, cam
 		mapChecksum: customConfigFileStateConnector.getChecksumOfAssignedMaps(),
 		customConfigVersion: CUSTOM_CONFIG_API_VERSION,
 		stateSettings: {
-			appSettings: undefined,
-			dynamicSettings: undefined,
-			fileSettings: undefined,
-			treeMap: undefined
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
The intention of the current implementation was to notice unintented breaking changes. Users are going to rely upon the format as it is. But if that format changes, it could cause the custom configs to not being functional anymore (causing errors) or to have an unintented behavior.

ce-bo:
As @BridgeAR described, the implementation and the experience for the developers in such cases was intentional. I think we should document this approach, so that the intention is clear when working on this point.

Hall-Ma:
> The intention of the current implementation was to notice unintented breaking changes. Users are going to rely upon the format as it is. But if that format changes, it could cause the custom configs to not being functional anymore (causing errors) or to have an unintented behavior.

I see your point. Nevertheless, we have tests to cover this case. They will give the alarm if something was changed in the custom config model. I would like to continue to change it.

BridgeAR:
@Hall-Ma do the tests cover all entries similar to this implementation? And is it clear from the tests that changing them implies a breaking behavior?

PR #3068, opened 2022-10-02
Refactor/2318/migrate threejs services

shaman-apprentice:
Migrate isAttributeSideBarVisible.service, threeScene.service, threeCamera.service and threeStats.service to Angular


visualization/app/codeCharta/services/isAttributeSideBarVisible.service.ts
```diff
@@ -0,0 +1,16 @@
...
+
+	constructor(@Inject(ThreeSceneService) private threeSceneService: ThreeSceneService) {
+		this.threeSceneService.subscribe("onBuildingSelected", () => {
+			this.isOpen = true
+		})
+		this.threeSceneService.subscribe("onBuildingDeselected", () => {
+			this.isOpen = false
+		})
+	}
+}
```
Hall-Ma, line 16: Why did you choose that 'attritbute sidebar is open' property should be in a service? I am just curious.

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -35,15 +33,15 @@ export class CodeMapArrowService
...
+	onBuildingSelected = (data: { building: CodeMapBuilding }) => {
+		if (this.isEdgeApplicableForBuilding(data.building)) {
 			this.clearArrows()
 			this.showEdgesOfBuildings()
 		}
 		this.scale()
 	}
 
-	onBuildingDeselected() {
+	onBuildingDeselected = () => {
```
Hall-Ma, line 44: Why did you choose an arrow function here? What is the advantage of it?

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
```diff
@@ -35,15 +33,15 @@ export class CodeMapArrowService
 		this.scale()
 	}
 
-	onBuildingSelected(selectedBuilding: CodeMapBuilding) {
-		if (this.isEdgeApplicableForBuilding(selectedBuilding)) {
+	onBuildingSelected = (data: { building: CodeMapBuilding }) => {
```
Hall-Ma, line 36: Why do you choose to make this type less restrictive?

visualization/app/codeCharta/services/isAttributeSideBarVisible.service.ts
shaman-apprentice: good question! Previous we had isAttributeSideBarVisbile in the store, which was always set, when appStatus.selectedBuildingId was set in the store. But the store should only contain atomic values and isAttributeSideBarVisbile is obvious derived from selectedBuildingId. Naming can probably be further improved, but better as before... :)

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
shaman-apprentice: the arrow function binds the `this` of the class to the function. In JS `this` is not bound to a class by default but derived from its calling context (a design flaw in my personal opinion). In case you are really curious https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this ;)

visualization/app/codeCharta/ui/codeMap/codeMap.arrow.service.ts
shaman-apprentice: I haven't. It is just wrapped in an object with a key "building" to align with the EventEmitter schema

Issue #3074, opened 2022-10-06
IDE gets confused with toMatch function


LuisGuenther:
IDE gets confused with different toMatch functions of jest (unit tests) and jest/expect-puppeteer (e2e tests)

Jest's toMatch does not need an await because it is not async, but the other toMatch is. IDE marks code for missing await keyword. The expect is set globally, so it does not know of the expect comes from jest or jest/expect-puppeteer

Acceptance criteria:

- [x] all appearances of toMatch are checked for synchronous or asynchronous behavior
- [x] research for possibilities to evade this problem

PR #3075, opened 2022-10-06
Fix/3074/Remove await when using jest 'toMatch' function

Hall-Ma:
# Remove asynchronous function call when using jest `toMatch()`

Closes: #3074

## Description

Switch asynchronous to synchronous function calls when using jest `toMatch()`. IntelliJ get confused because of jest/expect-puppeteer `toMatch()` function (that is not used in our case).

### DIFF
--- a/visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
+++ b/visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
@@ -61,14 +61,14 @@ describe("AttributeSideBarComponent", () => {
 		expect(isSideBarOpen).toBe(true)
 
 		const name = container.querySelector("a[class=node-link]").textContent
-		await expect(name).toMatch("root/big leaf.ts")
+		expect(name).toMatch("root/big leaf.ts")
 
 		const pathName = container.querySelector("cc-node-path").textContent
-		await expect(pathName).toMatch("/root/big leaf")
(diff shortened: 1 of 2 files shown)


visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
```diff
@@ -61,22 +61,22 @@ describe("AttributeSideBarComponent", () => {
 		expect(isSideBarOpen).toBe(true)
 
 		const name = container.querySelector("a[class=node-link]").textContent
-		await expect(name).toMatch("root/big leaf.ts")
+		expect(name).toMatch("root/big leaf.ts")
```
ce-bo, line 64: The test cases are async functions. So we also would have to change the `it` cases. However, I don't think that this is a good solution. IntelliJ should be able to resolve the right method that is actually called.

ce-bo (changes requested):
Let's investigate further how we can fix this by not changing the test cases.

visualization/app/codeCharta/ui/attributeSideBar/attributeSideBar.component.spec.ts
Hall-Ma: I am sorry but I think you are wrong. That's all jest synchronous 'toMatch'. 
image

ce-bo (approved):

PR #3090, opened 2022-10-12
Feature/2489/implement partially applicable entries in custom configs

Hall-Ma:
# Implement partially applicable entries in custom configs

Issue: #2489

## Description

- custom configs are now applicable even if selected maps or map selection mode differs
- partially applicable custom configs are greyed out but still clickable
- a tooltip shows what is still missing to complete the view (when something is missing)


ce-bo (commented):
LGTM so for :-)
I tested it lokally and the behavior looks fine!

jannikr (approved):
LGTM :)

Issue #1066, opened 2020-07-05
Provide tooltip / description for each metric


BridgeAR:
# Feature request

## Description

As a user, I want to understand each metric right away. To do so, I want to read tooltips about the functionality of the visualized metric.

Feature: Provide tooltip / description for each metric - both primary and secondary. These descriptions have to be added to the cc.json file from the importer. The fields are optional to stay backwards compatible

## Acceptance criteria

- Change the cc.json in a backwards compatible way to accept new meta data. This meta data contains descriptions for the used metrics
- The descriptions are defined in each importer. The importer adds them to the cc.json
- Additionally, read-more links (documentation links) can be specified optionally.
- ~~The descriptions explain what a high or low value indicates for a specific metric.~~ (Will be done in #3232)
- The visualizer understands the new cc.json file and adds the new information to the metrics as tooltips
- Metric Names in the Frontend have a tooltip with the short description including the read-more link (Legend, Selectboxes of metrics, primary/secondary metrics in sidebar when clicking on a building, etc.).
- Full test coverage for the new / changed code

## Development notes (optional Task Breakdown)

Next important TODO(s):
- [x] Display the new attribute descriptors in the legend #3162
- [ ] Display the new attribute descriptors in the metric choosers #3273 ~~#3163~~ 
- [ ] Display the new attribute descriptors in the sidebar. #3273 #3239 
    - [x] Show Descriptor Title, Description, Hint High Value and Hint Low Value maybe in a tooltip
    - [ ] Link to external documentation (Attribute Descriptor Link). #3273 

We can skip the Custom Views for now. Keep in mind that we want the current approach with metricTitles.ts (or similarily named) to provide descriptions for cc.json files without attribute descriptors as fallback.

It is recommended to open separate PRs to update the importers (maybe just one at a time), to update the cc.json structure for the new information and to update the visualizer. That way it's easy to review and to quickly land the PRs.

> "title" field added in #3158 

- [x] Model #3091 #3158
- [x] Filter/EdgeFilter #3091
- [x] Filter/MergeFilter #3091
- [x] Filter/StructureModifier #3091
- [x] Import/CodeMaatImporter #3091
- [x] Import/CSVImporter #3091
- [x] Import/GitLogParser #3091
- [x] Import/MetricGardenerImporter #3091
- [x] Import/SonarImporter #3149
- [x] Import/SourceCodeParser #3166 
- [x] Import/SVNLogParser #3176 
[...]

PR #3091, opened 2022-10-14
Implementation of attributeDescriptors, a new json field that describes used metrics

phanlezz:
# attributeDescriptors basic implementation #1066

Issue: #1066 

## Description

This adds a new field to the produced json files from cc-analysis. This field functions as a map to describe the various attributes produced from the different importers/parsers. Those attribute/metric explanations can then be displayed in the cc-visualization to give the user additional information.

## Screenshots or gifs
grafik

This is still a draft and only includes the outlines of the implementation, some features are not done or incomplete.
If there are any recommendations for changes, let me know 👍


analysis/model/src/main/kotlin/de/maibornwolff/codecharta/model/AttributeDescriptor.kt
```diff
@@ -9,4 +9,26 @@ class AttributeDescriptor(
...
+        if (javaClass != other?.javaClass) return false
+
+        other as AttributeDescriptor
+
+        if (description != other.description) return false
+        if (hintLowValue != other.hintLowValue) return false
+        if (hintHighValue != other.hintHighValue) return false
+        if (link != other.link) return false
+
+        return true
```
ce-bo, line 24: You can return the result directly if you use && conditions.

analysis/test/golden_test.sh
```diff
@@ -58,7 +58,7 @@ check_gitlogparser_repo_scan() {
   echo " -- expect GitLogParser repo-scan subcommand to produce valid cc.json file"
   ACTUAL_GITLOG_JSON_REPO_SCAN="${INSTALL_DIR}/actual_gitlogparser_repo_scan.cc.json"
   returnCode="0"
-  timeout 60s "${CCSH}" gitlogparser "repo-scan" --repo-path "$(dirname "$(dirname "$PWD")")" -o "${ACTUAL_GITLOG_JSON_REPO_SCAN}" -nc || returnCode="$?"
+  timeout 120s "${CCSH}" gitlogparser "repo-scan" --repo-path "$(dirname "$(dirname "$PWD")")" -o "${ACTUAL_GITLOG_JSON_REPO_SCAN}" -nc || returnCode="$?"
```
ce-bo, line 61: Is it really needed to double up the time due to your changes?

phanlezz: Not required due to my changes, but rather due to my system. I was not able to fully run the integration test, because the command took so long. 
The change does not make the tests slower (only if it would fail, because it needs to wait longer until timeout).
 If you want, I can revert this change, and use a local version for testing?

ce-bo: Personally, I would not push those changes that are ment only for my local system.

ce-bo (approved):
In general it is LGTM! I did not have a look at the Unit Tests so far, but I am planning to do this (it should not be a blocker for the merge though).

analysis/test/golden_test.sh
phanlezz: I definitly agree, but it makes sense for this command, because it is a 2-in-1, so it should be able to run for longer than the other one.

ce-bo: Ah ok, then the merge of the two gitlog parser commands does increase the needed time. I see. I would recommend to use the lowest possible/useful time.

analysis/filter/EdgeFilter/src/test/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ProjectMergerTest.kt
```diff
@@ -88,59 +54,37 @@ class ProjectMergerTest : Spek({
...
-            val number: Int = getAttributeValue(leaf4.attributes, "avgCommits")
-            val expectedValue = 20 + 30 + 40 // see testfile
-            MatcherAssert.assertThat(number, CoreMatchers.`is`(expectedValue))
-        }
+    @Test
+    fun `should contain attributeDescriptors after merge`() {
+        val originalProject = ProjectDeserializer.deserializeProject(
+            InputStreamReader(this.javaClass.classLoader.getResourceAsStream(TEST_EDGES_JSON_FILE_3)!!)
+        )
+        val project = EdgeProjectBuilder(originalProject, '/').merge()
```
ce-bo, line 75: With what is the originalProject merged here? I think it would make sense to merge to different sets of attributeDescriptors from two different projects/jsons, right?

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemoverTest.kt
```diff
@@ -91,4 +94,22 @@ class NodeRemoverTest {
 
         Assertions.assertThat(result.blacklist.map { it.path }).doesNotContainSubsequence("/root/foo")
     }
+
+    @Test
+    fun `should copy attributeDescriptors`() {
```
ce-bo, line 99: Small remark: It is not directly obvious for me what this method is testing. Suggestion: Add the reason why it should copy the descriptors.

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractorTest.kt
```diff
@@ -80,4 +82,15 @@ class SubProjectExtractorTest {
         val edges = result.edges
         Assertions.assertThat(edges).isEmpty()
     }
+
+    @Test
+    fun `Subproject still contains all necessary attributeDescriptors`() {
```
ce-bo, line 87: Small remark: Keep the given/when/then pattern in mind. I would always include the WHEN and the THEN in the method name.This makes it clear to every other developer whats going on here.

analysis/import/CSVImporter/src/test/kotlin/de/maibornwolff/codecharta/importer/csv/CSVImporterTest.kt
```diff
@@ -35,8 +35,10 @@ class CSVImporterTest {
             )
         )
         val file = File("src/test/resources/csvimporter.cc.json")
-        val testJsonString = File("src/test/resources/csvimporter.cc.json").readText()
-        val expectedJsonString = File("src/test/resources/csvimporter_different_path_column_name.cc.json").readText()
+        val expectedJsonFile = File("src/test/resources/csvimporter_different_path_column_name.cc.json")
+        val testJsonString = file.readText()
+        System.err.println(testJsonString)
```
ce-bo, line 40: I think the System println is not intentional, right?

analysis/import/CSVImporter/src/test/kotlin/de/maibornwolff/codecharta/importer/csv/CSVProjectBuilderTest.kt
```diff
@@ -93,9 +94,31 @@ class CSVProjectBuilderTest {
...
         val project = csvProjectBuilder
-            .parseCSVStream(this.javaClass.classLoader.getResourceAsStream("sourcemonitor.csv"))
+            .parseCSVStream(this.javaClass.classLoader.getResourceAsStream("sourcemonitor.csv")!!)
             .build()
 
         assertEquals(39, project.size)
     }
+
+    @Test
+    fun `should include given attributeDescriptors`() {
```
ce-bo, line 104:
```suggestion
    fun `when importing a valid csv file then appropriate attribute descriptors have been added`() {
```

analysis/import/GitLogParser/src/test/kotlin/de/maibornwolff/codecharta/importer/gitlogparser/GitLogParserTest.kt
```diff
@@ -13,13 +15,13 @@ class GitLogParserTest {
             arrayOf(
                 "repo-scan",
                 "--repo-path=../../..",
-                "--output-file=src/test/resources/gitlog-analysis.cc.json",
+                "--output-file=src/test/resources/gitlog-analysis-repo.cc.json",
```
ce-bo, line 18: Thank you for clarifying this 👍

ce-bo (approved):

analysis/filter/EdgeFilter/src/test/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ProjectMergerTest.kt
phanlezz: The EdgeProjectBuilder does not have a build function, it only has the function called merge.
This is because it "merges" the edges into the existing nodes (if there are any). 
I will change the name of the test to clarify. This tests sole purpose is to verify, if the attributeDescriptor field gets copied while "merging"/building the new project out of the old parts.
I don't think that there is more to cover for testing attributeDescriptors in EdgeFilter, because if there were two sets of attributes, they were already combined in the attributeDescriptor-field before the filter becomes active, which means that a copy of the complete field is enough.
Or am I missing something here?

ce-bo: Alright, that's fine for me.

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/NodeRemoverTest.kt
phanlezz: I will rename this test. This test verifies that the attribute descriptors do not get changed by a remove action, when there is no need to do so.

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/SubProjectExtractorTest.kt
phanlezz: I will change this name.

PR #3093, opened 2022-10-14
Refactor/3092/state restoring

### LINKED ISSUE #3092: Refactor Improve event data flow for restoring state
# Feature request

## Description

We should refactor the way we set the state of the store when a custom config is applied. 
Currently the properties of appsettings, dynamicsettings, filesettings etc. are restored through their splitter functions.
This leads to redundant code and it's also easy to forget to create new splitter functions when the properties of the Custom Configs change. The consequence is that the Custom Config is not fully applied.

The solution is to write a new root reducer that takes care about the new state (comming from the Custom Config) and only the reducer has to know how the state is set. After that we can delete all splitter function. As additional advantage, only one action is fired instead of an action for every field in store

## Acceptance criteria
- All splitter related code is deleted
- Everything works as before

shaman-apprentice:
close #3092


visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -227,19 +225,8 @@ export class CustomConfigHelper {
...
-		//  due to issues of the event architecture.
-
-		// TODO: Check if state properties differ
-		// Create new partial State (updates) for changed values only
 		store.dispatch(setState(customConfig.stateSettings))
 
-		// Should we fire another event "ResettingStateFinishedEvent"
-		// We could add a listener then to reset the camera
-
-		store.dispatch(setColorRange(customConfig.stateSettings.dynamicSettings.colorRange))
```
shaman-apprentice, line 240: this is actual needed. Otherwise wie set the state, which sets blacklist to [], which leads to recalculation of selectedColorMetricDataSelector and therefore resetColorRange.effect is triggered. 

I wonder if it is really expected to reset color range when excluding a node? I suggest to remove `selectedColorMetricDataSelector` from resetColorRange.effect.

ce-bo: This should not be the case. Excluding a file/folder should not reset the color range.

visualization/app/codeCharta/state/store/state.reducer.ts
```diff
@@ -16,4 +18,52 @@ const rootReducer = combineReducers({
...
+		if (value === null || value === undefined) {
+			continue
+		}
+
+		const newComposedPath = [...composedPath, key]
+		const composedJoinedPath = newComposedPath.join(".")
+
+		if (!isKeyOf(applyTo, key)) {
+			continue
+		}
```
ce-bo, line 56:
```suggestion
                if (!isKeyOf(applyTo, key)) {
			continue
		}

		const newComposedPath = [...composedPath, key]
		const composedJoinedPath = newComposedPath.join(".")
```

ce-bo: Change order of if condition and const assignments.

ce-bo (commented):
LGTM so for!

visualization/app/codeCharta/util/customConfigHelper.ts
shaman-apprentice: I will update accordingly on Monday

shaman-apprentice: done

LuisGuenther (approved):
Looks good to me.

PR #3098, opened 2022-10-20
refactor: extract tree map size into a const as this is never changed

shaman-apprentice:
extract tree map size into a const as this is never changed

ref #2318


RomanenkoVladimir:
LGTM!

My changes should also be fine, as I am no longer using treeMap.mapSize for calculation.
It is only kept for legacy calculations with fixed folders

RomanenkoVladimir (approved):

PR #3107, opened 2022-10-27
Refactor/2318/update babel

shaman-apprentice:
`transform-es2015-modules-commonjs` was renamed / moved to `@babel/plugin-transform-modules-commonjs` a long time ago and still used `core-js@2`. Unfortunately we missed it, as renovate doesn't catch packages renaming


BridgeAR (approved):

shaman-apprentice:
as discussed in weekly: We don't know why the pipeline fails with heap out of memory. Locally it works. Further investigation is postponed until removal of AngularJS things.

shaman-apprentice:
superseded by #3128

PR #3120, opened 2022-11-03
Add font-family to global style sheet

Hall-Ma:
# Add font family to global style sheet

## Description

- Due to the final migration we have lost our font family
- Now the font family `Roboto, "Helvetica Neue", sans-serif` is set in the global style sheet


phanlezz (approved):
LGTM! 🚀

Issue #3122, opened 2022-11-04
Focusing or hovering a building with size 0 throws errors and leads to a crash


RomanenkoVladimir:
# Bug

## Expected Behavior

GIVEN
A building with size 0 (e.g. kotlin test files in the codeCharta map file, when using MCC as area metric)

WHEN
the user focuses the building or hovers it in the side bar

THEN
nothing happens/ they get a notification that it is not possible.

## Actual Behavior
the application throws console errors for hovering or crashes and stops responding to further input for focus

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes | no
- CodeCharta Version: 1.110.0
- OS: all
- Browser: all

RomanenkoVladimir:
I decided to implement it with a skip to keep it consistent with other behavior: exclude, flatten, highlight
As all of those are ignored as well. We should probably disable all those options once #3113 is implemented.
But for now a hard crash freezing the app should be avoided

Issue #3113, opened 2022-10-31
Highlight folders/buildings with area zero (0) in navigation bar


RomanenkoVladimir:
# Feature request

<Please note by far the quickest way to get a new feature is to file a Pull Request.>
<We will consider your request but it may be closed,>
<if it's something we're not actively planning to work on>
<or it does not match our visison of this software.>

## Description

As an auditor, I want to identify buildings that aren't rendered due to a metric area of 0 to avoid confusion due to them not appearing on the map.

Currently, some buildings are not rendered due to having a building area of 0. However, this is difficult to see for a user hovering them in the map and leads to confusion. We should provide some form of visual feedback that this buildings/ folders are not missing but skipped due to having no area value for this metric.

## Acceptance criteria
- Un-rendered buildings with area 0 are highlighted in some form in the navigation bar.

## Open questions

- How should we highlight buildings? @BridgeAR proposed crossing them though, or we could use a color for this.
- Do we need/ want an additional tooltip that notifies the user when they attempt to hover a missing building from the navigation bar?

PR #3123, opened 2022-11-07
Bug/3122/focusing on building with size zero crashes the app

RomanenkoVladimir:
# Focusing or hovering an un-rendered building results in a crash

closes: #3122 

## Description

Currently focusing an un-rendered building, e.g. one that has an area metric of 0 or is missing due to margins crashes the app and prevents further input.

I decided to ignore the input in such cases to keep it consistent with the behavior for flatten, highlight, and exclude.
In the future all those options should be disabled by default, see feature #3113 

Furthermore after the latest change hovering any building with area 0 leads to console error as we attempt to draw an edge, this was also addressed in this PR.


BridgeAR (approved):
RSLGTM

PR #3128, opened 2022-11-09
refactor: split unit and e2e configs

### LINKED ISSUE #2300: Seperate unit and e2e test config
e2e and unit tests need different test setup. E.g. unit test need jsdom and jest-preset-angular/setup-jest. On the other side e2e tests need jest-puppeteer/jest-preset and must not use jest-preset-angular/setup-jest.

Therefore they should use different jest config files instead of adding custom cli flags or if else before require statements.

shaman-apprentice:
close #2300

OPTIONAL todos:
- [x] update test dependencies: Updated babel-loader and transform-es2015-modules-commonjs. Jest Dependencies are petulant. Probably it is easer to migrate to ng tools first and clean up out custom build process
- [ ] ~extract and share common jest configs between unit and e2e~
- [ ] ~**sometimes** jest exists with open handlers. First pipeline had one failing test within codeMap.arrow.service.spec. Investigation would be nice (not related to this PR / it is the same in current main branch)~


Hall-Ma (approved):
Everything works fine :)

Issue #3116, opened 2022-11-01
Upgrade to Angular 15


shaman-apprentice:
# Feature request

Angular 12 has end of life on [2022-11-12](https://angular.io/guide/releases#actively-supported-versions). Therefore we should migrate to Angular 13+.

## Development notes (optional Task Breakdown)
- [x] split e2e test settings from unit test settings (#2300)
- [x] upgrade jest to at least v27 as jest-preset-angular seems to need it in combination with angular's build tools
- [x] replace custom build through Angular's ng tooling. Replace hard coded templates requires (`template: require("./metricDeltaSelected.component.html")`)  through plain paths, and use `templateUrl`, so that Angular's build can pick them up
- [x] Follow Angular's migration guide from [12 -> 13](https://update.angular.io/?v=12.0-13.0) -> 14 -> 15

## Clean up tasks / further enhancements

- [x] ~~Update with Material 15 marked deprecated things~~ Own ticket #3225
- [x] @Inject(...
- [x] build warnings "CommonJS or AMD dependencies can cause optimization bailouts." - did quick wins
- [x] unify place for assets (visualizsation/assets and visualizsation/codeCharta/assets) 
- [x] "strictTemplates": false -> true
- [x] .md-fab // dead css
- [x] remove deprecated nodeEdgeMetricsMap
- [x] ~~remove clone from getUndecoratedAccumulatedData~~ is still needed. See https://github.com/MaibornWolff/codecharta/pull/3217
- [x] replace custom `updateProperties.ts` again

shaman-apprentice:
everything is done or an own issue was created

PR #3184, opened 2023-01-02
refactor: use correct brackets

shaman-apprentice:
which surprisingly doesn't change evaluation due to JS type conversion. But strict template checking of angular build tools would complain, as e.g. `false > 0` is not allowed

ref #3116

### DIFF
--- a/visualization/app/codeCharta/ui/searchPanel/mapTreeView/mapTreeViewItemName/mapTreeViewItemName.component.html
+++ b/visualization/app/codeCharta/ui/searchPanel/mapTreeView/mapTreeViewItemName/mapTreeViewItemName.component.html
@@ -1,7 +1,7 @@
 <span
 	class="node-name"
 	[class.flattened]="node.isFlattened"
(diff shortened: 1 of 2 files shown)


BridgeAR (commented):
> which surprisingly doesn't change evaluation due to JS type conversion

The JS type conversion should happen before the evaluation of the statement, right? This does seem to change the behavior from `(!value) > 0` to `!(value >0)`?

shaman-apprentice:
> > which surprisingly doesn't change evaluation due to JS type conversion
> 
> The JS type conversion should happen before the evaluation of the statement, right? This does seem to change the behavior from `(!value) > 0` to `!(value > 0)`?

Exactly.

Strict template checking of angular build tools will complain, that `>` operator is only allowed for numbers, as `!value` will already be a boolean before `>` gets executed.

I think the relevant cases are as followed:
|           | (!value) > 0 | !(value > 0) |
|-----------|--------------|--------------|
| 2         | false        | false        |
| 0         | true         | true         |
| undefined | true         | true         |

BridgeAR (approved):
This works due to the right side of the check being zero.

PR #3186, opened 2023-01-02
refactor: use string union instead of enum for blacklist type to impr…

shaman-apprentice:
change 

```ts
export enum BlacklistType {
	flatten = "flatten",
	exclude = "exclude"
}
``` 

to

```ts
export type BlacklistType = "flatten" | "exclude"
```

which satisfies strict type checking for template of [searchBar.component](https://github.com/MaibornWolff/codecharta/blob/main/visualization/app/codeCharta/ui/searchPanel/searchBar/searchBar.component.html#L24)

ref #3116


BridgeAR:
Already done

Issue #3193, opened 2023-01-12
A todo in the customConfigHelper is due and now blocks


MW-Friedrich:
# A todo in the customConfigHelper is due 

The removal of transformLegacyCameraSettingsOfCustomConfig has had its TODO due on the 1st of January 2023.
This was put in by #2880 

## Steps to Reproduce the Problem
1. Run the pipeline checks

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): ~~yes~~ | no
- CodeCharta Version:
- OS:
- Browser:

PR #3196, opened 2023-01-12
Fix/3193/custom config todo due

MW-Friedrich:
# The removal of the camera attribute migration for custom view files is due.

Issue: #3193 

## Description

In #2880 the way camera position and angle were saved changed in an incompatible way. To give everyone some time to transition, a mapping method was introduced to parse the old camera attributes into the new format.

Now, the time has come for its removal, the camera position from old files will now simply be ignored.
This affects all cc.config files generated with CodeCharta 1.101.1 and earlier.

This PR also updates the babel suite because tests were failing locally on my machine.


ce-bo (commented):
After some considerations I think that we should just remove the transform and a check if the config is conform with the supported config strucutre is also not needed as long as the system does not crash. If it would crash we should handle the error gracefully.

What is happening when loading/importing a config with old camera properties? I guess, they are just ignored and a default camera position is set instead. in that case, I would tolerate this because such deprecated configs must have been created before releasing the feature as stable (so we do not necessarily have to support the transformation of "unstable" configs).

The reason for rollbacking the before discussed solution is that in the future we should handle backwards incompatible config structure changes by the config structure api version and add a appropriate validation method for applying/importing configs.

MW-Friedrich:
> After some considerations I think that we should just remove the transform and a check if the config is conform with the supported config strucutre is also not needed as long as the system does not crash. If it would crash we should handle the error gracefully.
> 
> What is happening when loading/importing a config with old camera properties? I guess, they are just ignored and a default camera position is set instead. in that case, I would tolerate this because such deprecated configs must have been created before releasing the feature as stable (so we do not necessarily have to support the transformation of "unstable" configs).
> 
> The reason for rollbacking the before discussed solution is that in the future we should handle backwards incompatible config structure changes by the config structure api version and add a appropriate validation method for applying/importing configs.

The changes have been made, and everything seems to run bug free on my machine. The camera position is now simply ignored.

I have also bumped the api version of the config files to better signalise the change.

ce-bo:
Let's not bump the version because the deprecation of theese properties occurred when the Custom Views feature was not released stable. It is not ideal that we are already at version 1.0.0. Before releasing the feature as stable the structure should have been at 0.x.x.

LuisGuenther (approved):

Issue #3198, opened 2023-01-13
Git Information does not get added to web-demo cc.json


MW-Friedrich:
# Bug

## Expected Behavior

WHEN we make a release

THEN several CC importers, including the gitlogparser, run on our codebase to provide an accurate view of the project, as well as give anyone curious a good overview.

## Actual Behavior
Since the behaviour of the gitlogparser was changed, the command in the 'run building script build gh pages' step of the release pipeline fails, since the command was not updated there.

## Screenshots
![grafik](https://user-images.githubusercontent.com/108457762/212345218-92ce2ab1-8aa1-439f-b573-5134adf6c49e.png)

## Steps to Reproduce the Problem
1. Trigger the release pipeline

## Notes
It might also be a good idea to check our documentation for further uses of outdated commands. Additionally, I would propose to change the name of this step in the pipeline, it is confusing and not very intuitive.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes | ~~no~~
- CodeCharta Version: 1.114.0
- OS: n/a
- Browser: all

PR #3201, opened 2023-01-19
Fix/3198/codecharta demo file without git info

MW-Friedrich:
# Fixes missing git information in web demo cc.json

Issue: #3198

## Description
Due to a wrong command name, git information was not added to the cc.json file that is displayed in the web demo. This is now fixed, together with a few small enhancements to naming and formatting.


script/build_demo_file.sh
```diff
@@ -7,13 +7,14 @@ cp -R visualization/dist/webpack/ gh-pages/visualization/app/
...
-cd gh-pages/demo_files
+cd gh-pages/demo_files || exit
 CCSH=../../analysis/build/install/codecharta-analysis/bin/ccsh
 
 git log --numstat --raw --topo-order --reverse -m > git.log
 git ls-files > file-name-list.txt
 
-$CCSH gitlogparser git.log -o codecharta_git.cc.json -n file-name-list.txt -nc
+# Data for for both visualization and analysis
+$CCSH gitlogparser log-scan --git-log git.log --repo-files file-name-list.txt -o codecharta_git.cc.json -nc
```
ce-bo, line 17: Couldn't we use the simpler repo-scan command. Therefore, we do not need to maintain the commands used for generating the git log and the file list.

ce-bo (approved):
LGTM! Thank you for fixing all the typos and updating the documentations. :-)

script/build_demo_file.sh
MW-Friedrich: Yes, that would definetely work, this change is more of a "drop in" fix with the minimal possible changes to not disturb anything else. Additionally, doing the git-log and git-ls calculations like this in the shell is a bit faster.

Issue #3036, opened 2022-09-12
Remove redundant dependencies in analysis


MW-Friedrich:
# Feature request

## Description

As a developer, I want fewer dependencies in my projects so that redundant functionality is removed and attack surface is minimized.
For example, we have a lot of dependencies/libraries for handling json files and content, maybe we can switch to one single library and use all of its functionality.

## Acceptance criteria
- 
- 
- 

## Assumptions & Exclusions
In our Weekly on the 13th of September 2022 we decided that this should be an ongoing task, rather than one "big" issue. 

## Development notes (optional Task Breakdown)
- [ ] Check if we need Wiremock and if yes, switch to the new https://github.com/wiremock/wiremock repository
- [x] Switch from Spek and JUnit4 to JUnit 5 #3202 
- [ ] Decide on one assertion plugin (  jsonassert and assertJ, ... ) are currently in use)
- [ ] Decide on one JSON plugin (JSON, GSON, Jackson and jsonassert are currently in use)
- [ ] Unify/Verify which mockking library we use (io.mockk // mockito-core // com.nhaarman.mockitokotlin etc.)
- [ ] Update ADR to write down these decisions for future reference

ChristianHuehn:
I will pull this out and look for a better place to store it, so that we will keep in in mind when developing CC further.

PR #3202, opened 2023-01-19
Tech/3036/remove spek

phanlezz:
# Removel of Spek2 test suit and hamcrest/hamkrest assertions

Issue: #3036

## Description

During the implementation of new features in the analysis part, we started refactoring tests from the spek2 suit to junit. This PR finishes up the remaining test cases, so that we only use junit5. 
Additionally converts all hamcrest/hamkrest assertions to assertj, since these are not being developed further.

> This branch is a subbranch of an already merged feature branch, thus containing more commits than expected.


MW-Friedrich (approved):
Beautiful new tests, LGTM!

ce-bo:
Very nice, thanks!

PR #3217, opened 2023-01-31
refactor: make manipulation more clear

shaman-apprentice:
refactor: make manipulation more clear and add tests

note that the changes in the snapshot are the expected behavior

ref #3116


BridgeAR:
The actual code has no functionality changed? It is a small refactoring but it does not remove any cloning or the like?

shaman-apprentice:
> The actual code has no functionality changed? It is a small refactoring but it does not remove any cloning or the like?

exactly. I thought we could simply remove the cloning. But we copy the file states and then mutate the copied file states in place to become the aggregated map. We could likely get rid of the cloning, but I see that as a further enhancement out of scope in the Angular migration context.

phanlezz (approved):
LGTM 👍

PR #3221, opened 2023-02-05
refactor: remove lodash as Angular warns about it

shaman-apprentice:
remove lodash as Angular warns about it:

> CommonJS or AMD dependencies can cause optimization bailouts.

ref #3116


visualization/app/codeCharta/util/debounce.ts
```diff
@@ -0,0 +1,7 @@
+export const debounce = <F extends (...arguments_: Parameters<F>) => ReturnType<F>>(f: F, waitInMS = 300) => {
```
BridgeAR, line 1: Let's keep the former scope by using a regular function, accessing its scope, saving it to a variable and then binding the passed through function to that scope.
Also please use at least fn or func as variable name.  Single character variable names are always tricky.

visualization/package.json
```diff
@@ -78,12 +78,11 @@
 		"ajv": "^6.12.6",
 		"color-convert": "^2.0.1",
 		"d3-hierarchy": "^2.0.0",
+		"fast-deep-equal": "^3.1.3",
```
BridgeAR, line 81: Please use https://github.com/lukeed/dequal instead. It is safer and does a better comparison.

shaman-apprentice: [dequal/lite](https://github.com/lukeed/dequal#dequallite) is already enough for us

visualization/app/codeCharta/util/debounce.ts
shaman-apprentice: I have a strong dislike against any usage of an implicit "this". But it is heavily opinionated, so I just added it.

shaman-apprentice: (no need to save `const self = this` as an arrow function  keeps outer this)

visualization/app/codeCharta/util/reduxHelper.ts
```diff
@@ -1,9 +1,9 @@
 import { clone } from "./clone"
-import isEqual from "lodash.isequal"
+import { dequal } from "dequal/lite"
```
BridgeAR, line 2: I understand that we currently only use very this function in a very limited way and lite would here be sufficient. This could in theory change over time though and e.g., `removeItemFromArray()` could handle data that does contain more complex data. As such, we should always use the strictest possible comparison algorithm possible as it is the only way to guarantee that things properly work.
The only case where the lite mode would be safe to use is in case only JSON data is used and that's not the case.

visualization/app/codeCharta/ui/ribbonBar/artificialIntelligence/suspiciousMetrics/suspiciousMetrics.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Component, Input, OnChanges, SimpleChanges, ViewEncapsulation } from "@angular/core"
-import isEqual from "lodash.isequal"
+import { dequal } from "dequal/lite"
```
BridgeAR, line 2: The my comment below.

visualization/app/codeCharta/util/debounce.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+		expect(f).toHaveBeenCalledTimes(1)
+	})
+
+	it("should have type safe parameters", async () => {
+		const debouncedSquare = debounce((n: number) => n * n, 0)
+		// @ts-expect-error
+		debouncedSquare("2")
+		debouncedSquare(2)
+		await wait(0)
+	})
```
BridgeAR, line 23:
```suggestion
```

This test seems rather confusing to me.  There is no assertion in it and we should not include that. There are ways to verify types programmatically but this is something we should IMO not do.

As a sidenote: `wait(0)` is in fact going to wait one millisecond.

BridgeAR (approved):
LGTM % comments

visualization/app/codeCharta/util/debounce.ts
BridgeAR: I am not certain I can follow about the implicit context. It's now explicitly set to the original context? :-)

visualization/app/codeCharta/ui/codeMap/codeMap.mouseEvent.service.ts
```diff
@@ -260,7 +257,7 @@ export class CodeMapMouseEventService {
 		}
 	}
 
-	onDocumentMouseMove(event: MouseEvent) {
+	onDocumentMouseMove = (event: MouseEvent) => {
```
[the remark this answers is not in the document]
shaman-apprentice: I prefer this over previous extra arrow function which binds "this" as well in line 95: `debounce(event => this.onDocumentMouseMove(event), 60)`

visualization/app/codeCharta/ui/ribbonBar/artificialIntelligence/suspiciousMetrics/suspiciousMetrics.component.ts
shaman-apprentice: done

visualization/app/codeCharta/util/debounce.ts
shaman-apprentice: We can have a nice :coffee: / :beers: about it the next time we see each other :)

visualization/app/codeCharta/util/reduxHelper.ts
shaman-apprentice: currently `removeItemFromArray` is in `reduxHelper` and all redux data should be plain JSON data by idea of redux design. But I guess these comparison are currently no bottleneck, so lets be better safe than sorry - adjusted.

visualization/app/codeCharta/util/debounce.spec.ts
shaman-apprentice: > As a sidenote: wait(0) is in fact going to wait one millisecond.

yes, it's the same as omitting the value :+1:

PR #3223, opened 2023-02-05
refactor: replace md5 through native crypto

shaman-apprentice:
as Angular warns about it "CommonJS or AMD dependencies can cause optimization bailouts"

ref #3116

I haven't adjusted the tests so far. @BridgeAR  do you have objections against the replacement of md5 to native crypto?

### DIFF
--- a/visualization/app/codeCharta/services/loadInitialFile/urlExtractor.ts
+++ b/visualization/app/codeCharta/services/loadInitialFile/urlExtractor.ts
@@ -55,7 +55,7 @@ export class UrlExtractor {
 		if (response.status >= 200 && response.status < 300) {
 			const data = response.body
 			const responseData: string | ExportCCFile | ExportWrappedCCFile = ungzip(data, { to: "string" })
-			const content = getCCFileAndDecorateFileChecksum(responseData)
+			const content: ExportCCFile = await getCCFileAndDecorateFileChecksum(responseData)
 			const parsedFileName = this.getFileName(fileName, content.projectName)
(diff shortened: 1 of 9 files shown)


BridgeAR (commented):
The original reason to use md5 was performance: there is no need for a safe hashing algorithm here but performance is critical.
Question is, if the native implementation of the more complex algorithm could still outperform the js implementation. That might be the case.
This does however invalidate all existing hashes. That's a pretty bad breaking change as such. And if that's done, `SHA-1` should be used for performance reasons. I would however just stick to the current implementation.

That aside: the warning provided by Angular is about _potentially_ not being able to remove unneeded files from the bundled code. That is not critical for any of the here used implementations.

BridgeAR:
> I would guess yes, but I wouldn't be too surprised if not tipping_hand_man But as the hash is only calculated when uploading a file, I think it doesn't matter for usability if it takes 2 or 10 seconds. Native implementation is async in contrast to md5, so the loading spinner will smoothly keep spinning :)

Loading time is already pretty bad with big files. Let's not make the experience worse.

> Is it? Do we need the hash for something other than checking if we already have uploaded it?

I am not certain I follow.  No existing custom configs would work anymore. That seems rather bad to me.

> in general I am currently favoring to use as little deps as possible as the updating is quite some work, especially when there is a native solution.

That is definitely a good idea.

> Feel free to close the PR, if you think this is not worth the effort :)

Right now, I do not think we should pursue this.

PR #3224, opened 2023-02-06
Test/3116/cleanup

shaman-apprentice:
I noticed that very rarely codeMap.arrow.service.spec.ts fails in pipeline (e.g. [here](https://github.com/MaibornWolff/codecharta/actions/runs/4102205565/jobs/7074874968#step:6:1865). I have the assumption that it happens, when any parallel running test dispatches hoveredNodeId as we call the debounced function in subscription of constructor. But I couldn't reproduce it.

However I noticed, that we don't clean up our subscriptions. As far as I know in root provided services are never cleaned up in production. So this effects only tests.

ref #3116

### DIFF
--- a/visualization/app/codeCharta/services/loadFile/loadFile.service.ts
+++ b/visualization/app/codeCharta/services/loadFile/loadFile.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from "@angular/core"
+import { Injectable, OnDestroy } from "@angular/core"
 import { tap } from "rxjs"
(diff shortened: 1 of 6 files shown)


ce-bo:
Did I get it right?
Cleaning up our subscriptions might only be necessary for tests in the case the service is provided in root.
In production, it should not cause problems without cleaning up subscriptions because the providedInRoot services are shared single instances.

ce-bo (approved):

Issue #3225, opened 2023-02-07
Migrate to Angular Material v15 MDC components


IhsenBouallegue:
# Feature request

In the context of upgrading to Angular 15 #3116 we need to migrate away from materials legacy components.
But what should have been a straightforward migration is now a full on construction site. 

This is the result after upgrading. (UI completely broken)
![ui after upgrade)](https://user-images.githubusercontent.com/48621967/217210592-8a0be032-42ba-4255-9316-c092a08dabff.png)
Sadly, the update brings a lot of braking changes which affect almost every component. A lot of refactoring needs to be done.

[Material v15 MDC Migration](https://material.angular.io/guide/mdc-migration)

shaman-apprentice:
I noticed another styling issue (now in main) (@IhsenBouallegue ?)

![image](https://user-images.githubusercontent.com/3596742/223539642-f95b8642-4086-4466-bc40-3c500332cf95.png)
vs now
![image](https://user-images.githubusercontent.com/3596742/223539710-00a6b649-dd4f-47ec-9fe7-7147340838af.png)

IhsenBouallegue:
> I noticed another styling issue (now in main) (@IhsenBouallegue ?)
> 
> ![image](https://user-images.githubusercontent.com/3596742/223539642-f95b8642-4086-4466-bc40-3c500332cf95.png)
> vs now
> ![image](https://user-images.githubusercontent.com/3596742/223539710-00a6b649-dd4f-47ec-9fe7-7147340838af.png)
> 

Oh missed that one, will take a look if find the time this week.

ce-bo:
@valenfx also had a look. After merging main into #3234 the custom view list could not be uncollapsed anymore.

Hall-Ma:
What still needs to be done for this?

ce-bo:
Will be closed by #3251.

PR #3226, opened 2023-02-07
Migrate to MDC-based Angular Material Components

IhsenBouallegue:
# Migrating to MDC-based Angular Material Components

Issue: #3225 

## Description

This PR migrates to the new Material Design. The migration breaks the UI which is why a lot of refactoring is needed
This is part of the Angular update #3116

## What changed

- Upgraded all material components to MDC (not using the legacy components any more)
- Changed the primary color to match the color of the CC logo
- Display the CC logo next to MW
- Fixed all unit and e2e tests to work with MDC components (this took a lot of time, we should use the test harnesses provided by material ui instead of accessing the internal CSS classes. I already changed 2 test suites)
- Made the background color a bit lighter
- Saturated the distribution bar colors a little bit
- Improved the layout of all metric menus
image
image
image
image
- Improved the layout of the Global Configuration dialog
image
- Improved the layout of the metric chooser panel. Also fixed the problem where the panel would appear on top of the select.
image
- Fixed various margin problems in the UI. Let's start using css properties flex and gap more, They are more flexible 👍


visualization/app/codeCharta/ui/metricChooser/metricChooser.component.spec.ts
```diff
@@ -56,7 +56,7 @@ describe("metricChooserComponent", () => {
 		})
 
 		await userEvent.click(await screen.findByText("aMetric (1)"))
-		await screen.findByText("search metric (max value)")
+		await screen.getByPlaceholderText("search metric (max value)")
```
shaman-apprentice, line 59: I haven't tested it, but `screen.get*` methods are not async, in contrast to `screen.find*`
```suggestion
		screen.getByPlaceholderText("search metric (max value)")
```

visualization/app/codeCharta/ui/resetSettingsButton/resetSettingsButton.component.html
```diff
@@ -1,4 +1,6 @@
 <button (click)="applyDefaultSettings()" [title]="tooltip">
-	<i class="fa fa-undo"></i>
+	<div class="icon-wrapper">
```
shaman-apprentice, line 2: [optional] do we need this wrapper or could we solve it through selectors of existing elements?

visualization/app/codeCharta/ui/ribbonBar/areaSettingsPanel/areaSettingsPanel.component.spec.ts
```diff
@@ -14,17 +12,6 @@ describe("AreaSettingsPanelComponent", () => {
 		})
 	})
 
-	it("should reset default margin checkbox on changes of margin slider", async function () {
```
shaman-apprentice, line 17: I guess this test is an open todo?

visualization/app/codeCharta/ui/ribbonBar/showScenariosButton/addCustomScenario/addCustomScenario.component.spec.ts
```diff
@@ -19,25 +19,26 @@ describe("AddCustomScenarioComponent", () => {
 			providers: [
 				{ provide: ThreeCameraService, useValue: { camera: { position: new Vector3(0, 300, 1000) } } },
 				{ provide: ThreeOrbitControlsService, useValue: { controls: { target: new Vector3(177, 0, 299) } } },
-				{ provide: MatLegacyDialogRef, useValue: { close: jest.fn() } }
+				{ provide: MatDialogRef, useValue: { close: jest.fn() } }
 			]
 		})
 	})
 
-	it("should enable the user to add a custom scenario, with all properties set initially", async () => {
```
shaman-apprentice, line 27: We should fix this test or create a follow up ticket for it. I strongly prefer to immediately fix it :)

visualization/app/codeCharta/ui/slider/slider.component.spec.ts
```diff
@@ -1,56 +1,56 @@
...
 describe("SliderComponent", () => {
 	beforeEach(() => {
 		TestBed.configureTestingModule({
-			imports: [SliderModule]
-		})
+			imports: [SliderModule],
+			declarations: [SliderComponent]
+		}).compileComponents()
+		fixture = TestBed.createComponent(SliderComponent)
+		loader = TestbedHarnessEnvironment.loader(fixture)
```
shaman-apprentice, line 19: do we really need to dig into Angular's internals? I will check, if I can get the tests working again with testing-library style.

visualization/app/codeCharta/ui/toolBar/globalConfigurationButton/globalConfigurationDialog/mapLayoutSelection/mapLayoutSelection.component.spec.ts
```diff
@@ -1,50 +1,53 @@
...
 	beforeEach(() => {
 		Store["initialize"]()
 		TestBed.configureTestingModule({
-			imports: [MapLayoutSelectionModule]
-		})
+			imports: [MapLayoutSelectionModule],
+			declarations: [MapLayoutSelectionComponent]
+		}).compileComponents()
+		fixture = TestBed.createComponent(MapLayoutSelectionComponent)
+		loader = TestbedHarnessEnvironment.loader(fixture)
```
shaman-apprentice, line 24: do we really need to dig into Angular's internals? I will check, if I can get the tests working again with testing-library style.

shaman-apprentice (changes requested):
First of all thx a lot for tackling the big job of the migration! :) Just a view things we should further tackle:

I have no strong feelings about Design. Is  the central alignment of the ctx menu OK?
image

I couldn't find out why this section is now one px larger to the bottom?
image

shaman-apprentice:
Is it me, my screen or have the colors of the buildings lost some brigthness?

image

visualization/app/codeCharta/ui/slider/slider.component.spec.ts
IhsenBouallegue: That is actually the recommended way in the Material UI docu (with harnesses). Actually the reason why a lot of tests broke (and why this took so long to migrate) is that we used to access internal Material UI classes and hardcode them. Then Material decided to change almost all internal classes, and some parts of the API. To avoid this from happening again, I started using their recommended way with the Test harnesses they provide. Which will be updated when material updates. So no extra work on our part.
I only see upsides to this, so can you shed some light why this is bad?

IhsenBouallegue: Those tests were also super annoying to fix as they were before, but as soon as I switched to using the harnesses it went in a few minutes and was more readable and concise.
Here is more to it here:
https://material.angular.io/guide/using-component-harnesses

IhsenBouallegue:
> First of all thx a lot for tackling the big job of the migration! :) Just a view things we should further tackle:

Most welcome! 😄 

> I have no strong feelings about Design. Is the central alignment of the ctx menu OK? image

That was actually broken because of the migration. It was not intended, will fix it 👍 

> I couldn't find out why this section is now one px larger to the bottom? image

Will investigate that too

visualization/app/codeCharta/ui/resetSettingsButton/resetSettingsButton.component.html
IhsenBouallegue: You are right! we can get by without it.

visualization/app/codeCharta/ui/metricChooser/metricChooser.component.spec.ts
IhsenBouallegue: It works without await too 👍

visualization/app/codeCharta/ui/ribbonBar/areaSettingsPanel/areaSettingsPanel.component.spec.ts
IhsenBouallegue: It is, I would be up for fixing it, if we can use the material ui checkbox harness (As far as I know, this needs access to Angular's internal testing library)
The problem with this test is that we use queryByRole to get the checkbox, but that doesn't seem to work any more.

shaman-apprentice: This test now fails for two reasons:

First of all, because the input was removed. 
image
To be honest - I think this checkbox is useless, as we can reset all settings. I wouldn't mind to remove it to keep the UI lean - @ce-bo  what do you think?

Second we need to make the selector in line 21 more precise, as the slider already has an input and the querySelector finds the first occurence. We could e.g. change to `const marginInput = container.querySelector("input[type=number]")`

IhsenBouallegue: @shaman-apprentice Oh yeah, I removed that because it never made sense to me. I made resetting the settings the same for all menus (a reset button at the end of the menu). Already discussed removing the checkbox with @ce-bo 👍 

I will look into the second point

shaman-apprentice: ah, when we remove the checkbox, we also need to remove the test for the checkbox. So it makes sense :) I think we should add a generic changelog entry like "Rework UI Design"

visualization/app/codeCharta/ui/toolBar/globalConfigurationButton/globalConfigurationDialog/mapLayoutSelection/mapLayoutSelection.component.spec.ts
shaman-apprentice: s. above

visualization/app/codeCharta/ui/ribbonBar/showScenariosButton/addCustomScenario/addCustomScenario.component.spec.ts
shaman-apprentice: done

shaman-apprentice (approved):
Have I missed something? I think we are good to go. Great work, thanks a lot!

visualization/app/codeCharta/ui/slider/slider.component.spec.ts
IhsenBouallegue: Awesome! Glad to hear that we can use both 😄

Issue #3213, opened 2023-01-25
Add notes to a Custom View


ce-bo:
# Feature request

<Please note by far the quickest way to get a new feature is to file a Pull Request.>
<We will consider your request but it may be closed,>
<if it's something we're not actively planning to work on>
<or it does not match our visison of this software.>

## Description

As an Auditor, I want to add custom notes to a Custom View, so that I e.g. can remember what this Custom View was intended for.

## Acceptance criteria
- I can add custom notes when creating a Custom View
- I can edit custom notes for a existing Custom View
- The notes are included in the config file and therefore can be exported automatically

PR #3234, opened 2023-02-22
Feature/3213/add notes to a custom view

RomanenkoVladimir:
# Add notes to a custom view

closes: #3213 

## Description

Allows a user to add a note upon creating a new custom view, as well as Edit an existing custom view note from the "custom-view-menu".


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ## [unreleased] (Added 🚀 | Changed | Removed 🗑 | Fixed 🐞 | Chore 👨‍💻 👩‍💻)
 
+### Added 🚀
+
+-   Supports adding note to a custom view, provides a preview and an edit option for notes from the selection menu [#3234](https://github.com/MaibornWolff/codecharta/pull/3234)
```
ce-bo, line 12: A screenshot would be lovely here :)

visualization/app/codeCharta/ui/customConfigs/addCustomConfigButton/addCustomConfigDialog/addCustomConfigDialog.component.html
```diff
@@ -7,9 +7,22 @@ <h2>Add Custom View</h2>
...
+		<br />
+		<mat-form-field>
+			<label>Provide custom note for configuration (optional)</label>
+			<textarea
+				cdkAutosizeMinRows="8"
+				cdkTextareaAutosize
+				matInput
+				class="custom-config-input"
+				[(ngModel)]="customConfigNote"
+				required
```
ce-bo, line 22: Shouldn't it be optional?

visualization/app/codeCharta/ui/customConfigs/customConfigList/customConfigItemGroup/customConfigDescription/customConfigTruncateNote.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { CustomConfigTransformNote } from "./CustomConfigTransformNote.pipe"
+
+describe("customConfigTruncateNotePipe", () => {
+	it("should return empty placeholder sting if no string given", () => {
+		expect(new CustomConfigTransformNote().transform("", 0)).toBe("Add note")
+	})
+
+	it("should return a shortened string when given string above limit", () => {
+		expect(new CustomConfigTransformNote().transform("Shorten This", 5)).toBe("Short...")
```
ce-bo, line 9: Nit: this would be more readable
```suggestion
		expect(new CustomConfigTransformNote().transform("123456", 5)).toBe("Short...")
```

visualization/app/codeCharta/util/customConfigHelper.ts
```diff
@@ -139,7 +138,8 @@ export class CustomConfigHelper {
 				customConfigVersion: exportedConfig.customConfigVersion,
 				mapSelectionMode: exportedConfig.mapSelectionMode,
 				stateSettings: exportedConfig.stateSettings,
-				camera: exportedConfig.camera
+				camera: exportedConfig.camera,
+				note: exportedConfig.note
```
ce-bo, line 142: I think this could break if a config without the new "note" property is uploaded.
We should safely handle this case, if not already done.

ce-bo (commented):
The Custom View Items are a bit shifted in the UI. Whould you please have a look?
I think the Icons for the metrics should be at the same x position. So I guess the area on the left including the name and note should have a kind of fixed size.
image

ce-bo:
I migrated the angular legacy dialogs. Now the custom views are visible again but the list items are looking broken.
image

It could be the case that I did a mistake when merging the main branch.
Maybe someone can check this and continue here?

Hall-Ma:
> I migrated the angular legacy dialogs. Now the custom views are visible again but the list items are looking broken. image
> 
> It could be the case that I did a mistake when merging the main branch. Maybe someone can check this and continue here?

I will take over the issue.

visualization/app/codeCharta/util/customConfigHelper.ts
Hall-Ma: I have adjusted the 'note' property. Now it is optional and is only set when a note has been set by the user. We no longer have empty strings and backward compatibility problems.

phanlezz (approved):
Looks clean now! 
If we merge #3239 we could merge the truncate pipe into the new simplePipes module.

PR #3242, opened 2023-03-07
Refactor/3225/clean up dynamic margin

shaman-apprentice:
I noticed that we forgot to remove the callback for the dynamic margin checkbox which we removed within the area settings in https://github.com/MaibornWolff/codecharta/commit/22b2fd795acd974b54e6abb003338c0ce0a63965 (fst commit of this PR).

Afterwards `dynamicMargin` had no usage anymore. So I removed it (snd commit of this PR).

Custom configs which still have the `dynamicMargin` value still work.

ref #3225


visualization/app/codeCharta/ui/ribbonBar/areaSettingsPanel/areaSettingsPanel.component.html
```diff
@@ -11,7 +11,7 @@
 	<mat-checkbox (change)="setEnableFloorLabel($event)" [checked]="enableFloorLabels$ | async"> Enable Floor Labels </mat-checkbox>
 	<cc-invert-area-option></cc-invert-area-option>
 	<cc-reset-settings-button
-		[settingsKeys]="['appSettings.dynamicMargin']"
+		[settingsKeys]="['dynamicSettings.margin']"
```
shaman-apprentice, line 14: I wonder why we don't reset "invert area" and "enable floor labels" here as well? @ce-bo what do you think about a small FR for that?

ce-bo: Good catch @shaman-apprentice . That would be lovely. It is more like a bug because we forget to update this list. Would you be so kind and extend it?

ce-bo (approved):
LGTM! Thank you for cleaning this up 👍

Issue #3040, opened 2022-09-13
Bumpy animation of the map when moving


jannikr:
# Bug

## Expected Behavior

When dragging the map in different directions and zooming, a smooth animation should be performed.

## Actual Behavior

When dragging the map in different directions and zooming, a bumpy animation is performed. However, as soon as the cube is executed to move, a soft animation is possible (see video from the 8th second).

## Video

https://user-images.githubusercontent.com/31436472/189859803-8b10cbb1-328a-4db4-8486-bfc0985a30a6.mov

## Things to consider

- How much main memory is needed for a smooth animation? The Issue [2951](https://github.com/MaibornWolff/codecharta/issues/2951) has already addressed this issue.

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes
- CodeCharta Version: 1.105.0
- OS: Mac / Windows
- Browser: Chrome

PR #3244, opened 2023-03-10
Optimize animations by reducing debounce time

ce-bo:
# Optimize animations by reducing debounce time of...
- showing arrows and labels (on hovering a building)
- map movements 
- mouse wheel zooming

closes: #3040


Hall-Ma (commented):
It definitely runs more smoothly :)

Hall-Ma (approved):

PR #3248, opened 2023-03-14
Fix/improve custom views dialog styling

IhsenBouallegue:
# Improve custom views dialog styling

## Description

  - Fix custom views dialog styling and make it consistent with the others.
 
Note: I discovered that the expansion panel / accordion is not being styled correctly  from material UI. I couldn't find the reason as everything is imported correctly.
It should look like this per default:
image

## Screenshots or gifs
Before:
image
After:
image
image

### DIFF
--- a/visualization/app/codeCharta/ui/customConfigs/customConfigList/customConfigItemGroup/customConfigDescription/applyCustomConfigButton.component.html
+++ b/visualization/app/codeCharta/ui/customConfigs/customConfigList/customConfigItemGroup/customConfigDescription/applyCustomConfigButton.component.html
@@ -7,15 +7,20 @@
 	<p class="config-name" title="{{ customConfigItem.name }}">
 		<b>{{ customConfigItem.name }}</b>
 	</p>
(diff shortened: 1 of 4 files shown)


IhsenBouallegue:
I won't be available for the next 2 weeks, so if everything is green you can merge this! 😄

ce-bo:
For me it looks differently:
image

shaman-apprentice:
> For me it looks differently: image

For me it looks fine
image

When you reduce browser width so the dialog contents doesn't fit anymore into it, it looks like your pic. For me that is OK, what do you think? @ce-bo

Hall-Ma:
> > For me it looks differently: image
> 
> For me it looks fine image
> 
> When you reduce browser width so the dialog contents doesn't fit anymore into it, it looks like your pic. For me that is OK, what do you think? @ce-bo

I have fixed the problem. Before, it looked like Cedric's screenshot when you have the full width of the screen. Now it's more flexy, when you reduce the screen size :)

PR #3255, opened 2023-03-24
Fix/accurate testing of dialogs

Hall-Ma:
# Accurate testing of dialogs

## Description

- use Angular component harnesses to test the presence of dialogs in tests


shaman-apprentice (approved):
nice, LGTM

Issue #3262, opened 2023-03-30
Missing Icons for Suspicious Metrics Feature in ribbon bar


ce-bo:
# Bug
Missing Icons for Suspicious Metrics Feature in ribbon bar

## Screenshots
![image](https://user-images.githubusercontent.com/51205356/228859267-dcd1f4eb-988e-4d11-918f-7cca9360be95.png)

## Steps to Reproduce the Problem
1. Open current web demo

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes | no
- CodeCharta Version: 1.115.0
- OS: windows
- Browser: chrome 111

ce-bo:
Maybe it is related to this error:
![image](https://user-images.githubusercontent.com/26900540/228878562-d42d7ce3-7107-49a0-ad2e-3e3079995c08.png)

PR #3263, opened 2023-03-31
Fix/3262/Missing suspicious metrics feature

Hall-Ma:
# Fixing the disappearance of suspicious metric

Close: #3262

## Description


ce-bo (approved):
LGTM. You might want to add a regression test for that?

Issue #3261, opened 2023-03-30
Changelog is empty in visualization popup


phanlezz:
# Bug

When visiting or starting codecharta-visualization the popup changelog is empty.

## Expected Behavior

After a version change the popup (modal) should display everything from the changelog, it should show all changes since the last seen version.

## Actual Behavior

The modal just shows the version change, but no changelog information.

## Screenshots
![grafik](https://user-images.githubusercontent.com/65733509/228837265-74d60253-d641-45fd-89aa-3ebb527e12eb.png)

## Steps to Reproduce the Problem
1. Visit/Open Visualization

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes
- CodeCharta Version: 1.115.0
- OS: doesn't matter
- Browser: firefox (doesn't matter)

PR #3264, opened 2023-03-31
Fix/3261/Changelog Dialog

Hall-Ma:
# Fix disappearance of changelog entries in dialog

Close: #3261

## Description

  - ensures that new lines are recognized when the change log is read in and that a new array entry is added per line.
  - to ensure that the changelog dialogue renders content, an e2e test was created

### DIFF
--- a/visualization/app/codeCharta/ui/dialogs/changelogDialog/changelogDialog.component.ts
+++ b/visualization/app/codeCharta/ui/dialogs/changelogDialog/changelogDialog.component.ts
@@ -11,7 +11,8 @@ export class ChangelogDialogComponent {
 	changes: Record<string, string>
 
 	constructor(@Inject(MAT_DIALOG_DATA) public data: { previousVersion: string; currentVersion: string }) {
-		let changelogLines = markdownFile.split("\n")
+		const startNewLinePattern = /(?<=>)\s+(?=<)/g
+		let changelogLines = markdownFile.split(startNewLinePattern).map(line => line.trim())
 		const currentVersionFirstLine = this.findVersionLine(changelogLines, this.data.currentVersion)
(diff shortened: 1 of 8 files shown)


visualization/app/codeCharta/ui/dialogs/changelogDialog/changelogDialog.component.ts
```diff
@@ -11,7 +11,8 @@ export class ChangelogDialogComponent {
 	changes: Record<string, string>
 
 	constructor(@Inject(MAT_DIALOG_DATA) public data: { previousVersion: string; currentVersion: string }) {
-		let changelogLines = markdownFile.split("\n")
+		const newLineStartPattern = /(?<=>)\s+(?=<)/g
```
ce-bo, line 14: I am not certain if the regex is correct. I checked it by regex101. Can you please confirm this?

ce-bo (approved):

visualization/app/codeCharta/ui/dialogs/changelogDialog/changelogDialog.component.ts
BridgeAR: We should indeed split the lines similar to before. I guess we might also remove potential carriage returns.

visualization/package.json
```diff
@@ -128,8 +128,8 @@
 		"lint-staged": "^13.0.0",
 		"nw-builder": "^3.7.4",
 		"puppeteer": "^17.0.0",
-		"remark-html": "^13.0.1",
-		"remark-loader": "^4.0.0",
+		"raw-loader": "^4.0.2",
```
ce-bo, line 131: Raw loader seems to be deprecated soon.
"DEPREACTED for v5: please consider migrating to [asset modules](https://webpack.js.org/guides/asset-modules/)."

Hall-Ma: I replace it.

Issue #3298, opened 2023-05-15
MergeFilter gives no hint if selected input does not exist


moritz-suckow:
# Bug

## Expected Behavior

WHEN

I input non-existing files or folders into the MergeFilter

THEN

I want the shell to notify me about the non-existing ressources with an appropriate error message.

## Actual Behavior

When a non existing folder is entered, an error message is shown.
If the shell did not find atleast two cc.json files, a warning will be displayed that no nodes were merged.
When multiple files are entered, with atleast two existing and valid cc.json files, no error is displayed.

moritz-suckow:
Should the execution be cancelled if atleast one of the provided files does not exist?

PR #3305, opened 2023-05-16
fix/3298/output-error-for-nonexistent-file-params

moritz-suckow:
# Output error for each nonexistent file specified

Issue: #3298 

## Description

This PR adds an error message which is always output for each specified file parameter which is nonexistent.
In this PR this will only be implemented for the MergeFilter, all other parsers/filters will be added in another PR with the issue #3307 
Edit: We will abort execution if one or more of the specified files is invalid.


moritz-suckow:
This basically does very similar things compared to the new #3307, but only for the MergeFilter. Therefore to not create additional refactoring work I tried to implement this here, like I would have done in #3307, so this is basically implementing #3307 for the MergeFilter.

ce-bo (approved):
LGTM but let's add a changelog entry.

phanlezz (approved):
LGTM 🚀

Issue #3274, opened 2023-04-20
Use current working directory as default path for file input


moritz-suckow-mw:
# Feature request
## Description

As an Auditor, I want to enter an empty string when trying to analyze the current directory so that I do not need to manually find out and enter the path myself. 

## Acceptance criteria
- When using the interactive shell
  - the default for input path is the current directory
  - the users full/absolute current working directory is shown as a hint' in the shell when the question is shown to the user, so that they can accept it by pressing enter
  - This should be implemented for all applicable parsers (but this may apply only for certain parsers)
  - If a parser expects only a path to a single file, then just a file path example is given as a hint but no directory path as default.

ce-bo:
@moritz-suckow are there any questions?

Ideally, a hint is shown that the current working directory is used as the default when not specifiying any path.

moritz-suckow:
Do we implement this for all parsers who can receive a directory as input or just the RawTextParser?

moritz-suckow:
Additionally I had some doubts if this made sense to implement since this could introduce some problems, because of how the interactive shell currently works with copy-pasted input. It sometimes happens, that when you paste something into the shell that the input is automatically entered, sometimes even the input after the current question. What this could mean is, that if you do not watch out you could get your current working directory completely parsed.

moritz-suckow:
I updated the issue to include the decision from the weekly!

PR #3319, opened 2023-05-26
Feature/3274/use-current-directory-if-input-empty

moritz-suckow:
# Use current working directory if input is empty for parsers

Issue: #3274 

## Description

This adds a new default value to all parsers where a folder can be input to be parsed. This default value will be the current working directory.


analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import com.github.kinquirer.KInquirer
+import com.github.kinquirer.components.promptInput
+import java.io.File
+import java.nio.file.Paths
+
+class ParserDialogHelper {
+
+    companion object {
+        fun getInputFiles(isSourceMonitor: Boolean): MutableList<String> {
```
moritz-suckow, line 11: I just made this class to avoid some duplication, only thing new here is the KInquirer hint.

analysis/export/CSVExporter/src/main/kotlin/de/maibornwolff/codecharta/exporter/csv/ParserDialog.kt
```diff
@@ -5,12 +5,14 @@ import com.github.kinquirer.components.promptInput
...
 import de.maibornwolff.codecharta.tools.interactiveparser.ParserDialogInterface
 import java.math.BigDecimal
+import java.nio.file.Paths
 
 class ParserDialog {
     companion object : ParserDialogInterface {
         override fun collectParserArgs(): List<String> {
             val inputFileName = KInquirer.promptInput(
                 message = "What is the cc.json file that has to be parsed?",
+                    hint = Paths.get("").toAbsolutePath().toString()
```
ce-bo, line 15: Is  this misleading because a cc.json is requested instead of a directory path? Or is the directory scanned for cc.json if no cc.json is specified in the path?

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/ParserDialog.kt
```diff
@@ -13,7 +14,9 @@ class ParserDialog {
 
         override fun collectParserArgs(): List<String> {
 
-            val inputFileName = KInquirer.promptInput(message = "What is the $EXTENSION file that has to be parsed?")
+            val inputFileName = KInquirer.promptInput(
+                    message = "What is the $EXTENSION file that has to be parsed?",
+                    hint = Paths.get("").toAbsolutePath().toString())
```
ce-bo, line 19: Here as well

analysis/filter/StructureModifier/src/main/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/ParserDialog.kt
```diff
@@ -5,13 +5,16 @@ import com.github.kinquirer.components.promptInput
...
 
 class ParserDialog {
     companion object : ParserDialogInterface {
 
         override fun collectParserArgs(): List<String> {
             val inputFolderName =
-                    KInquirer.promptInput(message = "What is the cc.json file that has to be modified?")
+                    KInquirer.promptInput(
+                            message = "What is the cc.json file that has to be modified?",
+                            hint = Paths.get("").toAbsolutePath().toString())
```
ce-bo, line 17: Same here.

analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
```diff
@@ -0,0 +1,30 @@
...
+            val inputFileNames = mutableListOf(KInquirer.promptInput(
+                    message = if (isSourceMonitor) { "What is the sourcemonitor CSV file that has to be parsed?" } else { "Please specify the name of the first sourcemonitor CSV file to be parsed" },
```
ce-bo, line 13: The if condition does not makes sence, in both cases "sourcemonitor" is mentioned.

analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
```diff
@@ -0,0 +1,30 @@
...
+                    hint = Paths.get("").toAbsolutePath().toString() + File.separator + "yourInput.csv"))
+
+            while (true) {
+                val additionalFile = KInquirer.promptInput(
+                        message = "If you want to parse additional sourcemonitor CSV files, specify the name of the next file. Otherwise, leave this field empty to skip.",
```
ce-bo, line 18: Here the if condition for type sourcemonitor is missing.

analysis/export/CSVExporter/src/main/kotlin/de/maibornwolff/codecharta/exporter/csv/ParserDialog.kt
moritz-suckow: You are absolutely right, in some other dialogs I added this, but here I forgot!

analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
moritz-suckow: As in my other comment, previously we asked this question for the CSVImporter and the CSVSourceMonitorImporter. I can adjust this so it is neutral, and rather ask for csv files so it fits both or ask an unique question for both. What do you think?

analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
moritz-suckow: I merged the input from the CSVImporter and CSVSourceMonitorImporter here, those were the dialogs as they were before. But I agree with you, I will adjust those questions.

moritz-suckow:
Should I also adjust the directory question which is asked for the parser suggestions? Currently an empty input is mapped to current directory, but I find the solution here to be better (setting default value to current directory instead of just mapping it).

analysis/import/CSVImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/util/ParserDialogHelper.kt
ce-bo: I like that idea!

ce-bo:
> Should I also adjust the directory question which is asked for the parser suggestions? Currently an empty input is mapped to current directory, but I find the solution here to be better (setting default value to current directory instead of just mapping it).

That would be a nice improvement, thank you.

ce-bo (approved):
LGTM :-) Maybe @phanlezz could overtake a short test of the new feature, I am not sure, when I can do this at the moment. Maybe you can present your changes @moritz-suckow?

phanlezz:
I did some testing and it worked for me :) 
We could remove the hint from the `gitlogparser`, that an empty input would be equal to the current working directory, as the current working directory is already inserted as a path. The underlying functionality that actually then puts a "`.`" as input could be removed as well, because it might be the only parser that does this. 
But this is optional, as it is still a feature, but just not really required anymore. But I might missed a use-case here.

moritz-suckow:
I removed the hint and conversion from the GitLogParser, for now it just aborts when there is no input. This will change soon anyways when #3325 is merged, so I did not put any further work in it.

Issue #3307, opened 2023-05-17
Detect invalid file input


moritz-suckow:
# Feature request

## Description

As an Auditor, I want to be warned if the files I specified for input do not exist so that I can adjust the files names or correct paths.

## Acceptance criteria
- For each parser with files as input, if one or more input file do not exist, execution is stopped and an error message is output.
  - The same applies to invalid file paths
- Execution is also stopped if the user does not provide any file paths (if applicable)
- Execution is also stopped if the input file paths do not provide any files
- For each parser, with only files and no folders as input, execution is stopped and an error message is output if user specifies a folder as input.

## Development notes (optional Task Breakdown)
Since this is part of many parsers and to avoid duplication and inconsistent behaviour, we should solve this in one place for all parsers like the model.

ukinimod:
I also encountered this for gitlogparser, e.g. with specifying "--repo-path" for a repo path that doesn't exist.

phanlezz:
While it is important to stop the execution of any parser/filter/etc. when a users provides a **wrong** file/-path;
 we need to **continue** if a users does not provide any `0` input files, because the input could be piped in. This may only applicable to certain parsers.

ce-bo:
@moritz-suckow I guess this can be closed.

PR #3325, opened 2023-05-30
Feature/3307/detect-invalid-file-input

moritz-suckow:
# Detect invalid file input

Issue: #3307 

## Description

This PR adds validation for file input for all parsers, by checking a row of conditions (existence, non-emptiness, ..., for details look at the specification in the issue).


analysis/import/SourceCodeParser/src/test/kotlin/de/maibornwolff/codecharta/importer/sourcecodeparser/e2e/FileDoesNotExist.kt
moritz-suckow, line 1: Deleted because that should now be tested in the main test

analysis/import/MetricGardenerImporter/build.gradle
```diff
@@ -1,6 +1,5 @@
 dependencies {
     implementation project(':tools:InteractiveParser')
-    implementation project(':import:SourceCodeParser')
```
moritz-suckow, line 3: I dont know why MetricGardener was dependent on SourceCodeParser, I removed that. Since nothing broke I assume this was just an unneeded leftover?

analysis/filter/StructureModifier/src/test/kotlin/de/maibornwolff/codecharta/filter/structuremodifier/StructureModifierTest.kt
```diff
@@ -38,6 +54,7 @@ class StructureModifierTest {
         assertThat(errorStream.toString()).contains("invalid_project.cc.json is not a valid project")
     }
 
+    // TODO: Why is this disabled? Should it not be removed?
```
moritz-suckow, line 57: I would remove this disabled test, since the last time it was touched was in 2020

analysis/filter/EdgeFilter/src/main/kotlin/de/maibornwolff/codecharta/filter/edgefilter/EdgeFilter.kt
```diff
@@ -22,8 +24,8 @@ class EdgeFilter(
     @CommandLine.Option(names = ["-h", "--help"], usageHelp = true, description = ["displays this help and exits"])
     var help: Boolean = false
 
-    @CommandLine.Parameters(arity = "1..*", paramLabel = "FILE", description = ["files to filter"])
```
moritz-suckow, line 25: I am not sure if this was the correct arity, since from my understanding we only ever wanted to use one file with the EdgeFilter.

moritz-suckow:
Regarding the code smell from Sonar: I would ignore it. Moving the return out of a double nested if, is in my opinion, not good for the code's readability.

moritz-suckow:
Currently this feature is implemented in a way, that invalid resources can only be detected when running a command. That means when using the interactive parser or parser suggestions, the user first answers all questions to configure the parser, the parser/s start/s running and THEN the user gets warned if files/folders do not exist or are invalid.

Is that okay for now or should this PR also implement a warning when configuring?

phanlezz (approved):
LGTM

moritz-suckow:
Maybe after changing the action following invalid input from just printing a message and returning to throwing an exception we should also adapt the tests to check if an exception gets thrown?

Issue #3327, opened 2023-05-31
Parser Suggestions very slow


moritz-suckow:
# Bug

## Expected Behavior

WHEN

I use the `ccsh` without any arguments and enter the path for the repository I want to analyse

THEN

I want to receive parser suggestions without waiting a lot of time.

## Actual Behavior

The parser suggestions take several seconds before they popup, even for just a medium sized project like CodeCharta. Furthermore, no hint is shown indicating that something is happening and the user might be wondering if the application froze.

## Specifications
- CodeCharta Version: 1.117.0
- OS: All

PR #3329, opened 2023-06-02
fix/3327/fix-parser-suggestions-slow

moritz-suckow:
# Speed up parser suggestions

Issue: #3327 

## Description

This PR aims to make the parser suggestions less slow and make the user aware that the shell did not crash when applicability is calculated.
To do this, some messages are logged when each parser finished checking for applicability and which folders are scanned for applicability.

Additionally some optimizations are done to increase the checking speed, especially for the MetricGardener.


analysis/model/src/main/kotlin/de/maibornwolff/codecharta/util/ResourceSearchHelper.kt
```diff
@@ -6,26 +6,42 @@ import java.nio.file.Paths
 class ResourceSearchHelper {
 
     companion object {
-        fun isResourcePresent(resourceName: String, searchToken: String, searchOperator: (String, String) -> Boolean,
+
+        /**
+         * Checks whether a resource contains one of the specified searchTokens. To check that, the given searchOperator is used.
+         *
+         */
+        fun isResourcePresent(resourceName: String, searchToken: List<String>, searchOperator: (String, List<String>) -> Boolean,
```
ce-bo, line 14: The comment makes clear, that the method name is not able alone to express what it is doing.

analysis/import/SonarImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/sonar/SonarImporterMain.kt
```diff
@@ -96,13 +96,15 @@ class SonarImporterMain(
...
         val trimmedInput = resourceToBeParsed.trim()
 
         if (trimmedInput.contains("^http(s)?://".toRegex())) {
             return true
         }
 
-        return ResourceSearchHelper.isResourcePresent(trimmedInput, "sonar-project.properties",
+        return ResourceSearchHelper.isResourcePresent(trimmedInput, listOf("sonar-project.properties"),
                 ResourceSearchHelper::doStringsEqual, 0,
                 shouldSearchFullDirectory = true, resourceShouldBeFile = true)
```
ce-bo, line 109: I think it is not necessary to search the full directory since the file should be created in the root directory of a project.

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/util/ResourceSearchHelper.kt
```diff
@@ -6,26 +6,42 @@ import java.nio.file.Paths
...
+                if (resource.endsWith(token)) {
+                    return true
+                }
+            }
+            return false
         }
 
-        fun doStringsEqual(string1: String, string2: String): Boolean {
-            return (string1 == string2)
+        fun doStringsEqual(resource: String, searchToken: List<String>): Boolean {
```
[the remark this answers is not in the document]
moritz-suckow: I was writing generic functions to pass into `isResourcePresent`. I will adjust the naming in the whole class a bit :)

analysis/import/SonarImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/sonar/SonarImporterMain.kt
moritz-suckow: Theoretically it is, in some cases (like CodeCharta), the sonar-properties file is not in the root folder. But I agree that we should probably only search the given directory and no subdirectories since this case is rather rare and special.

ce-bo: Well, another option would be to search with a maxDepth of 3 for example that should be enough I think.

ce-bo (approved):
LGTM! Though, added one a small remark.

analysis/import/SonarImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/sonar/SonarImporterMain.kt
moritz-suckow: I would say we can search at a depth of 2 with the following reasoning:
Normally a project should have one sonar project file at root directory level, but as in CodeChartas case, the project is split into several smaller projects. 
In that case, the smaller projects should be at the second level, and so any sonar properties file should also be there. I do not think it is realistic to consider projects with 2 additional layers of sub-projects with sonar projects for the lowest level each.

moritz-suckow:
I refactored the ResourceSearchHelper because it was a "one class to fit all needs" implementation and was only tested by its users.

What I did was split the main function up into two separate functions, remove special cases which were only needed in edge cases for some of the using classes, add some additional validation to cover edge cases which were previously not thought of and add own unit testing for the ResourceSearchHelper.

Additionally I removed some unneeded operations which further sped up the parser suggestions.

analysis/import/SonarImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/sonar/SonarImporterMain.kt
ce-bo: We do not know if this assumption is correct in the most cases but it could be valid.
So let's go with a maxDepth of 2 for now.

analysis/import/SonarImporter/src/main/kotlin/de/maibornwolff/codecharta/importer/sonar/SonarImporterMain.kt
```diff
@@ -96,15 +96,23 @@ class SonarImporterMain(
...
 
-        if (trimmedInput.contains("^http(s)?://".toRegex())) {
+        val trimmedInput = resourceToBeParsed.trim()
+        val inputFile = File(trimmedInput)
+        if (trimmedInput.contains("^http(s)?://".toRegex()) || (inputFile.isFile && inputFile.name == searchFile)) {
             return true
         }
+        if (!inputFile.isDirectory || trimmedInput == "") {
+            return false
+        }
```
ce-bo, line 109: I would recommend to combine these if-statements.
One Idea would be to have a private method for `(inputFile.isFile && inputFile.name == searchFile)` like `isSonarPropertiesFile(...)`. That would make it more readable.
Another Example: `isUrl()`

analysis/model/src/main/kotlin/de/maibornwolff/codecharta/util/ResourceSearchHelper.kt
```diff
@@ -5,53 +5,62 @@ import java.io.File
...
-                        .filter { it.isFile }
-                        .map { it.name }
-                        .filter { searchOperator(it, searchToken) }
-                        .any()
-            } else {
-                fileSearch.asSequence()
-                        .map { it.name }
-                        .filter { searchOperator(it, searchToken) }
-                        .any()
+        private fun doesInputFileNameEndWithOneOrMoreOfEndings(inputFile: File, fileEndings: List<String>): Boolean {
```
ce-bo, line 57: I would rather prefer this method signature:
`endsWithAtLeastOne(inputString: String, endings: List<String>)` It is more generic and not related to a file object, thus it has a better reusability.

ce-bo (approved):
LGTM.

Issue #3326, opened 2023-05-31
Running a particular parser interactively is not working anymore


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN
running a particular parser interactively by calling for instance `ccsh sonarimport`

THEN
an error message is shown complaining about missing arguments

## Actual Behavior
the interactive dialog should start.

## Specifications
- CodeCharta Version: 1.117.0
- OS: all
- Browser: all

PR #3332, opened 2023-06-06
fix/3326/fix-directly-run-interactive-parser

moritz-suckow:
# Add automatic running of interactive parser when no args are specified

Issue: #3326 

## Description

This adds automatic running of the interactive parser dialog if you call a specific parser but with no arguments, e.g. `ccsh sonarimport`.

## Note

Would it make sense to also add this whenever insufficient arguments are provided and not only none?


phanlezz:
I have some trouble to get it running on my local windows machines. I need to investigate before merging

ce-bo:
Few ideas/points to check:
- do you have enough disk space?
- the line endings are different between the two screenshots
- Are you able to use the GitLogParser in the main branch to parse another branch?
- How much bytes are written to this temp file? Are there limitations of the used FileWriter approach like in terms size limitations?

Issue #3320, opened 2023-05-26
Add parser name before configuration start in multi parser usage


moritz-suckow:
# Feature request

## Description

As an Auditor, I want to directly know which parser I am configuring while using the multi parser configuration setup so that this feature is more easily usable. Right now, you have to look at the order of parsers during selection to guess which parser you are currently configuring.

## Acceptance criteria
- Before the interactive configuration of a parser starts during parser suggestions, the name of the parser the user is about to configure is printed

phanlezz:
Closed by #3335

PR #3335, opened 2023-06-09
feature/3320/print-parser-before-config

moritz-suckow:
# Output which parser is currently being configured

Issue: #3320 

## Description

This PR adds a message telling which parser is being configured before the interactive dialog of each parser that was chosen.


phanlezz (approved):
LGTM

phanlezz:
Auto-merge made a mistake in the changelog.

PR #3354, opened 2023-08-01
Fix ccsh cli problems by providings custom bin-links for the gradle executables #2600

### LINKED ISSUE #2600: Broken codecharta-anlysis npm package / ccsh command
# Bug

Open cmd.exe
>npm i codecharta-analysis@1.87.0 -g
> ccsh --help

not working.
But Powershell works as expected.

cmd.exe works for 1.86.0

## Screenshots
image

1. Install version 1.87.0 of npm package codecharta-analysis via cmd.exe
1. Enter `ccsh --help`

##
Applies only for Windows cmd.exe
Powershell works as expected

phanlezz:
# Provide custom bin-links for gradle executables

Closes #2600

## Description
This PR provides a set of two bin-links scripts. Each sets contains three scripts (bin-links) that then call the correct bin script to execute ccsh. There are two ccsh-scripts produced by gradle, one for unix and one for windows, that need to be callled / linked in different ways. 
Npm generates three links to those gradle scripts on package install. Sadly, npm does not provide any native method to change the behavior of those scripts, so this PR introduces custom scripts / bin-links. Those scripts are modified versions of the generated links.
Via a postinstall lifecycle call the modified links get copied to the npm directory. The location varies for a local or a global install.

The original issue was probably caused by broken java paths, because either the environment expected windows paths (`\`) and got unix versions (`/`) or vice-versa. 
This fix enables direct and native execution of those gradle scripts, without any node calls or nested executions.


analysis/node-wrapper/binScripts/global/ccsh.ps1
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env pwsh
+$basedir=Split-Path $MyInvocation.MyCommand.Definition -Parent
+
+$exe=""
```
ce-bo, line 4: Not sure about $exe because it seems not to be in use.

ce-bo (approved):
LGTM!

analysis/node-wrapper/install.js
```diff
@@ -0,0 +1,34 @@
...
+							res();
+						})
+					} catch (err) {
+						console.error(err);
+						rej(err);
+					}
+				})
+			)
+		})
+		await Promise.allSettled(deletionPromises);
```
BridgeAR, line 32:
```suggestion
    await Promise.allSettled(bins.map(asnyc (file) => {
      await fs.promises.copyFile(path.join(scriptLocation, file), path.join(scriptTarget, file));
      console.log(`${file} was moved`);
    }));
```

BridgeAR: Files that were not moved are ignored?

phanlezz: I'm not sure why, but the catch did not catch the exception and the process was terminated correctly as it is intended. It seems like the callback was not affected by the catch. During testing the install process would be terminated if there was an error in the file copy call, which is the behavior I wanted for the script. 

Do you have a suggestion for this segment, that includes an error throw?

analysis/node-wrapper/package.json
```diff
@@ -13,13 +13,15 @@
...
-		"prepare": "rm -rf ../build/install/codecharta-analysis public && ../gradlew -p .. installDist && cp -R ../build/install/codecharta-analysis public"
+		"prepare": "rm -rf ../build/install/codecharta-analysis public && ../gradlew -p .. installDist && cp -R ../build/install/codecharta-analysis public",
+		"postinstall": "node install.js"
 	},
 	"bin": {
 		"ccsh": "public/bin/ccsh"
 	},
 	"license": "BSD-3-Clause",
 	"dependencies": {
+		"is-windows": "^1.0.2",
```
BridgeAR, line 24: Please just check `process.platform === 'win32'`. The library is not maintained and not needed.

analysis/node-wrapper/install.js
```diff
@@ -0,0 +1,26 @@
...
+        : path.join(process.env.npm_config_local_prefix, "node_modules/.bin");
+
+      await Promise.all(bins.map((async (file) => {
+        try {
+          await fs.promises.copyFile(path.join(scriptLocation, file), path.join(scriptTarget, file));
+          console.log(`${file} was moved to ${scriptTarget}`);
+        } catch (err) {
+          console.error("File could not get copied")
+          throw err
+        }
```
BridgeAR, line 23:
```suggestion
        await fs.promises.copyFile(path.join(scriptLocation, file), path.join(scriptTarget, file));
        console.log(`${file} was moved to ${scriptTarget}`);
```

Using try / catch like this is an anti-pattern. The error itself already contains the necessary information and it will show up when the code fails. Please be aware that the script is going to fail as soon as a single error is coming up.

analysis/node-wrapper/install.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
+
+const fs = require("fs");
+const path = require("path");
+var isWindows = require("is-windows");
+
+if (isWindows() === true && process.env.npm_package_name === "codecharta-analysis") {
```
BridgeAR, line 7:
```suggestion

if (process.platform === "win32") {
```

This should only be executed in the analysis anyway, no?

And instead of indenting all the code, it's possible to use an early return:

```js
if (process.platform !== 'win32') {
  return;
}
// Other code
```

phanlezz: > This should only be executed in the analysis anyway, no?

The name check is intended to verify if it is installed standalone and not as a dependency. But I am not sure if this works in that case.

analysis/node-wrapper/install.js
phanlezz: Just one more thing:
Why `await` the file copy? Doesn't this defeat the whole async  point? Isn't the `await Promise.all([])` enough to wait for all results?

BridgeAR: That depends, if you want to log that a file was moved or not. It's possible to directly return in case you don't want to log it.

Issue #3357, opened 2023-08-03
Sonar not working with old Java versions


BridgeAR:
> warning The version of Java (11.0.14) you have used to run this analysis is deprecated and we will stop accepting it soon. Please update to at least Java 17.
Read more [here](https://docs.sonarcloud.io/appendices/scanner-environment/)

https://github.com/MaibornWolff/codecharta/pull/3353#issuecomment-1663900072

phanlezz:
Closed via #3359 #3370

PR #3359, opened 2023-08-11
Setup SonarCloud Scan Enviroment to use Java 17+

phanlezz:
# Setup SonarCloud Scan Enviroment to use Java 17+

The linked [docs](https://docs.sonarcloud.io/appendices/scanner-environment/) state, that there should be no need for  additional setup while using the github action. We might need to update the github action, as it wasn't updated in a while.

Additionally, the [docs](https://docs.sonarcloud.io/appendices/scanner-environment/) state, that we should not use the github action and use the gradle plugin instead. 

Issue: #3357


phanlezz:
Gradle Sonarqube doesn't work with Java 8, so I will switch back to the github action to decouple the version requirements.

analysis/build.gradle
```diff
@@ -93,14 +94,22 @@ subprojects {
...
+        jvmToolchain(11)
     }
 
     kapt {
         arguments {
             arg("project", "${project.group}/${project.name}")
         }
     }
+
+    if("${project.name}".contains("RawTextParser")) {
```
ce-bo, line 106: Why is the path differnt for this parsre?

analysis/gradle.properties
```diff
@@ -9,3 +9,11 @@ org.gradle.parallel=true
 # Only relevant projects are configured which results in faster builds for large multi-projects.
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:configuration_on_demand
 org.gradle.configureondemand=true
+
+systemProp.sonar.organization=maibornwolff-gmbh
+systemProp.sonar.projectKey=maibornwolff-gmbh_codecharta_analysis
+systemProp.sonar.projectName=CodeCharta Analysis
+systemProp.sonar.host.url=https://sonarcloud.io
+systemProp.sonar.links.ci=https://travis-ci.org/MaibornWolff/codecharta
```
ce-bo, line 17: This link is outdated

.github/workflows/test.yml
```diff
@@ -178,7 +183,7 @@ jobs:
 
       - name: SonarCloud Scan
         if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
-        uses: SonarSource/sonarcloud-github-action@v1.6
+        uses: SonarSource/sonarcloud-github-action@v2.0.1
```
ce-bo, line 186: According to the docs we should consider switching this to the SonarQube Plugin for Gradle.

ce-bo (commented):
In general LGTM. We should discuss what kind of version update we wanna have due to the Java 11 Upgrade.

.github/workflows/test.yml
phanlezz: That is the sonar action for visualization, visualization is not managed by gradle, so this is the correct way.

analysis/build.gradle
phanlezz: There is an error during the sonar analysis run, which states, that it can't find the test results from the RawTextParser subproject. I am trying different methods of configuration, but none seem to make a difference

phanlezz: This didn't work either, so it will be changed next push.

analysis/gradle.properties
phanlezz: Good catch, I copied those settting from the original config. Will be removed next push.

phanlezz:
> In general LGTM. We should discuss what kind of version update we wanna have due to the Java 11 Upgrade.

For the next release? Minor version would be fitting.

ce-bo (approved):

PR #3378, opened 2023-10-13
Remove release information 1.120

phanlezz:
# Unrelease version 1.120

Due to issues with new packager, we will unrelease version 1.120 until the issue is fixed.


ce-bo (approved):
LGTM.

Issue #3338, opened 2023-06-14
Parser suggestions - Improvements


ce-bo:
# Feature request

## Description

As a User of the Parser Suggestions (running ccsh command without arguments):
- [x] I want not to be asked to merge output files, if only one parser was executed
- [x] I want to see a description of the individual parser that was suggested to me in this dialog:
  - ![image](https://github.com/MaibornWolff/codecharta/assets/26900540/79cbf34a-81ec-48a2-9578-489d2f6337ab)
- [x] I want to see the full path of the file(s) that has/have been created at the end

## Developer Notes
- Open a separate PR for the individual points.

phanlezz:
If I selected >no< before the execution, it would be nice if the command line equal call would be printed out, so that it can be copied for later.
(only if not too complicated)

PR #3384, opened 2023-10-26
feature/3338/improve-parser-suggestion-1

moritz-suckow:
# No longer ask to merge when only executing one parser

Issue: #3338

## Description

This PR simply adds a check to only ask for and merge output from parser suggestions, when more than one parser was executed. Additionally I fixed a few wrong usages of the KotlinLogger and also refactored the CCSH Unit-Tests because there was a whole lot of duplication with mocking which made these tests very confusing to understand.

### DIFF
--- a/analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
+++ b/analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
@@ -137,14 +137,14 @@ class Ccsh : Callable<Void?> {
             if (finalExitCode != 0) {
                 return finalExitCode
             }
-            // Improvement: Try to extract merge commands before so user does not have to configure merge args?
             if (configuredParsers.size == 1) {
                 logger.info { "Parser was successfully executed and created a cc.json file." }
                 return 0
-            } else {
-                logger.info { "Each parser was successfully executed and created a cc.json file." }
-                return askAndMergeResults(commandLine)
             }
+
+            // Improvement: Try to extract merge commands before so user does not have to configure merge args?
+            logger.info { "Each parser was successfully executed and created a cc.json file." }
+            return askAndMergeResults(commandLine)
         }
 
         private fun askAndMergeResults(commandLine: CommandLine): Int {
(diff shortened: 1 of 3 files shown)


analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
```diff
@@ -125,21 +125,26 @@ class Ccsh : Callable<Void?> {
...
             val finalExitCode = exitCode.get()
-            logger.info("Code: $finalExitCode")
+            logger.info { "Code: $finalExitCode" }
             if (finalExitCode != 0) {
                 return finalExitCode
             }
             // Improvement: Try to extract merge commands before so user does not have to configure merge args?
-            logger.info("Each parser was successfully executed and created a cc.json file.")
-            return askAndMergeResults(commandLine)
+            return if (configuredParsers.size == 1) {
```
Nereboss, line 141: I would find it easier to read if the return statement was inside of the if and else parts. The way it is currently it can potentially be confused with returning a function that consists if the if-else statement

Nereboss (commented):
Everything except my one comment is fine. 
It is to note that the tests only work when running with gradle and not with IntelliJs solution.

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
moritz-suckow: While I totally agree with you, I think this is the current standard way of doing it configured in Codecharta. If I moved the return statement into the if, I believe Sonar would throw a codesmell and ask me to move the return out of the if again which then might lead to us not reaching the quality gate.

What we should probably do is discuss to disable that rule since I think everyone on the team does not like that style :)

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/Ccsh.kt
```diff
@@ -125,21 +125,26 @@ class Ccsh : Callable<Void?> {
...
             // Improvement: Try to extract merge commands before so user does not have to configure merge args?
-            logger.info("Each parser was successfully executed and created a cc.json file.")
-            return askAndMergeResults(commandLine)
+            if (configuredParsers.size == 1) {
+                logger.info { "Parser was successfully executed and created a cc.json file." }
+                return 0
+            } else {
+                logger.info { "Each parser was successfully executed and created a cc.json file." }
+                return askAndMergeResults(commandLine)
+            }
```
ce-bo, line 147: no else block needed

ce-bo (approved):
LGTM

PR #3387, opened 2023-10-30
Feature/3338/improve parser suggestion 2

moritz-suckow:
# Add descriptions to suggested Parsers

Issue: #3338 

## Description

This PR adds the description of each parser to the list of suggested parsers which currently only contains the names of the suggested parsers. Additionally I refactored two things: 
First I moved the companion object of each of the main classes of each parser to the top since they were at random places in each file. 
Secondly I moved the name and description of each parser from the old centralized file into each parser itself to reduce coupling between the modules.


phanlezz (approved):
LGTM! 🚀

Issue #3253, opened 2023-03-21
Cannot install on mac m1


xndlnk:
# Bug

## Expected Behavior

WHEN

I install the Desktop Visualization via `npm i codecharta-visualization` locally

THEN

the application should install

## Actual Behavior

I get the error message:

npm ERR! code 1
npm ERR! path /Users/xxx/Projects/toys/codecharta/node_modules/codecharta-visualization/node_modules/nw
npm ERR! command failed
npm ERR! command sh -c -- node scripts/install.js
npm ERR! Response code 404 (Not Found)

## Steps to Reproduce the Problem
1. `npm i codecharta-visualization`

## Specifications
- CodeCharta Version: 1.114.0
- OS: MacOS 13.2.1 running on an Apple M1 Pro
- Browser: Chrome Version 111.0.5563.64

## Workaround

- I cloned your repository
- then changed dir to `./visualization`
- when I try to `npm i` the same error from above occurs
- then run `npm_config_nwjs_process_arch=x64 npm install nw` as described here https://github.com/nwjs/npm-installer/issues/83
- then run `npm i` and installation succeeds

Maybe this helps to find a solution.

phanlezz:
Thanks for the bug report! We are aware of this issue with NW. There is already an open [PR](https://github.com/nwjs/npm-installer/pull/84), that they need to merge to fix this issue.
I don't think that there is an elegant way for us to include this workaround, because it is an installation error.
I will move the workaround instructions from the [Build-it-yourself](https://github.com/MaibornWolff/codecharta/blob/main/visualization/README.md#build-it-yourself) section further up, so that they are more prominently displayed.

NearW:
https://github.com/MaibornWolff/codecharta/blob/main/visualization/script/build-nwjs.js

Update platforms from:
`platforms: ["osx64", "win32", "win64", "linux32", "linux64"]`
to
`platforms: ["osx64", "osarm64", "win32", "win64", "linux32", "linux64"]`

phanlezz:
To include an arm build we need to switch to the next major version of nw-builder. 
Sadly, the next major version is not yet ready for production, because there are several issues, which renders it not applicable for codecharta. 
As soon as it is in a more stable state for building we will update our package script.

NearW:
Electron suppports arm64. We can also think about switching to electron..

phanlezz:
Closed via #3373 #3389

PR #3389, opened 2023-10-30
Fix/3253/electron build

phanlezz:
# Electron Build Scripts

This adds a set of custom build scripts, which download, build and package the newly included electron desktop app for various platforms. Also includes `--local` flag to enable a build only for the developer's current OS & architecture combination. 

Packaging might require admin rights.

Issue: #3253 

## Description

- Prohibits the race condition which could occur during parallel download by pre-downloading the checksum values
- Enables building on GitHub UNIX system by excluding mac-universal builds, which can only be build on MacOS
- Provides local build option for quick developer testing
- Includes clearing of download, build and application folders


ce-bo (approved):
LGTM 🚀

Issue #3393, opened 2023-11-03
Remove New Buttons of Custom Views and Suspicious Metrics features


ce-bo:
# Feature request

## Description

As a Maintainer of CodeCharta, I want the new buttons of Custom Views and Suspicious Metric feature to be removed, so that users are not misleaded by outdated buttons.

## Acceptance criteria
- Both new buttons are not displayed anymore

PR #3399, opened 2023-11-08
feature/3393/remove-new-label-badges

Nereboss:
# Remove the 'new' label-badges

The badges were placed on the Custom views and Suspicious metrics menu points

Closes: #3393

## Description

The badges were placed on the menu points to indicate they were new but they are not anymore so they can be removed.
This removes the badges from the html and their sccs-styling.


ce-bo (approved):

phanlezz (approved):
LGTM 🚀

Issue #3400, opened 2023-11-09
Ccsh without flags does not find parsers


fritschldwg:
# Bug

## Expected Behavior

<GIVEN>

WHEN
I run `ccsh` command in my shell without any flags

THEN
The `ccsh` should be able to exectute any parser

## Actual Behavior
When I execute `ccsh` and select any parser, the parser is not found and the following error is thrown

Now configuring sourcecodeparser - generates cc.json from source code.
Could not find the specified parser with the name 'sourcecodeparser - generates cc.json from source code'!
Exception in thread "main" java.lang.IllegalArgumentException: Tried to configure non existing parser!
        at de.maibornwolff.codecharta.tools.ccsh.parser.ParserService$Companion.configureParserSelection(ParserService.kt:26)
        at de.maibornwolff.codecharta.tools.ccsh.parser.InteractiveParserSuggestionDialog$Companion.offerAndGetInteractiveParserSuggestionsAndConfigurations(InteractiveParserSuggestionDialog.kt:29)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh$Companion.executeParserSuggestions(Ccsh.kt:102)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh$Companion.executeCommandLine(Ccsh.kt:94)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh$Companion.main(Ccsh.kt:87)
        at de.maibornwolff.codecharta.tools.ccsh.Ccsh.main(Ccsh.kt)

## Specifications
- CodeCharta Version: 1.120.0
- OS: Windows

PR #3401, opened 2023-11-10
fix/3400/ccsh-not-finding-parser

fritschldwg:
# Fix ccsh not finding parsers without flags

Closes: #3400

## Description

Fixes ccsh not being able to select parsers because parsers were being selected with parsername and description
Fixed by removing parser description when calling the selected parsers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix an issue with web demo on Safari showing a white screen and not loading [#3396](https://github.com/MaibornWolff/codecharta/pull/3396)
--   Fix ccsh not finding parsers when run without flags [#3401](https://github.com/MaibornWolff/codecharta/pull/3401)
 
 ### Chore 👨‍💻 👩‍💻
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix an issue with web demo on Safari showing a white screen and not loading [#3396](https://github.com/MaibornWolff/codecharta/pull/3396)
+-   Fix ccsh not finding parsers when run without flags [#3401](https://github.com/MaibornWolff/codecharta/pull/3401)
```
ce-bo, line 18: I suggest to not have this changelog line. Let's instead add one line for the feature that has been added recently in commit 866544872eb17ba41343e0eaa8eaf3206e89c5e6 (#3387) in section `### Added`. The changelog entry for the feature itself has not been added yet.

By that we can remove the "fixed" line since it has not been released yet.

analysis/tools/ccsh/src/main/kotlin/de/maibornwolff/codecharta/tools/ccsh/parser/InteractiveParserSuggestionDialog.kt
```diff
@@ -21,12 +21,14 @@ class InteractiveParserSuggestionDialog {
                 return emptyMap()
             }
 
+            val parserRepository = PicocliParserRepository()
```
ce-bo, line 24: The repository object is used only in the else right? 
So let's move this line down into the else.

ce-bo (approved):
LGTM :-)

Issue #3336, opened 2023-06-12
Interactive RawTextParser - No default value for TabWidth?


ce-bo:
# Bug

## Expected Behavior

<GIVEN>

WHEN
configuering RawTextParser via `ccsh` parser suggestions

THEN
I can use "empty" as the default value for the tab width, so that it is going to be estimated like stated in the question `What is the tab width used (estimated if not provided)?`

## Actual Behavior
default value (empty) is rejected by "invalid input" message.

## Screenshots
![image](https://github.com/MaibornWolff/codecharta/assets/26900540/66cfeea4-1aba-4eda-af8f-2d14962c66a1)

## Steps to reproduce
1. run `ccsh`
2. select RawTextParser
3. configure the  parser by applying suggested default values
4. using empty default value for tabWith breaks it

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): -
- CodeCharta Version: 1.117.0
- OS: Windows
- Browser: -

PR #3404, opened 2023-11-13
Fix/3336/add default value for tab width

fritschldwg:
# Provide default value for tabwidth

Closes: #3336 

## Description

  - Sets the default value of tabwidth to 'unknown' when no value is entered 
     by the user during configuration of the RawTextParser and tabwidth will get estimated
  - When user enters a non-integer value, the tabwidth gets estimated as well

## Screenshots or gifs
Default tabwidth 3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix an issue with web demo on Safari showing a white screen and not loading [#3396](https://github.com/MaibornWolff/codecharta/pull/3396)
+-   Allow user to skip the value for tabwidth when configuring the rawtextparser and estimate its value [#3404](https://github.com/MaibornWolff/codecharta/pull/3404)
 
 ### Chore 👨‍💻 👩‍💻
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix an issue with web demo on Safari showing a white screen and not loading [#3396](https://github.com/MaibornWolff/codecharta/pull/3396)
--   Allow user to skip the value for tabwidth when configuring the rawtextparser and estimate its value [#3404](https://github.com/MaibornWolff/codecharta/pull/3404)
+-   Fix the ability for users to skip the value for tab-width when configuring the rawtextparser and estimate its value [#3404](https://github.com/MaibornWolff/codecharta/pull/3404)
 
 ### Chore 👨‍💻 👩‍💻
 
--- a/analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ParserDialog.kt
+++ b/analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ParserDialog.kt
@@ -34,8 +34,9 @@ class ParserDialog {
                 hint = "metric1, metric2, metric3 (leave empty for all metrics)"
             )
 
-            val tabWidth: Int =
-                KInquirer.promptInput(message = "What is the tab width used (estimated if unknown)?", default = "unknown", hint = "unknown").toIntOrNull() ?: 0
+            val tabWidth: String =
+                KInquirer.promptInput(message = "What is the tab width used (estimated if unknown)?", default = "unknown", hint = "unknown")
+            val tabWidthValue = tabWidth.toIntOrNull() ?: 0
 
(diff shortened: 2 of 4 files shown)


ce-bo:
I would prefer something like "What is the tab with used (estimated if unknown)?" Default would be "unknown".
Is ist possible to have a string prompt and to cast the value to an int accordingly? If the result of the cast would be a NaN the estimation can take place.

ce-bo (changes requested):
Would you please check my comment?

fritschldwg:
> I would prefer something like "What is the tab with used (estimated if unknown)?" Default would be "unknown". Is ist possible to have a string prompt and to cast the value to an int accordingly? If the result of the cast would be a NaN the estimation can take place.

Implementation of requested changes:

- If user does not enter anything - tabwidth will be estimated
- If user enters non-integer value - tabwidth will be estimated
- If user enters integer-value - tabwidth is set to specified value

ce-bo:
I think `0` should be an invalid `tabWidth`. Not sure how the parser would behave in this case.

fritschldwg:
> > I would prefer something like "What is the tab with used (estimated if unknown)?" Default would be "unknown". Is ist possible to have a string prompt and to cast the value to an int accordingly? If the result of the cast would be a NaN the estimation can take place.
> 
> Implementation of requested changes:
> 
> * If user does not enter anything - tabwidth will be estimated
> * If user enters non-integer value - tabwidth will be estimated
> * If user enters integer-value - tabwidth is set to specified value

Updated implementation:

- If user does not enter anything - tabwidth will be estimated
- If user enters non-integer value - tabwidth will be estimated
- If user enters negative integer-value or zero - tabwidth will be estimated
- If user enters positive integer-value - tabwidth is set to specified value

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 ### Fixed 🐞
 
 -   Fix an issue with web demo on Safari showing a white screen and not loading [#3396](https://github.com/MaibornWolff/codecharta/pull/3396)
+-   Allow user to skip the value for tabwidth when configuring the rawtextparser and estimate its value [#3404](https://github.com/MaibornWolff/codecharta/pull/3404)
```
ce-bo, line 18: Let's mention that a bug has been fixed. :-)

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ParserDialog.kt
```diff
@@ -34,8 +34,9 @@ class ParserDialog {
                 hint = "metric1, metric2, metric3 (leave empty for all metrics)"
             )
 
-            val tabWidth: BigDecimal =
-                KInquirer.promptInputNumber(message = "What is the tab width used (estimated if not provided)?")
+            val tabWidth: String =
+                KInquirer.promptInput(message = "What is the tab width used (estimated if unknown)?", default = "unknown", hint = "unknown")
```
ce-bo, line 38:
```suggestion
                KInquirer.promptInput(message = "What is the tab width used (estimated if unknown)?", default = "unknown", hint = "4")
```

Shall we mention that the tabWidth is counted in spaces / tabs? I think it is not clear, what a tabWidth 4 actually means.

analysis/parser/RawTextParser/src/test/kotlin/de/maibornwolff/codecharta/rawtextparser/ParserDialogTest.kt
```diff
@@ -25,40 +28,97 @@ class ParserDialogTest {
...
     }
 
+    companion object {
+        @JvmStatic
+        fun provideInvalidTabWidth(): List<Arguments> {
+            return listOf(
+                    Arguments.of("unknown"),
+                    Arguments.of("string-value"),
+                    Arguments.of("12."),
+                    Arguments.of("12.0"))
```
ce-bo, line 38:
```suggestion
                    Arguments.of("12.0")),
                    Arguments.of("")),
                    Arguments.of(null)),
                    Arguments.of(false)
```

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/ParserDialog.kt
ce-bo: Do I understand it right, that the tabWidth is needed only when spaces are used as  \t.
I ask this question to better understand how to formulate the question, so that it is more understandable.

ce-bo: Maybe we could ask something like "If spaces are used for indentations, how many of them represent one indentation level aka tab width?"

ce-bo (approved):
LGTM!

Issue #3316, opened 2023-05-25
Interactive RawTextParser asks for excludes twice


moritz-suckow:
# Bug

## Expected Behavior

<GIVEN>

WHEN

using the interactive shell, selecting the RawTextParser and answering the questions

THEN

I get asked for every configuration option once.

## Actual Behavior

I get asked `Do you want to exclude file/folder according to regex pattern?` twice.

## Notes

I looked into the code quickly and saw that the RawTextParser uses the same text for the parameters `excludes` and `fileExtensions`. This was probably just forgotten to change after copy-pasting?

PR #3405, opened 2023-11-15
Fix/3316/rawtextparser asking twice

Nereboss:
# Adjust RawTextParsers question about specific file extensions

RawTextParser was asking one questions about regex patterns twice, the 2nd question should have asked about only including specific file extensions, which was adjusted here.

Closes: #3316

## Description

Adjusted the asked question to correctly represent what the user is supposed to put in.
Also adjusted how the user input to that question was handled, as it previously broke the parsers result for all inputs other than a single file type. This was done by correctly checking for an empty user input (empty string in interactive mode) and splitting the user input in case multiple file types were put in.
Also included Warnings or Errors in case one of the or all the given file types were not found in the folder.


analysis/parser/RawTextParser/README.md
```diff
@@ -23,6 +24,8 @@ generates cc.json from projects or source code files
                              include build, target, dist, resources and out folders
                                as well as files/folders starting with '.'
   -e, --exclude=<exclude>    exclude file/folder according to regex pattern
+  -f, --file-extensions[=<file-extensions>...]
```
ce-bo, line 27: -f could be misleading. I would expect it can be used for specifying an input file

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
```diff
@@ -58,7 +58,10 @@ class MetricCollector(
     }
 
     private fun isParsableFileExtension(path: String): Boolean {
-        return fileExtensions.isEmpty() || fileExtensions.contains(path.substringAfterLast("."))
+        return fileExtensions.isEmpty() ||
+               fileExtensions.contentEquals(arrayOf("")) ||
+               fileExtensions.contains(path.substringAfterLast(".")) ||
+               fileExtensions.contains(".${path.substringAfterLast(".")}")
```
ce-bo, line 64: The two lines seem to be nearly indentical. What is the difference?

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/RawTextParser.kt
```diff
@@ -97,6 +101,30 @@ class RawTextParser(
...
+                }
+            }
+            if (!isFileExtensionIncluded) {
+                notFoundFileExtensions.add(fileExtension)
+            }
+        }
+        if (notFoundFileExtensions.isNotEmpty()) {
+            println()
+            notFoundFileExtensions.forEach { logger.warn("The specified file extension '$it' was not found within the given folder!") }
+        }
```
ce-bo, line 127: Let's move this to a helper function/class

analysis/parser/RawTextParser/src/test/kotlin/de/maibornwolff/codecharta/rawtextparser/MetricCollectorTest.kt
```diff
@@ -33,18 +33,57 @@ class MetricCollectorTest {
...
+        Assertions.assertThat(result).doesNotContainKey("/spaces/spaces_3.xyz")
+        Assertions.assertThat(result).doesNotContainKey("tabs.xyz")
+    }
+
+    @Test
+    fun `Should include only specified File extensions with multiple given`() {
+        val result = MetricCollector(File("src/test/resources/sampleproject").absoluteFile, fileExtensions = arrayOf("wrong", "xyz")).parse()
+
+        Assertions.assertThat(result.size).isEqualTo(4)
+        Assertions.assertThat(result).containsKey("/spaces/spaces_xyz.wrong")
```
ce-bo, line 58: "wrong" confuses me since the result contains this file. 
Let's improve the namings here a bit to indicate what is excluded and what not (.excluded, .included).

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
ce-bo: Ok I learned it from your tests.

ce-bo: Ideally, the fileExtensions Array contains "normalized" or valid file extensions.
At the moment, we always have to take care of different extensions schemas (with . and without) when we use the fileExtensions variable.

ce-bo (commented):
I am done with my review. would you please check my remarks. Thank you!

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
Nereboss: Alternatively `fileExtensions.any{it.contains(Regex(".?${path.substringAfterLast(".")}"))}` could be used to replace both lines but I find it much more difficult to understand

Nereboss: The problem with normalizing the fileExtensions array, is that we also use it to report back to the user if one of their specified file extensions have not been found in the folder. If we were to normalize it, we would need another array which serves a similar purpose to the existing one. Yes, it would be cleaner if it was normalized but I think having to deal with two separate arrays with comparable function would be worse for understandability.

analysis/parser/RawTextParser/README.md
Nereboss: Would it be better to rename it to file-type in the help and use `-t` or `--file-types`?

analysis/parser/RawTextParser/src/main/kotlin/de/maibornwolff/codecharta/parser/rawtextparser/MetricCollector.kt
ce-bo: I am not sure if it would be harder to understand. But it is fine for me to keep it like it is. As soon as we duplicate this logic to check for different alternatives of the extension (with and without dots) we should refactor it.

analysis/parser/RawTextParser/README.md
ce-bo: In that case I would rather go with -e, --extensions.

Nereboss: `-e`, `--exclude` is already in use to exclude files by a regex pattern.
Two other parsers (merge and sourcecodeparser) also use a `-f` flag (for `--move-file` and `--format`) so I think we should either allow it here too or change it for them as well.

 Alternatively we could use `-fe` for `--file-extensions`, similar to how some parser use `-nc` as short for `--not-compressed`.

ce-bo: It is hard to name these things.
What do you think? `-fe` is the best way I think.

analysis/parser/RawTextParser/README.md
```diff
@@ -9,31 +9,33 @@ This parser analyzes code regardless of the programming language used to generat
...
+  -h, --help                displays this help and exits
   -m, --metrics[=<metrics>...]
-                             metrics to be computed (select all if not specified)
-  -o, --outputFile=<outputFile>
-                             output File (or empty for stdout)
-  -p, --projectName=<projectName>
-                             project name
-  -v, --verbose              verbose mode
+                            metrics to be computed (select all if not specified)
+      --max-indentation-level=<maxIndentLvl>
```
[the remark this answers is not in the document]
Nereboss: The decision to change from camelcase was not by us, this change was already made in a commit at the start of 2020, just the readme (and GH-Pages) was never updated. Should this still be classified as a breaking change then?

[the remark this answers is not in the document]
ce-bo: I see, thank you for updating this. I think we don't have to mention this breaking change after 3 years.

ce-bo (approved):

PR #3422, opened 2023-11-30
Replace scanner with reader #3406

fritschldwg:
# Replace scanner with reader #3406

Issue: #3406

## Description

  - Fix for a bug in the piped input project checking code
  - Previously relied on a scanner to detect piped input, assuming no bytes meant no project
  - Bug occurred in interactive mode when pressing enter after the last question, causing bytes to be available
  - The scanner entered reading mode and hung indefinitely, waiting for user input
  - Resolved the issue by switching to a reader instead of a scanner
  - The reader provides a method to check whether the next read is blocking or not, preventing the hang bug
  - Also resolves point 5 of issue #3406
  
## Definition of Done

A PR is only ready for merge once all the following acceptance criteria are fulfilled:
- [x] Changes have been manually tested
- [x] All TODOs related to this PR have been closed
- [x] There are automated tests for newly written code and bug fixes
- [x] All bugs discovered while working on this PR have been submitted as issues (if not already an open issue)
- [x] Documentation (GH-pages, analysis/visualisation READMEs, parser READMEs, --help, etc.) has been updated (almost always necessary except for bug fixes)
- [x] CHANGELOG.md has been updated

Issue #3406, opened 2023-11-15
RawTextParser fails for multiple inputs


fritschldwg:
# RawTextParser fails for multiple inputs
This issue consolidates various problems associated with the RawTextParser. The following are specific problems identified:

- [x] (1) The 'metrics', 'exclude', and 'file-extensions' parameters are not handled correctly (handled in [#3415](https://github.com/MaibornWolff/codecharta/pull/3415))
- [x] (2) The default value for the 'max-indentation-level' parameter is set incorrectly (handled in [#3419](https://github.com/MaibornWolff/codecharta/pull/3419))
- [x] (3) RawTextParser crashes when provided with various 'max-indentation-level' values
- [ ] (4) RawTextParser hangs when using default parameters [#3422](https://github.com/MaibornWolff/codecharta/pull/3422))
- [ ] (5) RawTextParser exhibits diverging behaviors when configured using flags compared to configuration via the KotlinInquirer
- [ ] (6) RawTextParser crashes when given invalid metric [#3424](https://github.com/MaibornWolff/codecharta/pull/3424))

## Expected Behavior (1)

<GIVEN>

WHEN
- Executing 
   - ccsh rawtextparser → leave specified parameters empty
   - OR ccsh → select rawtextparser → leave specified parameters empty
   - OR ccsh -i → select rawtextparser → leave specified parameters empty
   - OR ccsh rawtextparser "path-to-folder-or-file" --metrics= --file-extensions=   
   --exclude=

THEN
- All files and metrics are included for parsing

## Actual Behavior (1)
- No files are included for parsing
- No metrics are included for parsing

## Screenshots (1)
<img width="652" alt="RawTextParser not including files and metrics" src="https://github.com/MaibornWolff/codecharta/assets/149768491/fce15475-8018-47e6-9f0b-b56752768f06">

## Steps to Reproduce the Problem (1)
1. Configure rawtextparser interactively
2. Leave specified parameters empty

## Expected Behavior (2)

<GIVEN>

WHEN
- Executing 
   - ccsh rawtextparser -h 
   - ccsh rawtextparser "path-to-folder-or-file"

THEN
- Max-indentation-level is set to 10 (as specified in the help command)

## Actual Behavior (2)
- Max-indentation-level is set to 6

## Steps to Reproduce the Problem (2)
1. Run: ccsh rawtextparser "path-to-folder-or-file"

## Expected Behavior (3)

<GIVEN>

WHEN
- Executing 
[...]

PR #3428, opened 2023-12-06
Revert "Replace scanner with reader #3406"

fritschldwg:
Reverts MaibornWolff/codecharta#3422

- The characters used as part of the newly introduced sync flag get handled inconsitently accross differenct consoles
- This causes the pipeing functionality to break


BridgeAR (approved):

fritschldwg:
- To keep the main git history clean, we don't revert the changes. Instead, we reopen the old issue, and push the fix to the old branch

Issue #3437, opened 2023-12-08
Deleted nodes in delta mode show up as half-grey half-red, as supposed to full-red


phanlezz:
# Bug: Delta mode shows deleted buildings not completely red

## Expected Behavior

<GIVEN> Two maps with at least one node (/file) that only exists in the first map

WHEN Entering delta mode with said maps

THEN The node that doesn't exist in the second map should be completely red

## Actual Behavior

The deleted node is half-grey half-red

## Screenshots

![grafik](https://github.com/MaibornWolff/codecharta/assets/65733509/ca370483-0ac4-4a12-89ae-bb6405189a36)

(every building on the left should be red, as they are non-existing in the comparison map)

## Steps to Reproduce the Problem
1. Open Web Demo
1. Enter Delta Mode

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes 
- CodeCharta Version: 1.121.0
- Browser: Firefox

## Dev Notes:

- The problem seems to be, that nodes which are not found in the second map get replicated from the first map. Only to set the delta values to the negative of the original ones (`attributes`). Additionally, the attributes should be set to 0 or deleted.

PR #3439, opened 2023-12-08
Fix half-grey half-red buildings for deleted nodes in delta mode

phanlezz:
# Fix half-grey buildings in delta mode for deleted nodes 

In delta mode, non-existing nodes would we be set to the values of the reference node (with identical attribute values like `mcc` and so forth).
This caused double height buildings, because negative delta values are added on top of the building height.

When the attributes instead get deleted or set to 0 the initial height of the building is not the same as the reference and the negative delta can be added on top as usual.

Closes: #3437 

## Open Points:

- [x] ~~The calculated buildings have a small grey strip across the bottom (height of a folder block) because `treeMapGenerator` and `treeMapHelper` use a minimal height of 2. Is this acceptable? (More implications *1)~~
- [x] Verify that folders don't cause any hiccups
- [x] Verify that folder's attribute values are still correct (sum of attributes / metrics)
- [x] Check possibility to add tests for the delta mode calculations after rendering.

## Definition of Done

- [x] All bugs discovered while working on this PR have been submitted as issues (if not already an open issue) #3472 #3470

## Screenshots or gifs
grafik

grafik

#### Added some code to calculate correct edge starting points:

| online | this branch |
|--------|--------|
| grafik | grafik | 

Reference of the current delta mode:
grafik


visualization/app/codeCharta/util/deltaGenerator.ts
```diff
@@ -98,6 +98,8 @@ export class DeltaGenerator {
 
 			for (const [key, value] of Object.entries(node.attributes)) {
 				node.deltas[key] = -value
+				delete node.attributes[key]
+				// node.attributes[key] = 0
```
BridgeAR, line 102: Setting it to zero seems better as the attribute should exist.

If you do not want to set any attributes (which I definitely would), it would be better to just assign `node.attributes = {}` after the loop. That way it's also faster as deleting properties from an object changes the internal form and the object access becomes slow.

BridgeAR (commented):
This is definitely the right direction!

The tests need an update though.

phanlezz:
Question:
If a node is not present in the second set (comparison), the file/node is not available in the node menu, because it doesn't have an area metric associated with them (value of `0` seems to trigger this). 

| Online | This branch |
|--------|--------|
| grafik| grafik| 

Is this fine as is, or should I look into this? I'm not sure if the node menu can be easily adapted to accept zeroes for buildings with delta values.

phanlezz:
(not specific for this branch, but)
Shouldn't a click on a node in the node/file menu select the clicked building on the map? (Like a direct click on the building would do?)

BridgeAR:
@phanlezz 
> If a node is not present in the second set (comparison), the file/node is not available in the node menu, because it doesn't have an area metric associated with them (value of 0 seems to trigger this).

That is a good question!

After giving it some more thought, I actually believe we should keep the attributes as they are and just change our drawing algorithm. Instead of just adding the delta value on top of the current attribute for removed entries, just draw them as being flat while keeping everything else in place. What do you think?

phanlezz:
> @BridgeAR  
> That is a good question!
> 
> After giving it some more thought, I actually believe we should keep the attributes as they are and just change our drawing algorithm. Instead of just adding the delta value on top of the current attribute for removed entries, just draw them as being flat while keeping everything else in place. What do you think?

I'm not sure. flat or greyed-out buildings suggest to me, that there is no change in these files.
But I don't really have a better idea on how to approach this. Maybe we could use a different color?

If we choose to make them flat, we also must set the attributes to zero, as they are directly displayed in the sidebar.

visualization/app/codeCharta/util/deltaGenerator.ts
ce-bo: If the edge metric is not used in delta mode, we could think about it to remove it.

phanlezz:
Concluding: 

- Δ Removed buildings/nodes should be completely red
- Δ Attributes in Sidebar should be zero with negative deltas for deleted nodes
- Δ In File/Node explorer deleted nodes should not be grayed out (HTML title needs to be updated also)
- Bonus: in File/Node explorer clicked nodes should highlight the corresponding nodes on the map

BridgeAR:
A couple of tests are still failing :)

```
Test Suites: 4 failed, 259 passed, 263 total
Tests:       8 failed, 1157 passed, 1165 total
Snapshots:   4 failed, 28 passed, 32 total
```

phanlezz:
grafik
Thanks sonar, I will

visualization/app/codeCharta/ui/searchPanel/mapTreeView/mapTreeViewLevel/mapTreeViewLevel.component.spec.ts
```diff
@@ -126,7 +126,7 @@ describe("mapTreeViewLevel", () => {
 		detectChanges()
 
 		const bigLeaf = container.querySelector("#\\/root\\/bigLeaf")
-		const isAngularGreen = bigLeaf.querySelector(".angular-green")
+		const isAngularGreen = bigLeaf.querySelector(".tree-search-result")
```
Nereboss, line 129: we dont search for green anymore so the name of the const should be adjusted aswell

Nereboss (approved):
Early approve, change the one variable name and then all is good to merge 👍

Issue #3080, opened 2022-10-07
Screenshot Improvements #2


ce-bo:
# Feature request

## Description

As a user, I want to export a screenshot of my map which is not extended to the whole screen size, so that screenshots are as small as possible.

## Acceptance criteria
- [ ] There is an option to do a cropped screenshot of the whole map, so that the complete map fills the complete screenshot with a tiny edge (edge size of up to 3% on each side).
    - [ ] Current view angle and zoom is respected

fritschldwg:
@ce-bo I remember us discussing this issue a while back. However I'm not certain anymore what came of that discussion. Is this feature still relevant, now that the legend panel is part of the screenshot? Should it be possible to do a cropped screenshot only when the legend panel is collapsed? If you could leave me a quick comment on what exactly we want to do

fritschldwg:
@ce-bo I recall us discussing this issue previously, but I'm unsure of the outcome. With the legend panel now included in the screenshot, is this feature still necessary? Should we enable cropped screenshots only when the legend panel is collapsed? I'd appreciate a quick update on our plans for this feature.

BridgeAR:
The feature sounds really useful for audits to improve the overall workflow. I would definitely like to get this in. The cropping should ideally happen no matter if the legend is included or not. If the legend is included, it would be considered the outer bounds.

PR #3527, opened 2024-03-21
Crop screenshots

fritschldwg:
# Crop screenshots

Closes: #3080

## Description

  - Remove whitespace on screenshots

## Definition of Done


Nereboss (commented):
No code quality issues but two points should be addressed before merging:
1. We should add tests for the cropping if possible
2. We should decide how the legend should be "attached" to the map. Either 1) like it is now, that the legend simply counts to the bounds of the cropping (which could lead to the legend overlapping the map or a lot of space between the map and the legend) or 2) that the map is always put right next to the map (would make it never overlap the map but this is maybe not what a user expects from a screenshot)

fritschldwg:
> No code quality issues but two points should be addressed before merging:
> 
> 1. We should add tests for the cropping if possible
> 2. We should decide how the legend should be "attached" to the map. Either 1) like it is now, that the legend simply counts to the bounds of the cropping (which could lead to the legend overlapping the map or a lot of space between the map and the legend) or 2) that the map is always put right next to the map (would make it never overlap the map but this is maybe not what a user expects from a screenshot)

Just added tests for the cropping part. The legend should indeed be included as is the map.

Issue #3385, opened 2023-10-26
Failed to load SLF4J during testing


Nereboss:
# Bug

## Expected Behavior

WHEN

Starting tests (either manual or automatic via gradle) 

THEN

As a developer I want the tests to only show test-relevant error and warnings

## Actual Behavior

Additionally, the following error messages appear:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

## Specifications
- CodeCharta Version: 1.119.1
- OS: All

## Developer Notes
Fix: adding the SLF4J libraries to the correct build.gradle of the sub-modules removes the error

Issue #3308, opened 2023-05-17
gitlogparser doesn't work with docker file


ukinimod:
# Bug

## Expected Behavior

GIVEN a git repository

WHEN

I run 

`docker run -it -v $(pwd):/src --workdir /src ccsh ccsh gitlogparser repo-scan --repo-path /src`

in this git repository

THEN

it doesn't run the git-log and filename-list  creation 
and puts out an almost empty cc.json

EXPECTED

it should run the git-log and filename-list creation
it should put out the same cc.json when running ccsh directly

WORKAROUND

the first problem about running the git commands can be solve by specifying the correct user for running the docker container:
`docker run -it -v $(pwd):/src --workdir /src --user 1000:1000 ccsh ccsh gitlogparser repo-scan --repo-path /src`
This should probably be specified correctly in the Dockerfile and not at execution time.

nevertheless specifying the above still results in an almost empty cc.json. 

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz): yes | no
- CodeCharta Version: 1.116.0
- Env: Windows WSL2, Rancher Desktop

phanlezz:
> nevertheless specifying the above still results in an almost empty cc.json.

I produced the `cc.json` via a NPM install and via the docker compose setup. Both files were equal.
Regarding the first part:
I need to do some more investigation why this command is required `--user 1000:1000`.

PR #3571, opened 2024-04-22
Fix `GitLogParser` in docker container

phanlezz:
# Fix `GitLogParser` in docker container

- Fix syntax in `Dockerfile` so that sonar-scanner can be downloaded correctly 
- change tokei download to static link (not latest anymore)
- Add a non-root user as default to docker image

- [x] Is it necessary to declare the user in the compose file?
- [x]  Add an exit code validation, to verify if git exited correctly.

Closes: #3308
Closes: #3385  

## Definition of Done

- - [x] Check code-maat
- - [x] Check tokei
- - [x] Check sonar-scanner (callable, but requires running sonar qube instance at localhost:9000)
- - [x] Check gitlogparser

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -152,7 +152,7 @@ jobs:
         run: |
           docker build ./analysis -t local-ccsh
           current_version=$(node -pe "require('./analysis/node-wrapper/package.json').version")
-          docker run -w /home/nonroot-docker-user/cc/analysis/test -v $(pwd):/home/nonroot-docker-user/cc local-ccsh bash ./golden_test.sh ${current_version} /usr/local/bin/ccsh
+          docker run -w /home/nonroot-docker-user/cc/analysis/test --user root:root -v $(pwd):/home/nonroot-docker-user/cc local-ccsh bash ./golden_test.sh ${current_version} /usr/local/bin/ccsh
         working-directory: ${{env.project-directory}}
 
       - name: Setup Java JDK
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,9 +150,10 @@ jobs:
 
       - name: Docker Test Analysis
         run: |
+          pwd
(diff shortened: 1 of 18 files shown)


phanlezz:
It might not be possible to test our Dockerfile with a custom user:
https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user

phanlezz:
I'm not sure why the e2e tests failed, now that the Docker Tests ran through 😆.
Problem is that the current setup is not really applicable:

- I changed the user id (`uid`) inside the docker container to `1001`, which is the `uid` of the GitHub runner via the `Dockerfile`
- Now the owner `uid` of the repository files matches the `uid` inside the container and git can run (without [dubious ownership](https://github.com/MaibornWolff/codecharta/issues/3308#issuecomment-2066319617) issues)
- BUT: If now executed locally on e.g. Ubuntu with a native user 1000, git doesn't work anymore (because now the repository file owner is different to the docker user `uid:1000 != uid:1001`)

For me, this means, that we can't really just use `uid:1000` as proposed in the original issue, because a regular user of the docker might have a different `uid`.
 
 Some ideas:
 
 ### ARG
 We could use an `ARG` in the `analysis/Dockerfile` where we default a value of `1000` and inject a value of `1001` for GitHub.
 This does not fix the issue with regular users, which have `uid`s different to `1000`. 
 > We could accept this, and the user would have to adept in this special case. There might also be a possibility of a dynamic `uid` on container creation which copyies the host `uid`.
 
 ### Entrypoint 
 With an entrypoint script, we might have different options to deal with this problem. Changing the `uid`, using `git config` to declare the folders as safe, or changing ownership of the files.
 
 I suggest we try to use the ARG approach, as it keeps the test run and the regular usage close together.

phanlezz:
Sonar down? 👀

ViktoriaGordeevaVG (approved):

ViktoriaGordeevaVG (approved):

Issue #2498, opened 2021-11-11
Remove example maps as soon as the user loads own cc.json files


BridgeAR:
# Bug

## Expected Behavior

GIVEN you just open up CodeCharta

WHEN loading a map

THEN the example maps (the default ones) should be removed again

## Actual Behavior

The example maps are kept similar to other maps. These are however not useful and should not count as user loaded maps.

## Story
Imagine the following scenario:
- Person opens web demo
- Sees CodeCharta with Analaysis and Visualization
- uploads his own maps
- Visualization and analysis should be removed as they want to focus on the stuff that they uploaded
To make this easier we should probably just remove the maps from the url query parameters once

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json&file=codecharta_analysis.cc.json): yes 
- CodeCharta Version: 1.83.1
- OS: all
- Browser: all

Issue #3641, opened 2024-06-13
Make reload button clear webstorage


Nereboss:
# Feature request

Make the reload button not only reset the maps and settings but also the web storage as some parts of our application are stored there (e.g. the metric templates or custom views). 

![Image](https://github.com/user-attachments/assets/9b94cd16-2cb0-4886-be23-2967fe6c9323)

## Description

As a user, I want the reset button to clear the web storage so that it feels like a real reset button because currently some stuff from the old session is kept.

## AC
- Add multiple checkboxes so you can decide on what to clear and what not
- - (try to group some settings together as good as possible, here is an example. Come up with something better for the user :))![Image](https://github.com/user-attachments/assets/ac360bfa-b79e-4209-be8a-d949216c5070)
- Update all the descriptions / tooltips related to the modal. To make it explicit what is getting deleted
- After selecting multiple checkboxes and hitting accept, clear all the corresponding data and only reload if needed

 
## Open questions
Should we include settings for the reset button what it will reset?

phanlezz:
Maybe we could include a hard reset button in the settings panel

ChristianHuehn:
I wouldnt want to mess with the local storage in general. I would rather do this: https://github.com/MaibornWolff/codecharta/issues/2572

and define what we want to keep in localstorage etc.

phanlezz:
⚠️ We may need to include some sort of workflow where we increase the version of the webstorge (or invalid old storage after an update) as non-power uses will get confused, when many changes in the changelog won't work without clearing the old webstorage
.

PR #3768, opened 2024-09-24
Fix/2498/remove example maps

BenediktMehl:
# remove example maps

Closes: #2498 

## Description

If there are sample files loaded as fallback maps when starting the application,
then those files are removed as soon as the user uploads own files.

## Definition of Done

- [ ] Changes have been manually tested
- [ ] All TODOs related to this PR have been closed
- [ ] There are automated tests for newly written code and bug fixes
- [ ] All bugs discovered while working on this PR have been submitted as issues (if not already an open issue)
- [ ] Documentation (GH-pages, analysis/visualization READMEs, parser READMEs, --help, etc.) has been updated (almost always necessary except for bug fixes)
- [ ] CHANGELOG.md has been updated


phanlezz:
Clearing webstorage war required. #3641.

phanlezz:
Currently clears sample1,sample2 when opening dev/standalone. Does not clear default maps when opening webdemo.
I think that behavior is good. What do you think @ChristianHuehn ?

CodeCharta:
Nah it should also delete them on the webdemo. 

In general we just want to get the following:
- Person opens web demo
- Sees CodeCharta with Analaysis and Visualization
- uploads his own maps
- Visualization and analysis should be removed as they want to focus on the stuff that they uploaded

To make this easier we should probably just remove the maps from the url query parameters once

BenediktMehl:
> To make this easier we should probably just remove the maps from the url query parameters once

Fixed the issue without addressing the removal of the URL parameter "file". I think this should be a new issue, since currently the URL parameter "file" is only related with the shown maps on initial loading of the site.

phanlezz (approved):

PR #3813, opened 2024-11-04
Tech(visualization): Change default test setting, add dependency doc, update rebase doc, format dev guide doc

phanlezz:
### Visualization: `npm run test` now uses only 25% of the available threads

Per default, jest uses all CPU cores and each one starts a node process which can easily reserve all system memory. This can bog down the test speed, because the system swap is used (when the RAM is full)

## Description

- Changes the npm run test command to append the `--maxWorkers=25%` setting
- New pipeline test run is now `npm run test:ci`
- DEV_START_GUIDE
- - Format headlines
- - Update section about rebase
- - Add section about dependency workflow
- Remove `.prettierignore` (biome is used now)
- Modify comment section in `.editorconfig`


IhsenBouallegue (approved):
LGTM

Issue #3848, opened 2024-12-12
CCjson from Gitlogparser with Authors cannot be imported


ChristianHuehn:
# Bug

## Expected Behavior
GIVEN
user navigated to the codecharta root folder

WHEN
ccsh gitlogparser "repo-scan" "--output-file=git" "--silent=true" "--add-author=true" "--repo-path=<path/to/root/Folder>"

THEN
User should be able to import the cc json into codecharta

## Actual Behavior
Webstudio shows an error ![Image](https://github.com/user-attachments/assets/2c497b68-8539-47cd-a00d-a6df41c13990)

## Steps to Reproduce the Problem
1. Use the gitlogparser with --add-author=true
example: ccsh gitlogparser "repo-scan" "--output-file=git" "--silent=true" "--add-author=true" "--repo-path=<path/to/root/Folder>"
2. Import the git.cc.json.gz into our webstudio
3. Profit

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz&currentFilesAreSampleFiles=true): yes | no
- CodeCharta Version:
- OS:
- Browser:

ChristianHuehn:
As we are currently dont have any use for the authors, we will ignore it as name for now.

PR #3897, opened 2025-01-28
Bug/3848/gitlog with authors

BenediktMehl:
# CCjson from Gitlogparser with Authors cannot be imported

Closes: #3848 

## Description

Added a remove Authors function that removes all Authors Attributes.
For every removed Authors Attribute there is a warning created.
Removing the Authors is done before validating the cc.json, so the automatic creating of the "generatedSchema.json" can still be used.

## Definition of Done

### DIFF
--- a/visualization/CHANGELOG.md
+++ b/visualization/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/)
 
 ### Fixed 🐞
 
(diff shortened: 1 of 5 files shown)


IhsenBouallegue (approved):
Code looks good 👍  just quick question whyis authors there if we do not need it in the first place? don't we parse this ourselves in ccsh?

BenediktMehl:
Yes you are right, at the moment it's not really used in the vis part, but it can be exported in the analysis (no body knows why we have this already). But the idea is to give some functionality to that in the future, which is why we wanted to fix the bug, instead of removing the feature in the ana part.

Issue #3899, opened 2025-01-29
Uploading the same map twice results in an endless loop


ChristianHuehn:
# Bug

## Expected Behavior

WHEN a map with is uploaded twice
THEN the map should be ignored and codecharta should still be operable

## Actual Behavior
CodeCharta shows an endless loading spinner

## Specifications
- is released in [online-demo](https://maibornwolff.github.io/codecharta/visualization/app/index.html?file=codecharta.cc.json.gz&file=codecharta_analysis.cc.json.gz&currentFilesAreSampleFiles=true): yes 
- CodeCharta Version: Prod

PR #3901, opened 2025-01-29
Bug/3899/upload same map twice

BenediktMehl:
# App freeze when uploading a file that is already loaded

Closes: #3899 

## Description

When uploading, if all files are already uploaded, the application remains stuck in the loading animation. The visibleFiles does not change if only already uploaded files are uploaded again (This behavior was introduced in #3788). This was intentional, but it had the side effect that the RgNx store did not trigger the effects that would normally follow a file upload. One of these effects was responsible for removing the loading animation.

I fixed the issue by checking if a file upload added any new files. If no new files were added, there is an error that is caught by the file upload process, which then stops the loading animation.

## Definition of Done

### DIFF
--- a/visualization/app/codeCharta/services/loadFile/fileParser.ts
+++ b/visualization/app/codeCharta/services/loadFile/fileParser.ts
@@ -46,7 +46,8 @@ export function enrichFileStatesAndRecentFilesWithValidationResults(
     fileValidationResults: FileValidationResult[],
     currentFilesAreSampleFilesCallback: () => boolean,
     setCurrentFilesAreNotSampleFilesCallback: () => void
-) {
+): boolean {
+    let newFilesWereAdded = false
     for (const nameDataPair of nameDataPairs) {
         const fileValidationResult: FileValidationResult = {
             fileName: nameDataPair?.fileName,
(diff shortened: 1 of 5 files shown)


visualization/app/codeCharta/services/loadFile/fileParser.ts
```diff
@@ -52,13 +53,23 @@ export function enrichFileStatesAndRecentFilesWithValidationResults(
 
         if (fileValidationResult.errors.length === 0) {
             fileValidationResult.warnings.push(...checkWarnings(nameDataPair?.content))
-            addFile(fileStates, recentFiles, nameDataPair, currentFilesAreSampleFilesCallback, setCurrentFilesAreNotSampleFilesCallback)
+            const addedFile = addFile(
```
IhsenBouallegue, line 56: small nitpick but this could be `hasAddedFile` instead of `addedFile` as its not the file itself and simply the boolean state of the operation. but to avoid this alltogether we can do
```
newFilesWereAdded = addFile(...)
```

IhsenBouallegue (approved):
Just a small comment but otherwise it works and love that we have tests for it now 🥇

visualization/app/codeCharta/services/loadFile/fileParser.ts
BenediktMehl: Good catch

PR #4275, opened 2025-10-20
style: prepare semmantic commits for different configurations

ChristianHuehn:
Clean up semantic commits implementation to later add different commit styles

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
 npx lint-staged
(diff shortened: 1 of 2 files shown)


analysis/analysers/parsers/GitLogParser/src/main/kotlin/de/maibornwolff/codecharta/analysers/parsers/gitlog/input/metrics/MatchPattern.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    data class Regex(val pattern: String) : MatchPattern() {
+        private val regex = kotlin.text.Regex(pattern, RegexOption.IGNORE_CASE)
+
+        override fun matches(message: String): Boolean {
+            return regex.containsMatchIn(message.trim())
+        }
+    }
+}
```
ChristianHuehn: check if this is really needed, follow yagni

